PR #8, opened 2024-12-31
chore: Add devcontainer support

tutman96:
Adds a devcontainer configuration with golang and nodejs support. Also updates `./dev_deploy.sh` to not require a python http server, making it easier to deploy from within a container network environment. 

This was tested on my JetKVM device, but a second round of validation wouldn't hurt.

### DIFF
--- a/dev_deploy.sh
+++ b/dev_deploy.sh
@@ -3,45 +3,34 @@ set -e
 
 # Function to display help message
 show_help() {
-    echo "Usage: $0 [options] -h <host_ip> -r <remote_ip>"
+    echo "Usage: $0 [options] -r <remote_ip>"
     echo
     echo "Required:"
-    echo "  -h, --host <host_ip>      Local host IP address"
     echo "  -r, --remote <remote_ip>   Remote host IP address"
     echo
     echo "Optional:"
     echo "  -u, --user <remote_user>   Remote username (default: root)"
-    echo "  -p, --port <port>          Python server port (default: 8000)"
     echo "      --help                 Display this help message"
     echo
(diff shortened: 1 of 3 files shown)


dev_deploy.sh
```diff
@@ -69,12 +58,8 @@ make build_dev
...
 cd bin
 
-# Start a Python HTTP server in the background to serve files
-python3 -m http.server "$PYTHON_PORT" &
-PYTHON_SERVER_PID=$!
-
-# Ensure that the Python server is terminated if the script exits unexpectedly
-trap "echo 'Terminating Python server...'; kill $PYTHON_SERVER_PID" EXIT
+# Copy the binary to the remote host
+cat jetkvm_app | ssh "${REMOTE_USER}@${REMOTE_HOST}" "cat > $REMOTE_PATH/jetkvm_app_debug"
```
SuperQ, line 62: Why not use `scp`?

tutman96: I found out when writing https://medium.com/@brandontuttle/installing-tailscale-on-a-jetkvm-3c72355b7eb0 that dropbear doesn't support scp (notably sftp)

SuperQ: I wonder if `-O` would work.

> Use  the  legacy  SCP protocol for file transfers instead of the SFTP protocol.`

tutman96: image
good thought, but I think that requires `scp` to installed on the remote

SuperQ (approved):

adamshiervani (approved):

PR #10, opened 2025-01-01
WIP: Add plugin system

tutman96:
Based on discussion #9 implement the basics of a plugin system that can be used to extend the functionality of JetKVM.

In order to accomplish this, the task is broken down into many steps:

- [X] Support uploading plugin archives from the web interface
- [ ] Extract and validate the plugin
- [ ] Web RPC methods to enumerate installed plugins and list them in the UI dynamically
- [ ] Plugin subprocess system that can launch and shutdown a plugin
- [ ] Uninstalling a plugin
- [ ] JSON RPC calls to the plugin and report back the results of getPluginStatus to the UI
- [ ] Dynamic plugin configuration UI (might be a separate PR)

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -1,0 +1,53 @@
+package plugin
+
+import (
+	"fmt"
+	"kvm/internal/storage"
+	"os"
+	"path"
+
+	"github.com/google/uuid"
+)
+
+const pluginsFolder = "/userdata/jetkvm/plugins"
+const pluginsUploadFolder = pluginsFolder + "/_uploads"
+
+func init() {
+	_ = os.MkdirAll(pluginsUploadFolder, 0755)
+}
+
+func RpcPluginStartUpload(filename string, size int64) (*storage.StorageFileUpload, error) {
+	sanitizedFilename, err := storage.SanitizeFilename(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	filePath := path.Join(pluginsUploadFolder, sanitizedFilename)
+	uploadPath := filePath + ".incomplete"
+
+	if _, err := os.Stat(filePath); err == nil {
+		return nil, fmt.Errorf("file already exists: %s", sanitizedFilename)
+	}
+
+	var alreadyUploadedBytes int64 = 0
+	if stat, err := os.Stat(uploadPath); err == nil {
+		alreadyUploadedBytes = stat.Size()
+	}
+
+	uploadId := "plugin_" + uuid.New().String()
+	file, err := os.OpenFile(uploadPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file for upload: %v", err)
+	}
+
+	storage.AddPendingUpload(uploadId, storage.PendingUpload{
+		File:                 file,
+		Size:                 size,
+		AlreadyUploadedBytes: alreadyUploadedBytes,
+	})
+
+	return &storage.StorageFileUpload{
+		AlreadyUploadedBytes: alreadyUploadedBytes,
+		DataChannel:          uploadId,
+	}, nil
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,10 @@
+{
+	"name": "JetKVM",
+	"image": "mcr.microsoft.com/devcontainers/go:1-1.23-bookworm",
+	"features": {
+		"ghcr.io/devcontainers/features/node:1": {
+			// Should match what is defined in ui/package.json
+			"version": "21.1.0"
+		}
+	}
+}
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build_dev: hash_resource
 	GOOS=linux GOARCH=arm GOARM=7 go build -ldflags="-s -w -X kvm.builtAppVersion=$(VERSION_DEV)" -o bin/jetkvm_app cmd/main.go
 
 frontend:
-	cd ui && npm run build:device
+	cd ui && npm ci && npm run build:device
 
 dev_release: build_dev
 	@echo "Uploading release..."
--- a/dev_deploy.sh
+++ b/dev_deploy.sh
@@ -3,45 +3,34 @@ set -e
 
 # Function to display help message
 show_help() {
-    echo "Usage: $0 [options] -h <host_ip> -r <remote_ip>"
+    echo "Usage: $0 [options] -r <remote_ip>"
     echo
     echo "Required:"
-    echo "  -h, --host <host_ip>      Local host IP address"
     echo "  -r, --remote <remote_ip>   Remote host IP address"
     echo
     echo "Optional:"
     echo "  -u, --user <remote_user>   Remote username (default: root)"
-    echo "  -p, --port <port>          Python server port (default: 8000)"
     echo "      --help                 Display this help message"
     echo
     echo "Example:"
-    echo "  $0 -h 192.168.0.13 -r 192.168.0.17"
-    echo "  $0 -h 192.168.0.13 -r 192.168.0.17 -u admin -p 8080"
+    echo "  $0 -r 192.168.0.17"
+    echo "  $0 -r 192.168.0.17 -u admin"
     exit 0
 }
 
 # Default values
-PYTHON_PORT=8000
 REMOTE_USER="root"
 REMOTE_PATH="/userdata/jetkvm/bin"
 
 # Parse command line arguments
 while [[ $# -gt 0 ]]; do
     case $1 in
-        -h|--host)
-            HOST_IP="$2"
-            shift 2
-            ;;
         -r|--remote)
             REMOTE_HOST="$2"
             shift 2
             ;;
         -u|--user)
             REMOTE_USER="$2"
-            shift 2
-            ;;
-        -p|--port)
-            PYTHON_PORT="$2"
             shift 2
             ;;
         --help)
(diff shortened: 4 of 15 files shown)


tutman96:
Also this currently stacks on top of #8 since that is what I used for development

tutman96:
First item `Support uploading plugin archives from the web interface` has been completed.

image
image
image

tutman96:
Finished support for extracting and validating the manifest.json of the plugin

image
image

internal/plugin/database.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create plugin database: %v", err)
+	}
+	defer file.Close()
+
+	encoder := json.NewEncoder(file)
+	encoder.SetIndent("", "  ")
+	if err := encoder.Encode(d); err != nil {
+		return fmt.Errorf("failed to encode plugin database: %v", err)
+	}
```
eric, line 51: Encode the JSON before saving the file will make sure that encoder issues won't cause an empty file to be written.

It also could be worth a strategy that writes the file to a different filename and then moves it into place (or use an atomic file library) to make sure that there isn't issues with disk space, etc.

internal/plugin/extract.go
```diff
@@ -0,0 +1,83 @@
...
+func extractPlugin(filePath string) (*string, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file for extraction: %v", err)
+	}
+	defer file.Close()
+
+	var reader io.Reader = file
+	// TODO: there's probably a better way of doing this without relying on the file extension
+	if strings.HasSuffix(filePath, ".gz") {
```
eric, line 31: Grabbing the first few bytes of the file and doing:

```
if bytes.HasPrefix(buf, []byte{0x1F, 0x8B, 0x08}) {
```

can be a nicer way to handle this.

internal/plugin/database.go
tutman96: Good call. I wrote it to a tmp file and then moved it as that solves both problems at once. Thanks for the suggestion.

tutman96:
Plugins are now listed dynamically, and can be enabled and disabled.

image
image
image

tutman96:
Subprocess support has been built out with IPC connection creation and exponential backoff auto-restart while enabled. There is also graceful shutdown support, where the process manager will send a SIGTERM, wait 30s and if not gracefully shutdown, forcefully kill the process.

image

internal/plugin/extract.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/google/uuid"
+)
+
+const pluginsExtractsFolder = pluginsFolder + "/extracts"
+
+func init() {
+	_ = os.MkdirAll(pluginsExtractsFolder, 0755)
+}
+
+func extractPlugin(filePath string) (*string, error) {
```
eric, line 22: Is there a reason why this is a `*string` instead of just a `string`?

internal/plugin/extract.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+			defer file.Close()
+
+			if _, err := io.Copy(file, tarReader); err != nil {
+				return nil, fmt.Errorf("failed to extract file: %v", err)
+			}
+		default:
+			return nil, fmt.Errorf("unsupported tar entry type: %v", header.Typeflag)
+		}
+	}
```
eric, line 80: Seems like it would be nice to clean up the directory on error so that it doesn't fill up disk space.

internal/plugin/extract.go
tutman96: I wanted to return nil, but realize now that's redundant with returning a non-nil error, so instead just return an empty string on error.

internal/plugin/extract.go
tutman96: Good point. Done!

internal/plugin/install.go
```diff
@@ -0,0 +1,154 @@
...
+	// Current active version of the plugin
+	Version string `json:"version"`
+
+	// Map of a plugin version to the extracted directory
+	ExtractedVersions map[string]string `json:"extracted_versions"`
+
+	manifest       *PluginManifest
+	runningVersion *string
+	processManager *ProcessManager
+	rpcListener    *net.Listener
```
eric, line 25: `net.Listener` is already an interface (which is a pointer) so you don't need to make this a pointer to be able to check for `nil`.

internal/plugin/process_manager.go
```diff
@@ -0,0 +1,113 @@
...
+		pm.LastError = fmt.Errorf("process exited with error: %w", err)
+		pm.scheduleRestart()
+	}
+}
+
+func (pm *ProcessManager) scheduleRestart() {
+	if pm.enabled {
+		log.Printf("Restarting process in %v...", pm.backoff)
+		time.Sleep(pm.backoff)
+		pm.backoff *= 2 // Exponential backoff
```
eric, line 76: Might want to put an upper limit on the backoff or it'll be so big as to be not useful anymore.

internal/plugin/process_manager.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"log"
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+// TODO: this can probably be defaulted to this, but overwritten on a per-plugin basis
+const GRACEFUL_SHUTDOWN_DELAY = 30 * time.Second
+const MAX_RESTART_BACKOFF = 30 * time.Second
```
joshuasing, line 13: I think these could probably be combined into one `const` block, like this:
```suggestion
const (
	GRACEFUL_SHUTDOWN_DELAY = 30 * time.Second
	MAX_RESTART_BACKOFF = 30 * time.Second
)
```

Effective Go and the Go stdlib also usually prefers lowerCamelCase/UpperCamelCase (depending on whether exported) for constant names. E.g., per https://go.dev/wiki/CodeReviewComments:
> See https://go.dev/doc/effective_go#mixed-caps. This applies even when it breaks conventions in other languages. For example an unexported constant is maxLength not MaxLength or MAX_LENGTH.

joshuasing (commented):
Thanks for working on this - it looks super cool, and I cannot wait to receive my JetKVM and test it out :)

I wanted to comment on a few things that stood out whilst reading the code - please see below.

P.S. There seem to be a few files without newlines at the end, which might not be preferable.

eric:
It might be nice to pull this jsonrpc refactor into its own pull request that gets merged separately so that it doesn't confuse what is related to the plugin change vs just general improvements to use it later.

tutman96:
> It might be nice to pull this jsonrpc refactor into its own pull request that gets merged separately so that it doesn't confuse what is related to the plugin change vs just general improvements to use it later.

https://github.com/jetkvm/kvm/pull/28 👍

tutman96:
image
JSON RPC calls have been implemented, and both error messages and pending-configuration states are displaying messages in the settings dialog for the plugin

tutman96:
Going to do the configuration after this work has been merged since I technically don't need it for Tailscale support :D not that it would be much help for other things without it. Removing the WIP status on this now as it's ready to review in its current state

kashalls:
Is it possible to read from the package and show what version of the plugin that instance is running? What happens if you install two or more of the same/different version of a plugin?

tutman96:
> Is it possible to read from the package and show what version of the plugin that instance is running? What happens if you install two or more of the same/different version of a plugin?

The system will just no-op if you try to install the same version. A different version should shut down the old one and start the new one up.... let me test that.

tutman96:
> Is it possible to read from the package and show what version of the plugin that instance is running? What happens if you install two or more of the same/different version of a plugin?

Added running version, description, and homepage to the configuration dialog
image

Nevexo:
For anyone looking to try this feature out before it's merged (without building it yourself), I've pulled it into jetkvm-next and added a built binary: 
https://github.com/Nevexo/jetkvm-kvm/releases/tag/next-5

Nevexo (approved):

goncalossilva:
Hey, all! I re-read the conversation just now and I can't tell what the next step is. I would very much like to see this go live!

internal/plugin/plugin.go
```diff
@@ -0,0 +1,257 @@
...
+	if err := pluginDatabase.Save(); err != nil {
+		return fmt.Errorf("failed to save plugin database: %v", err)
+	}
+
+	err := pluginInstall.ReconcileSubprocess()
+	if err != nil {
+		return fmt.Errorf("failed to start plugin %s: %v", name, err)
+	}
+
+	// TODO: Determine if the old extract should be removed
```
IDisposable, line 151: Maybe simplify life by just offering a "cleanup temporaries" button that deletes all historical/inactive versions and also [any partial uploads](https://github.com/jetkvm/kvm/pull/10/files#diff-034908473a1bf69a8b56962feef4c44ae39cf2ae61eebe107cf0a05d19a6809fR47) and [temp files](https://github.com/jetkvm/kvm/pull/10/files#diff-c8b8c056089c83b8be8f143dec61eb13aef9f1bd3c9d38c7b685566fb372e577R44)

internal/plugin/rpc.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	log.Printf("Plugin has supported methods: %v", supportedMethodsResponse.SupportedRpcMethods)
+
+	if !slices.Contains(supportedMethodsResponse.SupportedRpcMethods, "getPluginStatus") {
+		log.Printf("Plugin does not support getPluginStatus method")
+		return
+	}
+
+	ticker := time.NewTicker(1 * time.Second)
```
IDisposable, line 151: Should this ping frequency be tunable (or at least less that every second)?

IDisposable (approved):
This is awesome, can't wait for this to be merged

Syakyr:
Hmm, any reason why this PR is closed? I had assumed it could be due to a seperate plugin system that was considered over this implementation, but I couldn't find any PR related to that.

tutman96:
I closed it because it's way out of date and I don't have time to update it

IDisposable:
If there's a call for it, I would be happy to bring it up to date, but not sure what the vision is of late @ym and @adamshiervani ?

Issue #11, opened 2025-01-01
Web UI should bind() to any address family (IPv4/IPv6)


apalrd:
Currently, web UI is only bind()'ing to 0.0.0.0 (IPv4 any). This prevents IPv6 connections from succeeding. Changing the bind address to [::] will allow IPv4 or IPv6 connections.

https://github.com/jetkvm/kvm/blob/6e2177e427f424774b459b67bcd14abdb5b11140/web.go#L210 appears to be the line to change for web ui (I am not sure if there are any other bind()'s in webRTC)

Nevexo:
Gin's `r.Run(":80")` should make it listen on both IP stacks, and it does if you run jetkvm_app on anything but the JetKVM.

It does work if you do
```golang
server := &http.Server{Handler: r}
l, err := net.Listen("tcp6", "[::]:80")
if err != nil {
	  fmt.Printf(err.Error())
}
err = server.Serve(l)
```

But obviously, that only listens on IPv6. Specifying "tcp" which again, should listen on v4 and v6 makes it only listen on IPv4. Needs a bit more digging as it seems like a problem with Go on this platform rather than the JetKVM software itself.

Nevexo:
I suppose it'd be possible to start two http servers, one for each stack, but that feels super wasteful when Run() is supposed to bind to all addresses.

tutman96:
For reference, it does look like pion automatically binds to ipv6 and exposes it as candidates
![image](https://github.com/user-attachments/assets/82746307-dbeb-479b-8205-710383ba938d)
![image](https://github.com/user-attachments/assets/b4c491f8-a704-45cc-baff-cdf2a799f88a)

Dropbear does not, so we should also add that as well once we have access to the os scripts

Nevexo:
> For reference, it does look like pion automatically binds to ipv6
Very good, just Gin holding us back then.

kashalls:
Do we specifically need Gin? This might be possible with the built in http Server in golang

Nevexo:
> built in http Server in golang

Yeah but the problem is with the HTTP server itself, it's that not binding to both addresses

TonyHoyle:
It's not necessary to bind to both unless golang has been altering the v6only setting.  Binding to ipv6 will automatically bind to ipv4.

Nevexo:
> It's not necessary to bind to both unless golang has been altering the v6only setting.  Binding to ipv6 will automatically bind to ipv4.
> 
> 

[...]

PR #16, opened 2025-01-03
fix(network.go): fix stuck IPv4 addresses on linkstate cycle

### LINKED ISSUE #12: Dispaly will show a previous IPv4 address if an address was assigned from a previous connection
Steps:
- Plug device into network which provides DHCPv4
- Wait for DHCPv4 to assign address and for it to display on screen
- Disconnect from network
- Plug device into a link which will not provide DHCPv4
- As Link is up, IP address page is shown, however, as IPv4 address is now blank, previous IPv4 address is shown when it is not actually assigned.

https://github.com/jetkvm/kvm/blob/6e2177e427f424774b459b67bcd14abdb5b11140/display.go#L37 seems to be the bug. IPv4 is a blank string, updateLabelIfChanged does not write new blank string, and string is not reset back to "..." after network connection is lost.

Nevexo:
This PR fixes #12 by placing the `udhcpc` DHCP client into an idle state when the network interface goes link down.

When the link transitions down, this PR will:
- Send `udhcpc` the `USR2` signal which puts it into an idle state <sup>1</sup>
- Remove all IPv4 addresses from the interface

Then once the interface comes back up, it'll send `USR1` to `udhcpc` which activates it and forces it to get a new lease.

This not only fixes the issue reported in #12 by removing the active addresses, it ensures that if the JetKVM is moved to a new subnet, it'll get an IPv4 address from the DHCP server on the new network, rather than trying to hold on to a dead lease from the previous network.

## Implementation
I'd like some feedback on the implementation of this, as it's just invoking the `killall` command with the signals, rather than using `syscall.Kill` with a signal, I avoided that method as that requires figuring out the PID of the `udhcpc` service, and I'm not sure if that's worth doing, since `killall` will always be present when jetkvm_app is running on the official buildroot image.

It may also be worth checking that the `udhcpc` process is actually running, and if it's not, is it the programs responsibility to start it?

Further work is needed on the IPv6 stack to see if #12 also effects that, I'll try this soon, but issue #11 is the more important IPv6 issue at this time.

Open to feedback, thank you!

1 - https://udhcp.busybox.net/README.udhcpc

### DIFF
--- a/network.go
+++ b/network.go
@@ -6,6 +6,7 @@ import (
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
 	"net"
+	"os/exec"
 	"time"
 
 	"github.com/vishvananda/netlink"
@@ -23,6 +24,23 @@ type LocalIpInfo struct {
 	IPv4 string
 	IPv6 string
 	MAC  string
+}
+
+// setDhcpClientState sends signals to udhcpc to change it's current mode
+// of operation. Setting active to true will force udhcpc to renew the DHCP lease.
+// Setting active to false will put udhcpc into idle mode.
+func setDhcpClientState(active bool) {
+	var signal string;
+	if active {
+		signal = "-SIGUSR1"
+	} else {
+		signal = "-SIGUSR2"
+	}
+
+	cmd := exec.Command("/usr/bin/killall", signal, "udhcpc");
+	if err := cmd.Run(); err != nil {
+		fmt.Printf("network: setDhcpClientState: failed to change udhcpc state: %s\n", err)
+	}
 }
 
 func checkNetworkState() {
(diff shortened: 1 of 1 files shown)


eric:
What is the benefit of telling the client to release the IP when the interface is down? The preference would be for it to always try to hold on to the same IP.

Nevexo:
> What is the benefit of telling the client to release the IP when the interface is down? The preference would be for it to always try to hold on to the same IP.

The way I imagined it, is if the link is going down, the user is probably moving it to a different network, so it should drop that IP address entirely and let the DHCP client find a completely new one.

> It's only going to make it take longer for the device to come back after link is restored.
If the device is switching networks, having it hold onto the IP is going to make it take much longer as the client attempts to renew it's existing IP address, until it gives up and requests a completely new one.

In my testing, it was still less than a few seconds from link-up, to pingable, but I guess if you have a slow DHCP server, or a noisy network, it could be quite a bit slower.

However, if the general consensus is that the device will probably be connected back to the same network when the link restores, then we can drop the IP removal stage and just blank out the displayed address.

Nevexo:
Fair enough, I guess it all comes down to use-case, if the device spends most of it's time sat in a rack it'll almost always come back up on the same network.

My use usually ends up with the KVM running on a USB battery bank and jumping between devices, and potentially between different management networks.

apalrd:
I think at a minimum you must try to renew the lease when the link goes back up. When link is lost, you have no idea if you are on the same network as you were before. A user in a room with a huge mess of cables could be plugging the device into each one and looking at the DHCP IPs to see if they have 'got' the right one, for example. Going from the 'network is down' screen to a screen which still shows the old IP (even if it is renewing in the background, and will eventually get the right IP) would be misleading to the user in this case.

You could skip the USR2 to release and only do USR1 on link up, which will renew the lease if it already exists, but then you don't get great feedback (since afaik you are watching Netlink and not udhcp) on when the renewal is a success to display on the screen.

Nevexo:
> since afaik you are watching Netlink and not udhcp

Yeah, it currently watches the address with netlink, so if it's still "installed" the IP will just reappear on the LCD, even if `udhcpc` is in the process of trying to get a new lease, hence why my current version deletes all IPv4 addresses on link-down. 

> You could skip the USR2 to release and only do USR1 on link up
I could probably do this anyway really, there's no real need to put `udhcpc` into idle mode now I think of it, just need to tell it to renew as soon as the link comes up.

chemhack (approved):

PR #17, opened 2025-01-03
WIP: feat(display.go): Add display brightness settings

Nevexo:
Draft PR to allow display brightness setting.

## TODO
- [x] Implement setDisplayBrightness() method
- [ ] Add brightness to config
- [ ] jsonrpc command to alter brightness
- [ ] Frontend components for brightness setting


Nevexo:
The current branch automatically dims, and then switches off, the display when configured.

However, the display doesn't re-activate when touched, so the touchscreen menu is inaccessible once the display goes into the "off" state. This needs some extra work with jetkvm_native, either backlight control needs to move into jetkvm_native, or native needs to send the application a message when the display is touched, so it can call `wakeDisplay()`

The recent changes also introduce new RPC commands to access and update the backlight config (maxbrightness, dimafter, onafter), and adds it to the settings pane of the frontend, but currently doesn't change the max brightness value due to a type miss-match between the frontend code and the jsonrpc handler.

Very much WIP, and can't be merged as it breaks the touchscreen feature at the moment.

Nevexo:
Current plan to deal with the touchscreen, at least until we have access to jetkvm_native, is to watch /dev/input/event for touch events and call wakeDisplay, as found by Tutman.

Nevexo:
Display now wakes when the touchscreen is touched, either from dimmed or entirely off thanks to @tutman96's suggestion of reading `/dev/input/events1` - needs a better solution eventually but this will do for now.

WIll work on RPC and frontend.

display.go
```diff
@@ -83,6 +94,109 @@ func updateStaticContents() {
 	updateLabelIfChanged("ui_Status_Content_Device_Id_Content_Label", GetDeviceID())
 }
 
+// setDisplayBrightness sets /sys/class/backlight/backlight/brightness to alter
+// the backlight brightness of the JetKVM hardware's display.
+func setDisplayBrightness(brightness int) error {
+	if brightness > 100 || brightness < 0 {
```
tutman96, line 100: This can actually go all the way up to 255... although at that level, it is really bright.

Nevexo: I was going to say I went with 100 because that's the default, but it's actually not, they've got it set to 64 out of the box.
Though I'm glad you brought that up because I had considered writing it as a percentage from 0-100, but since it goes to 255, that isn't right!

display.go
```diff
@@ -91,6 +205,22 @@ func init() {
 		updateStaticContents()
 		displayInited = true
 		fmt.Println("display inited")
+		wakeDisplay()
 		requestDisplayUpdate()
 	}()
+
+	go func() {
+		// Start display auto-sleeping ticker
+		ticker := time.NewTicker(1 * time.Second)
```
tutman96, line 214: Since you are already using a ticker, you might want to consider using https://pkg.go.dev/time#Ticker.Reset to just Reset it back to the DisplayDimAfterMs time whenever there is activity. Would be a bit more efficient and would mean you wouldn't need to have this no-op loop running once every second.

display.go
```diff
@@ -83,6 +94,109 @@ func updateStaticContents() {
...
+	return nil
+}
+
+// displayTimeoutTick checks the time the display was last woken, and compares that to the
+// config's displayTimeout values to decide whether or not to dim/switch off the display.
+func displayTimeoutTick() {
+	tn := time.Now()
+	td := tn.Sub(lastWakeTime).Milliseconds()
+
+	// fmt.Printf("display: tick: time since wake: %vms, dim after: %v, off after: %v\n", td, config.DisplayDimAfterMs, config.DisplayOffAfterMs)
```
Nevexo, line 126: Need to drop this line.

display.go
Nevexo: Neat, I'll do that.

display.go
tutman96: That's what things like the edid selector do... its all frontend stuff to make it happen. I don't really see needing to go above 100, and there may be power limits that it would hit, especially as the extension port starts to be used. Would need input from the hardware team on that one. To be safe, should probably just limit it to 64.

Nevexo: I'll limit it to 64 in the frontend and keep the 100 limit on this function I think. @adamshiervani / Lian your input on this would be appreciated when you get time.

display.go
```diff
@@ -83,6 +94,109 @@ func updateStaticContents() {
...
+	}
+
+	// Set the value
+	bs := []byte(strconv.Itoa(brightness))
+	err := os.WriteFile(BACKLIGHT_CONTROL_CLASS, bs, 0644)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("display: set brightness to %v", brightness)
```
eric, line 118: the convention seems to be using `log.Printf()` and not `fmt.Printf()` in this code base.

Nevexo: There's about three different logging conventions in the codebase:
image

I was mainly going off the convention in `display.go` when I decided to use fmt, I just forgot the \n on that line.

display.go
Nevexo: @tutman96 Okay, pushed this - please take a look when you get chance, I'm not super happy with having two methods that are called when the tickers fire (tick_displayDim and tick_displayOff) as they practically do the same job. I wonder if it'd be possible to keep the tickers in an array and figure out which one of them fired, to have one handler for them?

display.go
Nevexo: Hopefully a standard logging method is chosen, will refactor as part of that, just following the scheme used elsewhere in Display.go for now.

Nevexo:
The latest version of this is now available in `jetkvm-next` https://github.com/Nevexo/jetkvm-kvm/releases/tag/next-6

lwbt:
> The latest version of this is now available in `jetkvm-next` https://github.com/Nevexo/jetkvm-kvm/releases/tag/next-6

I this supposed to work like your next-5 release or do I need something else? Mine would not complete startup of the application and started to reboot after a while. I switched back to next-5.

Nevexo:
> I this supposed to work like your next-5 release or do I need something else

The config schema was updated, so that might be causing it to fall over on startup, try running jetkvm_app from `/userdata/jetkvm/bin` on it's own (you'll need to `killall jetkvm_app` first) and see what the output is.

Otherwise, edit `/userdata/kvm_config.json` and change `DisplayDimAfterMs` and `DisplayOffAfterMs` to `DisplayDimAfterSec` and `DisplayOffAfterSec` respectively. If you're in [the Discord](https://jetkvm.com/discord) it's probably easier to follow up in there, or open an issue on my repo if you can't get it going.

lwbt:
I get this error:

```
root@JetKVM:~$  ./jetkvm_app_next_6
Starting mDNS server
2025/01/27 22:14:48 USB state changed from unknown to configured
display not inited, skipping updates
2025/01/27 22:14:48 No active RPC session, skipping update state update
2025/01/27 22:14:48 Server listening on /var/run/jetkvm_ctrl.sock
2025/01/27 22:14:48 Server listening on /var/run/jetkvm_video.sock
Syncing system time
2025/01/27 22:14:48 Reconciling plugin tailscale running , should be running 0.1.0
2025/01/27 22:14:48 Started plugin tailscale version 0.1.0
2025/01/27 22:14:48 Starting process: ./jetkvm-plugin-tailscale
Binary started with PID: 292
2025/01/27 22:14:48 Accepted plugin rpc connection from @
2025/01/27 22:14:48 Plugin rpc server started. Getting supported methods...
panic: assignment to entry in nil map

goroutine 46 [running]:
kvm/internal/jsonrpc.(*JSONRPCRouter).Request(0x1ba4c60, {0x8c36c1, 0x19}, 0x0, {0x7ba600, 0x1ba2c30})
        /home/cfleming/src/jetkvm/kvm/internal/jsonrpc/router.go:53 +0x1d4
kvm/internal/plugin.(*PluginRpcServer).handleRpcStatus(0x1ba4b70, {0x11708c4, 0x1ba4c90}, 0x1ba4c60)
        /home/cfleming/src/jetkvm/kvm/internal/plugin/rpc.go:137 +0x110
created by kvm/internal/plugin.(*PluginRpcServer).handleConnection in goroutine 45
        /home/cfleming/src/jetkvm/kvm/internal/plugin/rpc.go:104 +0x150
```

> Otherwise, edit `/userdata/kvm_config.json` and change `DisplayDimAfterMs` and `DisplayOffAfterMs` to `DisplayDimAfterSec` and `DisplayOffAfterSec` respectively.

My File does not include any of this as far as I can see.

`root@JetKVM:~$  cat /userdata/kvm_config.json`:
```json
{
  "cloud_url": "https://api.jetkvm.com",
  "cloud_token": "",
  "google_identity": "",
  "jiggler_enabled": false,
  "auto_update_enabled": true,
  "include_pre_release": true,
  "hashed_password": "",
  "localAuthMode": "noPassword",
  "wake_on_lan_devices": null
}
```

Nevexo:
That's something strange going on with the plugin system, I'll have a poke around to see what's happened there, I assume I've buggered up a merge conflict.

Nevexo:
Yeah something's happening with the plugin module. You can delete the plugin from /userdata/jetkvm/plugins, but then obviously you won't have the tail scale plugin anymore, so that's not a great solution.

I'll look into it more tomorrow, but since it's not an issue with the display feature, I'll track it here: https://github.com/Nevexo/jetkvm-kvm/issues/1 - I'll ping you over there once I have a fix.

display.go
```diff
@@ -1,12 +1,24 @@
...
 
 var currentScreen = "ui_Boot_Screen"
+var backlightState = 0 // 0 - NORMAL, 1 - DIMMED, 2 - OFF
+
+var dim_ticker *time.Ticker
+var off_ticker *time.Ticker
+
+const (
+	TOUCHSCREEN_DEVICE      string = "/dev/input/event1"
+	BACKLIGHT_CONTROL_CLASS string = "/sys/class/backlight/backlight/brightness"
```
joshuasing, line 20: Effective Go and the Go stdlib usually prefers lowerCamelCase/UpperCamelCase (depending on whether exported) for constant names. E.g., per https://go.dev/wiki/CodeReviewComments:
> See https://go.dev/doc/effective_go#mixed-caps. This applies even when it breaks conventions in other languages. For example an unexported constant is maxLength not MaxLength or MAX_LENGTH.

display.go
```diff
@@ -1,12 +1,24 @@
...
+	"os"
+	"strconv"
 	"time"
 )
 
 var currentScreen = "ui_Boot_Screen"
+var backlightState = 0 // 0 - NORMAL, 1 - DIMMED, 2 - OFF
+
+var dim_ticker *time.Ticker
+var off_ticker *time.Ticker
```
joshuasing, line 16: I think these variables (and the ones above) could be merged into blocks like this:
```suggestion
var (
	dim_ticker *time.Ticker
	off_ticker *time.Ticker
)
```

Also - Effective Go and the Go stdlib usually prefers lowerCamelCase/UpperCamelCase (depending on whether exported) for names. E.g., per https://go.dev/wiki/CodeReviewComments:
> See https://go.dev/doc/effective_go#mixed-caps. This applies even when it breaks conventions in other languages. For example an unexported constant is maxLength not MaxLength or MAX_LENGTH.

Some of the function names below also contain underscores, which is not best pratice.

joshuasing (commented):
Thanks for working on this :D
I have added some comments on things that stood out to me whilst reading the diffs.

display.go
Nevexo: Thanks! Updated.

display.go
Nevexo: Thanks, shifted those into one declaration and also renamed them to `dimTicker` and `offTicker` respectively. 

Thank you for your review!

display.go
```diff
@@ -83,6 +98,156 @@ func updateStaticContents() {
...
+func tick_displayOff() {
+	err := setDisplayBrightness(0)
+	if err != nil {
+		fmt.Printf("display: failed to turn off display: %s\n", err)
+	}
+
+	offTicker.Stop()
+
+	backlightState = 2
+}
```
joshuasing, line 150: These function names contain underscores (which as mentioned in my other review, Effective Go and Go stdlib recommend using lowerCamelCase/UpperCamelCase) - Also, I think it is very uncommon to put `()` after the function name in doc comments (see https://go.dev/doc/comment):

(this suggestion does not update uses of the functions)
```suggestion
// tickDisplayDim is called when when dim ticker expires, it simply reduces the brightness
// of the display by half of the max brightness.
func tickDisplayDim() {
	err := setDisplayBrightness(config.DisplayMaxBrightness / 2)
	if err != nil {
		fmt.Printf("display: failed to dim display: %s\n", err)
	}

	dimTicker.Stop()

	backlightState = 1
}

// tickDisplayOff is called when the off ticker expires, it turns off the display
// by setting the brightness to zero.
func tickDisplayOff() {
	err := setDisplayBrightness(0)
	if err != nil {
		fmt.Printf("display: failed to turn off display: %s\n", err)
	}

	offTicker.Stop()

	backlightState = 2
}
```

Nevexo: Thanks! I'll update these afterwards.

display.go
chemhack: 100 is a pretty good choice as max. the screen get warm at max brightness of 255

Mauker1:
Question: Would you also add functionality for scheduled on/off times? It would be nice to switch the display at certain times.

Nevexo:
Potentially. Probably not in this PR? Might be worth getting this one merged in first and open that as a new one afterwards.

chemhack:
@Nevexo if you'd like to merge this one, just resolve the merge conflict then rebase WIP flag. we can do the touch input event from jetkvm_native in a separate pr.

Mauker1:
> Potentially. Probably not in this PR? Might be worth getting this one merged in first and open that as a new one afterwards.

Sweet! Thanks for the quick reply :) And thank you for this PR too!

PR #18, opened 2025-01-04
feat(ui/ActionBar): Add a CTRL + ALT + DEL Button to the Action Bar

Nevexo:
This PR introduces a CTRL + ALT + DEL button to the Action Bar which works the same way as the button on the virtual keyboard.

## Overview/Rationale
This is primarily for people installing or debugging operating systems that may need to reboot the system frequently, it avoids having to open the Virtual Keyboard as would otherwise be required to send this key combination.

The second commit in this PR makes the button a setting, which defaults to off, allowing the button to be hidden if the user doesn't need it.

## Screenshots

Button on Action Bar: (I'm not 100% set on the lock icon, suggestions welcome on that)
image

Toggle in Settings
image

### DIFF
--- a/ui/src/components/ActionBar.tsx
+++ b/ui/src/components/ActionBar.tsx
@@ -207,22 +207,23 @@ export default function Actionbar({
               onClick={() => setVirtualKeyboard(!virtualKeyboard)}
             />
           </div>
-          <div className="hidden lg:block">
-            <Button
+            {useSettingsStore().actionBarCtrlAltDel && (
+            <div className="hidden lg:block">
+              <Button
               size="XS"
               theme="light"
               text="Ctrl + Alt + Del"
               LeadingIcon={FaLock}
               onClick={() => {
                 sendKeyboardEvent(
-                  [keys["Delete"]],
-                  [modifiers["ControlLeft"], modifiers["AltLeft"]],
-                )
+                [keys["Delete"]],
+                [modifiers["ControlLeft"], modifiers["AltLeft"]],
+                );
                 setTimeout(resetKeyboardState, 100);
-              }
-              }
-            />
-          </div>
+              }}
+              />
+            </div>
+            )}
         </div>
 
         <div className="flex flex-wrap items-center gap-x-2 gap-y-2">
--- a/ui/src/components/ActionBar.tsx
+++ b/ui/src/components/ActionBar.tsx
@@ -10,12 +10,14 @@ import Container from "@components/Container";
 import { LuHardDrive, LuMaximize, LuSettings, LuSignal } from "react-icons/lu";
 import { cx } from "@/cva.config";
 import PasteModal from "@/components/popovers/PasteModal";
-import { FaKeyboard } from "react-icons/fa6";
+import { FaKeyboard, FaLock } from "react-icons/fa6";
 import WakeOnLanModal from "@/components/popovers/WakeOnLan/Index";
 import { Popover, PopoverButton, PopoverPanel } from "@headlessui/react";
 import MountPopopover from "./popovers/MountPopover";
 import { Fragment, useCallback, useRef } from "react";
 import { CommandLineIcon } from "@heroicons/react/20/solid";
+import useKeyboard from "@/hooks/useKeyboard";
+import { keys, modifiers } from "@/keyboardMappings";
 
 export default function Actionbar({
   requestFullscreen,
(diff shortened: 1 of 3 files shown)


adrianmeraz:
Love the idea. Would it be feasible to add a dropdown spinner like TinyPilot does for various shortcuts? Maybe even an option to create your own?

Example:

image

Nevexo:
> Love the idea. Would it be feasible to add a dropdown spinner[...]?

Sure, I'll have to have a chat with Adam about that as I think itd need some new React components, and frontend isn't my strong suit! 

> Maybe an option to create your own 

Now this, this I like! 

Maybe more of a macros type feature? 

I think if we do add macros thatd need to be a new PR as its a bit more involved, id personally want macros to be stored on the KVM itself so it roams between browsers (for context, the setting used in my PR is in local storage, so you need to tick it on every browser you access it from.)

adrianmeraz:
> Maybe more of a macros type feature?
> 
> I think if we do add macros thatd need to be a new PR as its a bit more involved, id personally want macros to be stored on the KVM itself so it roams between browsers (for context, the setting used in my PR is in local storage, so you need to tick it on every browser you access it from.)
> 
> Plus, we'd need some UI to record macros. Might be worth opening an issue for this specifically.

Agreed, storing them on the JetKVM is the way to go since you can move it around and keep the configs. 

Sounds good, please let me know and I can create another issue.

Nevexo:
Yeah go ahead and open an issue for macros if you don't mind, reference this pull when you do and we'll go from there - I think it's worth it's own issue and then PR. 

Thanks again

adamshiervani:
I’d honestly prefer to keep this in the Virtual Keyboard, as it’s the most logical place for it semantically. As mentioned, I agree that we should add a Keyboard Combination recorder, so people can add their own common combinations. 

@Nevexo, what’s the current issue with the placement of the Virtual Keyboard?

> Love the idea. Would it be feasible to add a dropdown spinner like TinyPilot does for various shortcuts? Maybe even an option to create your own?

I’m not sure how Tiny Pilot handles it specifically, but dropdowns & popovers, _typically_ close after a click. For many of the use cases involving these keyboard combinations, users would likely need to spam the combination repeatedly, which could make a standard dropdown less practical.

Nevexo:
> what’s the current issue with the placement of the Virtual Keyboard?

The problem isn't where it is on the Virtual Keyboard, it's the fact it's _only_ in there (this PR doesn't remove it from the Virtual Keyboard), I often find myself needing to press CTRL+ALT+DEL while debugging a system that's failing to boot, and I obviously can't press CTRL+ALT+DEL on my actual keyboard, as that'll make my OS open it's menu.

Opening the Virtual Keyboard, just to press the button, then close it again, is quite annoying (especially because opening the keyboard makes the whole page move around to accommodate a whole QWERTY kb) - though I do realise not everyone will need to press CTRL+ALT+DEL that often, hence why I added a setting for it.

adamshiervani:
I get your concern with the reaching for the Virtual Keyboard feels too heavy and annoying. The current implementation won't scale though, I'm sure there are more keyboard shortcuts people would be interested in, and this action bar + settings model won't scale. 

We can introduce a Keyboard Shortcuts action bar item, which behaves like the WOL popover. It should contain the top 5 shortcuts. This would also be a good place to create new keyboard combinations - like a record button and store it locally so if they have a unique use case they can access it easily.

We should have an explicit Close button and trigger a notification when triggering the Keyboard shortcut, to emphasize the non-closing nature of the popover.

Nevexo:
Okay sure, I like that. 

I think it'd still annoy me to still have three clicks to press CTRL+ALT+DEL once just to reboot a machine.

Maybe a "pin" option to keep a key combination on the action bar would be useful? Much more scalable than my PR, but still allows for faster access, especially if you need to see the video properly while sending a key combination, a popover is going to hide it.

jsonger:
I know I'm late to the party with this comment, but most other products in this category (PiKVM, ScreenConnect, and Bomgar are my direct references) have certain high use non-captureable keyboard combos on their toolbars within 1-2 clicks.

Ctrl+Alt+Delete is also one of the most used combos on Windows with it used to unlock and bring up the security screen.  So discounting giving space to it on the top toolbar is a big hit to usability.

IDisposable (approved):
Let's get this feature shipped independently of macros

Nevexo:
PR isn't mergeable anymore as it's way behind the head and macros are done anyway.

PR #28, opened 2025-01-05
chore: Refactor jsonrpc server into its own package

tutman96:
The JsonRPC server will be useful in more scenarios outside of the DataChannel communication with the frontend, such as for #9. This extracts it into it's own package so that it can be used for other things.


naturboy:
any idea when this will be merged?

IDisposable (approved):

IDisposable (commented):
This is going to need to be rebased.

Issue #60, opened 2025-01-10
[web] Password length limit


Nevexo:
## Overview
web.go currently doesn't have a password length limit, however, the hashing library used, bcrypt, does have a limit of 72 bytes. [1]

Implementation at: https://github.com/jetkvm/kvm/blob/main/web.go#L227

As there is currently no limit on the size of the incoming password, if it's over 72 bytes, the process fails. The frontend should have a limit, and this endpoint should reject any requests with a password that's too long with a useful error message.

## References
[1] https://en.wikipedia.org/wiki/Bcrypt#:~:text=The%20input%20to%20the%20bcrypt%20function%20is%20the%20password%20string%20(up%20to%2072%20bytes)

Nevexo:
> bcrypt truncates passwords to the first 72 bytes. In other words: they shouldn't fail

Interesting, if that's the case I'll ask the reporter to share more logs with what went wrong, as the process _failed_ for them setting a 164 character password. I don't think they have a GitHub profile so will follow up on Discord.

ItzDerock:
just my quick thoughts, what if, instead of truncating or limiting password lengths, we just pass the password first through sha-256/512, then use that output for bcrypt. This is what I've seen some other projects and frameworks do to circumvent this limitation.

For backwards compatibility, only do this for passwords that will be too long for bcrypt to normally hash. That way, existing instances will continue to use just bcrypt.

Silently truncating like what's done in the above PR feels a little misleading to the end user--they have no idea only the first 72 bytes are being used.

zandercodes:
The solution would simply be to use argon2 for password hashing
And then I no longer limit the password length

AlaskanBeard:
> > bcrypt truncates passwords to the first 72 bytes. In other words: they shouldn't fail
> 
> Interesting, if that's the case I'll ask the reporter to share more logs with what went wrong, as the process _failed_ for them setting a 164 character password. I don't think they have a GitHub profile so will follow up on Discord, and test it for myself.

I just attempted to set my KVM up for the first time using a 90 character password, which it rejected as invalid. My second attempt, with a 36 char password was also rejected as invalid, though it doesn't sound like that's related to the password length issue.

SuperQ:
[...]

PR #63, opened 2025-01-11
Update web.go

samascience:
made changes to truncate to 70 characters

### DIFF
--- a/web.go
+++ b/web.go
@@ -225,45 +225,50 @@ func handleDevice(c *gin.Context) {
 }
 
 func handleCreatePassword(c *gin.Context) {
-	LoadConfig()
+    LoadConfig()
 
-	if config.HashedPassword != "" {
-		c.JSON(http.StatusBadRequest, gin.H{"error": "Password already set"})
-		return
-	}
+    if config.HashedPassword != "" {
+        c.JSON(http.StatusBadRequest, gin.H{"error": "Password already set"})
+        return
+    }
 
 	// We only allow users with noPassword mode to set a new password
 	// Users with password mode are not allowed to set a new password without providing the old password
 	// We have a PUT endpoint for changing the password, use that instead
-	if config.LocalAuthMode != "noPassword" {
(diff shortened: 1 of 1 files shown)


Nevexo:
Thanks for your PR! Just a couple of things:

Your editor has altered the spacing of some lines, some that aren't affected by your change. Please run "go fmt" before committing code to make sure the style aligns with the rest of the project, it's always work looking at the diff before pushing it, just to make sure your editor hasn't snaked in a bad change. 

Also, please add some more info to the commit message, giving a description of what the change is makes bisecting and troubleshooting much easier in the future. It's also worth linking to the original issue (#60) in your PR, just so everyone knows what's going on. 

Finally, this kind of change needs some form of messaging on the front end explaining to the user why the password is limited to 72 bytes, it's probably worth limiting the text box field size to the limit within the backend. But don't worry if you're not comfortable altering the front end, we can do that in a different PR. 

Thanks again!

samascience (commented):
edited

samascience (commented):
added truncated changes to truncate the password to 70 characters !

samascience:
truncatedPassword - added truncated password in the code to limit to 70 characters 
a

web.go
```diff
@@ -225,45 +225,50 @@ func handleDevice(c *gin.Context) {
...
+    var req SetPasswordRequest
+    if err := c.ShouldBindJSON(&req); err != nil || req.Password == "" {
+        c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid request"})
+        return
+    }
+
+    truncatedPassword := req.Password
+    if len(truncatedPassword) > 70 {
+        truncatedPassword = truncatedPassword[:70]
+    }
```
SuperQ, line 253: This is generally the wrong way to go about this.

Rather than truncating silently, the user should get feedback that the password they entered is too long. A `http.StatusBadRequest` or [`http.StatusRequestEntityTooLarge`](https://http.cat/status/413) should be returned.

IDisposable (commented):
This should not be merged. We should not arbitrarily truncate passwords before hashing

Issue #79, opened 2025-01-15
Keyboard is not detected in BIOS, for ASUS Crosshair V Formula


mircea-vutcovici:
![Screenshot From 2025-01-15 17-17-46](https://github.com/user-attachments/assets/6dd29903-cdc7-4029-a436-ca23622bf30c)

JetKVM virtual keyboard is not detected in BIOS/UEFI firmware and can't be used until the Linux Kernel boots.
It is holding the keystrokes in the keyboard buffer and it will send all of them at one when keyboard is detected by Linux kernel.
JetKVM keyboard is not accessible in the BIOS or the GRUB boot manager.. In Linux works fine.

jcastro:
Similar problem with an HP EliteDesk 800 G3 (the USB UI at the top shows up in gray with a loading spinner, though).

Schorsch26:
Similar problems with HP Laptops:

HP EliteBook 735G5: No USB detected in BIOS mode (No Virtual Boot Drive and no input devices

HP ProBook 665 G11 JetKVM shows USB connected and the Virtual BootDrive is active but the virtual keyboard & mouse don't work.

mrkessy:
Same with my HPz440. Unable to use the keyboard functions until the kernel starts to load. 
Which is unfortunate, cuz I need the kvm hit <enter> here 

![Image](https://github.com/user-attachments/assets/398a4d8c-50b2-42e9-aa89-cb9866fde276)

robocog:
No keyboard present in BIOS for a dell optiplex 550 (till it has booted to linux or windows)
bios actually logs it as no keyboard present in its logs

I ended up using a logitech wireless keyboard to get into BIOS after trying all the usb ports on the machine with the jetkvm to see if that made a difference...which it didn't - and proves that the USB is alive and capable of keyboard support in BIOS on this machine...just not from the jetkvm

the workaround worked for this install - as I was only a few feet away from the remote machine - but obviously not much use for any distance's or actual remote stuff

Not tried any other machines yet, had hoped it was just this machine being awkward...
Bummer if this is going to be a common theme and very limiting in its use

fouadchamoun:
i have the same problem on a Dell T110ii

pieterdevriesch:
[...]

Issue #62, opened 2025-01-11
keyboard keys not sent on seabios page 


samascience:
<img width="1422" alt="image" src="https://github.com/user-attachments/assets/0f737f32-eaa7-4ab9-b8a0-b3c32c710e17" />

jalarcon666:
The same thing here on DELL servers, you cannot interact with the server until the operating system boots, no keystroke is sent to enter BIOS, configure RAID controller or any other functionality, it simply does not work.

Nevexo:
This is related to the same issue effecting HP BIOS' - it's likely because the CD drive emulation comes up even if it's not being used. A future update will allow disabling drive emulation.

At some point I'll make a new issue that combines all of these and closes them, as we have a lot of reports for the same issue.

pieterdevriesch:
Glad to hear it's on the agenda. Hopefully it will receive some priority as this is a major reason for using a KVM(my only reason), since everything in the OS can be done via remote control software normally.

PR #113, opened 2025-01-29
Enable "Boot Interface Subclass" for keyboard and mouse.

### LINKED ISSUE #56: USB Does Not Work During FileVault Stage
On my relatively ancient mac mini, the USB mouse/keyboards do not work when inputting password for getting through FileVault. Once the mac has booted (using locally-attached USB and mouse), JetKVM works.

This is on an Intel mac mini, with JetKVM version `0.3.5-dev202501022149`

andnic:
Fixes #56, #62 and #79 - Boot Protocol is (often) required for the keyboard/mouse to be recognized in BIOS/UEFI firmware.

### DIFF
--- a/usb.go
+++ b/usb.go
@@ -132,7 +132,7 @@ func writeGadgetConfig() error {
 	}
 	err = writeGadgetAttrs(hid0Path, [][]string{
 		{"protocol", "1"},
-		{"subclass", "0"},
+		{"subclass", "1"},
 		{"report_length", "8"},
 	})
 	if err != nil {
@@ -152,7 +152,7 @@ func writeGadgetConfig() error {
 	}
 	err = writeGadgetAttrs(hid1Path, [][]string{
 		{"protocol", "2"},
-		{"subclass", "0"},
+		{"subclass", "1"},
 		{"report_length", "6"},
 	})
 	if err != nil {


Nevexo:
Will include this in jetkvm-next-7 (and/or 6.3) but in the mean time, for folks looking to try this out on their hardware, I've released a binary with this fix included: https://github.com/Nevexo/jetkvm-kvm/releases/tag/usb-potential-fix-1

If you don't have a JetKVM dev env on the go, some quick instructions on how to run it: https://github.com/jetkvm/kvm/issues/79#issuecomment-2621632931

Nevexo:
Docs behind this change: https://www.usb.org/sites/default/files/documents/hid1_11.pdf

Section 4.2: Subclass.

jcastro:
I was having problems with these computers:  
1. HP EliteDesk 800 G3  
2. Chuwi Larkbox S  

I can confirm that the issue with #2 is fixed, and I'm able to access the BIOS without any problems, even with a virtual USB mounted using JetKVM functionality.

Nevexo:
Fix is included in jetkvm-next-7 https://github.com/Nevexo/jetkvm-kvm/releases/tag/next-7

wabamad:
Gave this a try for raspberry pi boot.  Keypresses are not detected until os loads

jalarcon666:
The problem persists on DELL servers when trying to send the keys to EFI so the fix is ​​not functional in this case.

rtoma:
I was hoping upgrading to the latest dev release, would resolve this. No such luck.

Tested on 0.3.8-dev202502191413, connected to a HP EliteDesk, but no luck entering the BIOS. Jet KVM UI says USB state is 'addressed'. It stays like that until OS starts, then it changes to 'configured'.

wabamad:
Same behaviour on a raspberry pi as well. No keystrokes are registered until os loads.

nhaneezy:
On 0.3.8-dev202502191413, connected to a 2012 mac mini running Ubuntu 22.04. The GRUB menu does not recognize the keyboard. It then auto boots into Ubuntu cli, which has the keyboard working with the cli.

whytf:
I'm still having the issue on prodesk g4, can't enter bios, no input is being registered in grub menu, after boot everything works

galgier:
I have both an HP 705 G5 and Lenovo M920Q with the problem.  I just installed the trial fix and the problems went away for the Lenovo.  However, I still can't send keystrokes to the HP until Linux is running.  Unfortunately I need it more for the HP (no vPRO/AMT) than the Lenovo (with vPRO/AMT).

PR #141, opened 2025-02-07
Adds Podman support to Devcontainers

josharrington:
I'm developing against Fedora which has Podman installed by default. To use a devcontainer, it was necessary to ensure SELinux was not blocking access to the kvm project files in my home directory. 

To configure the Devcontainer extension to use Podman, it must be configured by the user first in their settings.json with `"dev.containers.dockerPath": "podman"`

- Creates a podman devcontainer config
- ~Ensures dev_update.sh runs in bash as the container image uses zsh by default and it is not compatible~ No longer needed

### DIFF
--- a/.devcontainer/podman/devcontainer.json
+++ b/.devcontainer/podman/devcontainer.json
@@ -1,10 +1,19 @@
 {
-	"name": "JetKVM",
+	"name": "JetKVM Podman",
 	"image": "mcr.microsoft.com/devcontainers/go:1-1.23-bookworm",
 	"features": {
 		"ghcr.io/devcontainers/features/node:1": {
 			// Should match what is defined in ui/package.json
 			"version": "21.1.0"
 		}
+	},
+	"runArgs": [
+		"--userns=keep-id",
+		"--security-opt=label=disable",
+		"--security-opt=label=nested"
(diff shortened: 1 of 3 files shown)


IDisposable (approved):
Nice shebang !

.devcontainer/podman/devcontainer.json
```diff
@@ -0,0 +1,19 @@
+{
+	"name": "JetKVM Podman",
+	"image": "mcr.microsoft.com/devcontainers/go:1-1.23-bookworm",
+	"features": {
+		"ghcr.io/devcontainers/features/node:1": {
+			// Should match what is defined in ui/package.json
+			"version": "21.1.0"
```
IDisposable, line 7: This now needs to be 22.15.0

dev_deploy.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
IDisposable, line 1: This is no longer needed in the PR

adamshiervani:
This can be merged; just resolve the conflicts and we'll merge it in.

IDisposable:
Reimplemented in #875 because we've diverged so much

PR #152, opened 2025-02-12
Support of img.xz images on JetKVM storage mount

authrequest:
Adding support for img.xz images created by mrouillard. 
Modified usb_mass_storage.go to enable .xz images functionality utilizing "github.com/ulikunitz/xz"

### DIFF
--- a/usb_mass_storage.go
+++ b/usb_mass_storage.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/pion/webrtc/v4"
+	"github.com/ulikunitz/xz"
 )
 
 const massStorageName = "mass_storage.usb0"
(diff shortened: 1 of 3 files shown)


tutman96:
This will cause a single image to use at least 2x the emmc space to decompress it and store it when mounted. There is a partial implementation of a FUSE (userspace file system) that would enable decompressing on the fly which I think would be better suited for this device? Thoughts?

duplexsystem:
>  There is a partial implementation of a FUSE (userspace file system) that would enable decompressing on the fly which I think would be better suited for this device? Thoughts?

I think this is a better solution, there are three places compression helps:

* Ease of use: Being able to upload compressed images downloaded from distributions without having to decompress them
* Speed of image upload: Self explanatory. This is the least important one.
* Storage space used: Self explanatory. Very important as the storage space on the JetKVM is relatively small and non upgrade-able.

A FUSE library that could support transparent file access of types like .xz, and .bz2 seems like the best solution here.

Cmdr-Zod:
bz2 and xz are both compressing relatively well, but zst is becoming more and more attractive due to its better speed for the same compression ratio. I would consider to also support zst, although it can use a lot more memory, when unpacking archives with extreme compression.

usb_mass_storage.go
```diff
@@ -269,16 +270,48 @@ func rpcMountWithStorage(filename string, mode VirtualMediaMode) error {
...
+		}
+
+		currentVirtualMediaState = &VirtualMediaState{
+			Source:   Storage,
+			Mode:     mode,
+			Filename: filename,
+			Size:     fileInfo.Size(),
+		}
+
+		return nil
```
IDisposable, line 299: This entire block is a copy/paste of the next code bit, so can't we just fall-though to that after adjusting the path
```go
fullPath = decompressedPath
```

PR #153, opened 2025-02-13
Bring up loopback interface via Netlink

### LINKED ISSUE #11: Web UI should bind() to any address family (IPv4/IPv6)

apalrd:
Since the Linux system doesn't bring up loopback, and the lack of a loopback address causes the kvm app to be unable to bind to IPv6, bring up the loopback interface if it isn't already up when we bring up eth0

At least partially resolves #11

### DIFF
--- a/network.go
+++ b/network.go
@@ -25,6 +25,23 @@ type LocalIpInfo struct {
 	MAC  string
 }
 
+func checkLoopback() {
+	iface, err := netlink.LinkByName("lo")	
+	if err != nil {
+		fmt.Printf("failed to get lo interface: %v\n", err)
+		return
+	}
+
+	if iface.Attrs().OperState != netlink.OperUp {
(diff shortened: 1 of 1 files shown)


tutman96 (approved):
Fantastic. This addresses some issues with running proxying sub-processes like Tailscale as well!

chemhack:
this was just fixed in system image, closing this one

Issue #59, opened 2025-01-09
Configurable Time and NTP Server settings


joshuamfawcett:
It would be handy to have configurable NTP settings, i.e for internal servers or alternative, currently the NTP servers are hardcoded with a fixed update interval.

bk-lg:
It would be really nice if the via dhcp announced ntp servers would be used. Many companies block the external access for ntp requests. Configuring the ntp servers via the ui would be good enough if this is easier to implement.

yferszt:
NTP configuration in the UI is definitly an important feature. I have the JetKVM in a isolated VLAN (No Internet) and I provide my own NTP server.

cemysce:
https://github.com/jetkvm/kvm/issues/516 is a duplicate of this.  I'm not sure which should remain, both have useful info.

hosebei:
this is some kind of a "huge" issue. without a proper time, the update mechanism fails, but not with a proper message, it just tells you that there is no update. not ideal.
only way to achieve an update is to allow the devices connect to various ips on port 123 (NTP).

PR #162, opened 2025-02-15
feat(ntp): add delay between sync attempts and support NTP servers from DHCP

### LINKED ISSUE #74: DNS Spam Issue

ym:
Fixed #74 and partially implemented #59.

The DHCP feature requires the latest system image, but it's backwards compatible.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/gwatts/rootcerts v0.0.0-20240401182218-3ab9db955caf
 	github.com/hanwen/go-fuse/v2 v2.5.1
+	github.com/hashicorp/go-envparse v0.1.0
 	github.com/openstadia/go-usb-gadget v0.0.0-20231115171102-aebd56bbb965
 	github.com/pion/logging v0.2.2
 	github.com/pion/mdns/v2 v2.0.7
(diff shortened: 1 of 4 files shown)


Nevexo:
Great that it has a fallback if DHCP doesn't set the NTP servers. We'll need a similar thing for getting NTP configuration from DHCP6 at some point (afaik that doesn't come from RAs?)

Unrelated to this PR directly, but might also be worth considering how static addressing is going to work, we're doing quite a lot of device-level network config inside of jetkvm-app here, so do we stop the system image from starting udhcpc and have it started entirely by the app, unless there's static addressing? My concern with that one is if jetkvm-app fails to start, the device would potentially be left without networking to recover it, but at that point you're probably restoring the factory image anyway.

PR #175, opened 2025-02-19
refactor(ui): simplify backlight settings handling

adamshiervani:
Make it consistent with the rest of the settings UI


Nevexo (approved):
Yeah looks much cleaner; I think I opted for the button in the PR to avoid writing to the eMMC too often, but realistically these settings aren't going to be changed very often. Happy with this :smile:

Issue #35, opened 2025-01-06
Spacebar pauses video stream in fullscreen mode


Nevexo:
# Overview

> This bug was reported by rrobinson2000 in [the Discord server](https://canary.discord.com/channels/1270105230989594634/1325675250750328952/1325675250750328952)

When the video is fullscreen, pressing the spacebar will "pause" the playback and freeze the video. Pressing space again resumes it.

## Reproducible on Chromium Browsers
@Nevexo rrobinson2000 @azfoxd tested on Chrome, Edge & Safari

## Not reproducible on
@azfoxd Firefox

# Steps to reproduce

- Start and connect to JetKVM.
- Select Fullscreen mode.
- Press spacebar.
- Video freezes.

Expected behaviour: SPACE key sent to host, video not paused.

This issue does not happen when not in fullscreen mode.

azfoxd:
I can confirm the issue also happens on Brave and Vivaldi but not on Firefox, so it seems limited to Chromium browsers and Safari.

majerus1223:
Same issue

moebis:
same issue

Nevexo:
This problem still applies to 0.3.7

General update to this: it'll likely be fixed by #116 - but it'd still be worth looking for solutions to upstream quickly.

adamshiervani:
We just pushed a fix for Chrome to the Dev update channel. To enable the Dev Update Channel - Settings > Check the "Dev Channel Updates"

Test it out and let us know how it goes!

MikeEllis-personal:
Tested here on Chrome 133.0.6943.127 (Official Build) (64-bit) under Windows 11 and the problem appears to have been resolved.

IDisposable:
This has been fixed for quite a while, if someone notices a similar issue, please submit a **new** Issue

PR #176, opened 2025-02-19
feat(ui): improve video playback and keyboard handling in WebRTCVideo

adamshiervani:


Nevexo:
Don't have my test rig powered up to give it a try, but seems fine until we have improved keyboard handling - although I'm not even sure if that would stop it happening on every browser, so likely still worth having this.

Nevexo:
Fixes #35

PR #90, opened 2025-01-24
Allow for configuring USB IDs

jackislanding:
resolves https://github.com/jetkvm/kvm/issues/15

A new button is added to configure the USB Emulation. 

USB descriptors can now be modified, and a successful configuration rebinds the USB Device.

USB Device ID Examples:

Before:

Shows the default JetKVM Descriptors

image

After:

PR #185, opened 2025-02-22
Feature/usb config - Rasing USB Config Changes on Dev Branch

jackislanding:
This is a rebasing of PR #90 onto the dev branch.

### DIFF
--- a/ui/src/components/USBConfigDialog.tsx
+++ b/ui/src/components/USBConfigDialog.tsx
@@ -1,0 +1,216 @@
+import { GridCard } from "@/components/Card";
+import {useCallback, useEffect, useState} from "react";
+import { Button } from "@components/Button";
+import LogoBlueIcon from "@/assets/logo-blue.svg";
+import LogoWhiteIcon from "@/assets/logo-white.svg";
+import Modal from "@components/Modal";
+import { InputFieldWithLabel } from "./InputField";
+import { useJsonRpc } from "@/hooks/useJsonRpc";
+import { useUsbConfigModalStore } from "@/hooks/stores";
+import ExtLink from "@components/ExtLink";
+import { UsbConfigState } from "@/hooks/stores"
+
+export default function USBConfigDialog({
+  open,
+  setOpen,
+}: {
+  open: boolean;
+  setOpen: (open: boolean) => void;
+}) {
+  return (
+    <Modal open={open} onClose={() => setOpen(false)}>
+      <Dialog setOpen={setOpen} />
+    </Modal>
+  );
+}
+
+export function Dialog({ setOpen }: { setOpen: (open: boolean) => void }) {
+  const { modalView, setModalView } = useUsbConfigModalStore();
+  const [error, setError] = useState<string | null>(null);
+
+  const [send] = useJsonRpc();
+
+  const handleUsbConfigChange = useCallback((usbConfig: object) => {
+    send("setUsbConfig", { usbConfig }, resp => {
+      if ("error" in resp) {
+        setError(`Failed to update USB Config: ${resp.error.data || "Unknown error"}`);
+        return;
+      }
+      setModalView("updateUsbConfigSuccess");
+    });
+  }, [send, setModalView]);
+
+  return (
+    <GridCard cardClassName="relative max-w-lg mx-auto text-left pointer-events-auto dark:bg-slate-800">
+      <div className="p-10">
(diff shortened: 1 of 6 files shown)


jackislanding:
@adamshiervani @Nevexo |
Please take a look when you get the chance. Thank you

ui/src/components/USBConfigDialog.tsx
```diff
@@ -0,0 +1,216 @@
...
+            Set custom USB parameters to control how the USB device is emulated.
+            The device will rebind once the parameters are updated.
+          </p>
+          <div className="flex justify-start mt-4 text-xs text-slate-500 dark:text-slate-400">
+            <ExtLink
+              href={`https://the-sz.com/products/usbid/index.php?v=${usbConfigState.vendor_id}&p=${usbConfigState.product_id}`}
+              className="hover:underline"
+            >
+              Look up USB Device IDs here
+            </ExtLink>
```
adamshiervani, line 137: The query parameters don't really make sense for me. I would suggest removing them and just let the user find USB device IDs in a general sense.

ui/src/components/sidebar/settings.tsx
```diff
@@ -952,7 +1072,43 @@ export default function SettingsSidebar() {
...
+                  size="SM"
+                  theme="light"
+                  text="Update USB Config"
+                  onClick={() => {
+                    setUsbConfigModalView("updateUsbConfig")
+                    setIsUsbConfigDialogOpen(true);
+                  }}
+                />
+              </SettingsItem>
+            )}
```
adamshiervani, line 1111: No need to add this under the developer mode flag. 

Additionally, a better place for this would be under the Hardware section, where we now have the display settings.

adamshiervani (changes requested):
Awesome work! Just need to address the two comments.

ui/src/components/sidebar/settings.tsx
jackstown: @adamshiervani moved the USB config settings

ui/src/components/USBConfigDialog.tsx
jackstown: @adamshiervani done, query params removed

ui/src/components/sidebar/settings.tsx
```diff
@@ -113,12 +130,90 @@ export default function SettingsSidebar() {
     });
   }, [send]);
 
+  const syncUsbConfigProduct = useCallback(() => {
+    send("getUsbConfig", {}, resp => {
+      if ("error" in resp) {
+        console.error("Failed to load USB Config:", resp.error);
+      } else {
+        console.error("syncUsbConfigProduct#getUsbConfig result:", resp.result);
```
adamshiervani, line 138: console.log

ui/src/components/sidebar/settings.tsx
```diff
@@ -954,6 +1074,41 @@ export default function SettingsSidebar() {
...
+              if (e.target.value === "custom") {
+                setUsbConfigProduct(e.target.value);
+              } else {
+                handleUsbConfigChange(e.target.value as string);
+              }
+            }}
+            options={[...usbConfigs, { value: "custom", label: "Custom" }]}
+          />
+        </SettingsItem>
+        {usbConfigProduct === "custom" && (
```
adamshiervani, line 1096: This doesn't work if you set to `custom` and then reload the page. If you set the Custom Product value to `X` and then reload, usbConfigProduct will be `X` and this won't render - showing `JetKVM USB Emulation Device`.

A solution could be to check for `custom` and if it's not included in the `usbConfigs`.

adamshiervani (changes requested):
Second round

ui/src/components/sidebar/settings.tsx
jackislanding: @adamshiervani done, I now add a check to set the product name as "custom" if the saved product name is not in the usb config product names. 

Now when setting the product name to a custom value and saving, a page reload results in the spinner being set to custom.

I understand that technically any change to a default usb config is "custom", but that would result in maybe having to make a hash of all the properties of a usb config object to check equivalency.

ui/src/components/sidebar/settings.tsx
jackislanding: fixed

adamshiervani (approved):

PR #189, opened 2025-02-24
feat(ci): GitHub action for pull requests - Go + NodeJS

timothystewart6:
A pull request action that will build the Go app and build the NodeJS app to ensure no one breaks the build.   It also caches dependencies for both go and NodeJS.

This would also be a good place to run linting and tests however it hasn't been run in the past so there's a lot to clean up.  Added it but commented it out.

I already added this to my fork and you can see the output here:

https://github.com/timothystewart6/kvm/actions/runs/13490998649

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,0 +1,75 @@
+name: 'Pull Request'
+on:
+  pull_request:
+    branches:
+      - dev
+    paths-ignore:
+      - .gitignore
+      - README.md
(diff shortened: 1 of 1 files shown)


timothystewart6:
After this is merged it should be added to the pull request status checks so it enforces the build to pass.

IDisposable (approved):

adamshiervani:
Thanks, Tim. We had to implement our own custom solution using a testing suite in a private repo in order to run tests on real devices.

PR #194, opened 2025-02-26
refactor(modals): Move modals to URLs

adamshiervani:
As the setting sidebar is becoming unusable, we will need to move it to a modal. If we do that, we need a proper way to do modals. Route-based modals moves a lot of the state management to the URL and allows us to use react-router framework properly with e.g. error handling. 

In this PR, I didn't change the underlying state management of each modal, but simply moved them into a route retaining all the state management it had before, except the logic for whether is supposed to be shown or not - that's offloaded to the URL now.


adamshiervani:
All is finished, but I'll remove from draft mode once https://github.com/jetkvm/kvm/pull/185 is merged, and I've rebased on top of it

Issue #83, opened 2025-01-18
Ability to Rename a JetKVM


beornlake:
It would be helpful to have the ability to "rename" a JetKVM and have that name appear in the browser tab and at the top of the UI so that we can see at a glance which machine a JetKVM is attached to. This would be especially useful in differentiating between browser tabs when multiple JetKVM devices are in use.

basisau:
Upvoting this, and also requesting an extension.

One of JetKVM's USPs is that with multiple devices all consoles are live simultaneously. However without at least a name on each browser tab, navigating between tabs can be like flying blind.

Also, because JetKVMs themselves are assets to be managed, I'd also request a footer string on the browser page, perhaps centre-aligned:

1. Device name / FQDN
2. Device App and System versions - would be very useful for quick audit of upgrade status
3. IPv4, and IPv6 if enabled - very useful if running browser in Chrome Application mode (no address bar)
4. Additionally, IP assignment mode e.g. Static (when 0.5 is released) or DHCP

Example:

jkvm1.domain.local - App 0.4.8 / System 0.2.5 - IPv4 10.1.3.121 (DHCP) - MAC 30:52:53:02:5c:c6

If this is too wide, then just FQDN, IPv4 and IPv6 would be ideal.

PR #196, opened 2025-02-26
Feature/set jetkvm name

jackislanding:
Resolves Issue #83

Setting is moved under "Appearance" section

image

Device Name appears on browser tab

image

Reloading page loads device name again, since it is stored in settings store

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/creack/pty v1.1.23
 	github.com/gin-gonic/gin v1.9.1
 	github.com/google/uuid v1.6.0
+	github.com/gosimple/slug v1.15.0
 	github.com/gwatts/rootcerts v0.0.0-20240401182218-3ab9db955caf
 	github.com/hanwen/go-fuse/v2 v2.5.1
 	github.com/hashicorp/go-envparse v0.1.0
@@ -42,7 +43,6 @@ require (
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.20.0 // indirect
 	github.com/goccy/go-json v0.10.2 // indirect
-	github.com/gosimple/slug v1.15.0 // indirect
 	github.com/gosimple/unidecode v1.0.1 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/cpuid/v2 v2.2.7 // indirect
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,8 @@ require (
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.20.0 // indirect
 	github.com/goccy/go-json v0.10.2 // indirect
+	github.com/gosimple/slug v1.15.0 // indirect
+	github.com/gosimple/unidecode v1.0.1 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/cpuid/v2 v2.2.7 // indirect
 	github.com/kr/pretty v0.3.0 // indirect
--- a/network.go
+++ b/network.go
@@ -124,7 +124,9 @@ func startMDNS() error {
 	}
 
 	// Start a new server
-	fmt.Printf("Starting mDNS server on jetkvm.local\n")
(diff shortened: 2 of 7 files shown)


SuperQ:
Should this be filtered to DNS hostname validity? This way we could use this value for mDNS announcements.

jackislanding:
> Should this be filtered to DNS hostname validity? This way we could use this value for mDNS announcements.

Sure, I can add a regex to filter the value

SuperQ:
Maybe we can use https://pkg.go.dev/github.com/asaskevich/govalidator?#IsDNSName

chemhack:
what we could also do is having the device name and dns name as two separate fields, with automatically filled dns name with slugify. thus you can make mDNS happy without limiting your choice of names.

SuperQ:
That's a good idea, something like https://github.com/gosimple/slug.

jackstown:
Made progress on this. As seen in the logs, the name can be set and the dns name is slugified:

image

jackstown:
This is what the setting looks like. Will add a field  for the current dns field, which will require returning the json of the name config on setting the value.

jackstown:
image

jackstown:
On success, a notification appears
image

go.mod
```diff
@@ -42,6 +42,8 @@ require (
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.20.0 // indirect
 	github.com/goccy/go-json v0.10.2 // indirect
+	github.com/gosimple/slug v1.15.0 // indirect
```
SuperQ, line 45: You probably need to `go mod tidy`.

network.go
```diff
@@ -146,7 +155,8 @@ func startMDNS() error {
 	}
 
 	mDNSConn, err = mdns.Server(ipv4.NewPacketConn(l4), ipv6.NewPacketConn(l6), &mdns.Config{
-		LocalNames: []string{"jetkvm.local"}, //TODO: make it configurable
+		LocalNames: []string{config.NameConfig.DNS},
+		//LocalNames: []string{"jetkvm.local"}, //TODO: make it configurable
```
SuperQ, line 159:
```suggestion
```

jackislanding:
@chemhack I reformatted the settings.tsx since it always seems to get formatted weird.

PR is now ready for review, thank you

ohmantics:
In addition to mDNS, it would be ideal if this also called through to hostname and then called setDhcpClientState in network.go to get the name known by DHCP servers many of which will add that name to local DNS.

network.go
Balnian: The `.local` should probably be kept/appended here as it is the proper TLD reserved for local DNS address (and specifically mDNS)
https://en.wikipedia.org/wiki/.local

network.go
```diff
@@ -146,7 +154,7 @@ func startMDNS() error {
 	}
 
 	mDNSConn, err = mdns.Server(ipv4.NewPacketConn(l4), ipv6.NewPacketConn(l6), &mdns.Config{
-		LocalNames: []string{"jetkvm.local"}, //TODO: make it configurable
+		LocalNames: []string{config.NameConfig.DNS},
```
Balnian, line 157: The `.local` should probably be kept/appended here as it is the proper TLD reserved for local DNS address (and specifically mDNS)
https://en.wikipedia.org/wiki/.local

jackislanding: Append ".local" to the mDNS domain? I can do that if that is what you mean

jackislanding:
> In addition to mDNS, it would be ideal if this also called through to hostname and then called setDhcpClientState in network.go to get the name known by DHCP servers many of which will add that name to local DNS.

Thank you for the suggestion. I think that can be added as a separate issue after this PR gets merged.

network.go
Balnian: Just noticed the `config.NameConfig.DNS` already has `.local` appended  to it. Sorry for the noise!
https://github.com/jetkvm/kvm/pull/196/files#diff-4ce98339156c9db92c77843a1c1b6256deb9c8219a44d37a1e6d2cc5e525a5a6R311

Balnian: Exactly! Solid work @jackislanding !

jackislanding:
Rebased changes from this PR on dev. PR #230  supersedes this PR

PR #201, opened 2025-02-26
Add more public NTP servers

IDisposable:
Using the [this comprehensive list](https://gist.github.com/mutin-sa/eea1c396b1e610a2da1e5550d94b0453)

- Added pool.ntp.org at the top since it's globally distributed and isolated from specific manufacturers (and is what most routers default use)
- Added the other major network providers Google and AWS
- Added Windows to pair with Apple

### DIFF
--- a/ntp.go
+++ b/ntp.go
@@ -22,8 +22,12 @@ const (
 var (
 	timeSyncRetryInterval = 0 * time.Second
 	defaultNTPServers     = []string{
+		"pool.ntp.org",
 		"time.cloudflare.com",
+		"time.google.com",
+		"time.aws.com",
 		"time.apple.com",
+		"time.windows.com",
 	}
 )
 
--- a/ntp.go
+++ b/ntp.go
@@ -24,10 +24,6 @@ var (
 	defaultNTPServers     = []string{
 		"pool.ntp.org",
 		"time.cloudflare.com",
-		"time.google.com",
-		"time.aws.com",
-		"time.apple.com",
-		"time.windows.com",
 	}
 )
 
@@ -98,8 +94,9 @@ func queryNetworkTime() (*time.Time, error) {
 		}
 	}
 	httpUrls := []string{
+		"http://pool.ntp.org",
 		"http://apple.com",
-		"http://cloudflare.com",
+		"http://microsoft.com",
 	}
 	for _, url := range httpUrls {
 		now, err := queryHttpTime(url, timeSyncTimeout)


timothystewart6:
it would be nice if there were an option to add your own vs a hard coded list.  I understand why people choose `pool.ntp.org` but it's not my top choice.

IDisposable:
To be clear, the default list is only used if the DHCP doesn't offer one... so most people are going to sync against their own router.

duplexsystem:
> it would be nice if there were an option to add your own vs a hard coded list. I understand why people choose `pool.ntp.org` but it's not my top choice.

I agree, some people may not have DHCP accessible on the subnet/vlan the JetKVM is on, but do have their own time server they want used.

timothystewart6:
@IDisposable understood, I just mean since it's a preference, it should be treated as such and exposed for people to choose or better yet, type.

IDisposable:
I totally agree, would be nice to expand the UI and preferences (we would still need the defaults until set :) ) and I would be happy to build that out if desired.

timothystewart6:
I am not sure who makes that call but I assume most people would appreciate it.  😅

ntp.go
```diff
@@ -24,8 +24,12 @@ var (
 	timeSynced            = false
 	timeSyncRetryInterval = 0 * time.Second
 	defaultNTPServers     = []string{
+		"pool.ntp.org",
 		"time.cloudflare.com",
+		"time.google.com",
```
SuperQ, line 29: I would avoid adding time.google.com mixed with these other sources. This time source does leap second smear, which will not play well with non-smear time sources.

SuperQ (changes requested):
I would probably remove everything except pool.ntp.org.

ntp.go
IDisposable: That's a good point, I doubt we'll ever get to there since presumably poot.ntp.org will always return (and google is way down...) but happen to remove it. I think there are _also_ get smeared returns from others available, but in this application, we're really not concerned with leap-second accuracy ;)

IDisposable: Who should make that canonical call? I'm happy to adjust if someone tells me what to use... and mine is on order, so hopefully it'll show soon :)

timothystewart6:
IMO since this is really a user preference, I would advocate that we not change this until it's adjustable in the UI.  It seems like any change to the existing codebase will not satisfy everyone and I feel changing it now for your preference would only add to churn.

PR #207, opened 2025-02-28
feat(cloud): Add support for custom cloud app URL configuration

adamshiervani:
- Removes all environment variables building in `device` mode
- Add Cloud app url to device config
- Simplifies using alternative cloud providers

image

cc @timothystewart6


timothystewart6 (approved):
This is awesome, it means the we'll no longer need build with variables!

Issue #215, opened 2025-03-01
Use structured logging


SuperQ:
It would be nice to switch the logging from basic Go `log` to the new `log/slog` package. This would allow for both logfmt and json formatted logs. Having structured key=value logs is nicer to read than `fmt.Sprintf()` strings.

Also, it would be good to cleanup any use of `fmt.Print(f|ln)`.

SuperQ:
It seems like `github.com/pion/logging` is the current main logging library. It does not support structured logs.

Nevexo:
It'd also be nice in the future to be able to stream logs to syslog etc.

SuperQ:
I think syslog and related would be something we would need to file against the system distro (https://github.com/jetkvm/rv1106-system).

PR #7, opened 2024-12-31
Chore: Enable golangci-lint

SuperQ:
Add a GitHub actions workflow to run golangci-lint.
* Fixup/ignore various linting issues.
* Make the `kvm` package a fully-qualified public package.

PR #216, opened 2025-03-01
Cleanup logging

SuperQ:
Make sure all logging output is called via the main logger instead of stdlib `"log"` or `fmt.Print(f|ln)`.


SuperQ:
It would be nice to get https://github.com/jetkvm/kvm/pull/7 merged. Then we can enforce this via linter settings.

SuperQ:
Somewhat related to https://github.com/jetkvm/kvm/issues/215.

Nevexo:
As this touches quite a few files and will become rebasing-hell quite quickly, a review of this would be great once you folks get time.

SuperQ:
> We should probably standardise the way messages are written

My idea is to switch logging to `log/slog` and use structured tagging instead of strings.

Within each feature you can do `featureLogger := logger.With("subsystem", "feature")`

SuperQ:
Maybe ping @ym @chemhack?

ym (approved):

PR #230, opened 2025-03-06
Allow for setting device name and mDNS (Rebasing PR #196 on Dev)

jackislanding:
Turns out it was just easier to rebase PR #196  on dev (w/ another branch) which contains the latest UI on dev branch (shoutout to @adamshiervani , look great! Easier to code for)

Here are the changes with the new UI:

image

### DIFF
--- a/jsonrpc.go
+++ b/jsonrpc.go
@@ -300,12 +300,10 @@ type SSHKeyState struct {
 }
 
 func rpcGetNameConfig() (NameConfig, error) {
-	LoadConfig()
 	return config.NameConfig, nil
 }
 
 func rpcSetNameConfig(deviceName string) (NameConfig, error) {
-	LoadConfig()
 	config.NameConfig = NameConfig{
 		Name: deviceName,
 		DNS:  slug.Make(deviceName) + ".local",
--- a/jsonrpc.go
+++ b/jsonrpc.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/gosimple/slug"
 	"github.com/pion/webrtc/v4"
 	"go.bug.st/serial"
 )
@@ -41,6 +42,11 @@ type BacklightSettings struct {
 	MaxBrightness int `json:"max_brightness"`
 	DimAfter      int `json:"dim_after"`
 	OffAfter      int `json:"off_after"`
+}
+
+type NameSettings struct {
+	Name string `json:"name"`
+	DNS  string `json:"dns"`
 }
 
 func writeJSONRPCResponse(response JSONRPCResponse, session *Session) {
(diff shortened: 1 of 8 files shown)


SuperQ (approved):

SuperQ (approved):

jsonrpc.go
```diff
@@ -293,6 +299,28 @@ type SSHKeyState struct {
...
 }
 
+func rpcGetNameConfig() (NameConfig, error) {
+	return config.NameConfig, nil
+}
+
+func rpcSetNameConfig(deviceName string) (NameConfig, error) {
+	config.NameConfig = NameConfig{
+		Name: deviceName,
+		DNS:  slug.Make(deviceName) + ".local",
```
IDisposable, line 309: Would be nicer to _not_ force the DNS name to always be computed as `.local` domain... could we do that if they don't give a DNS name (to default to a reasonable value), but honor whatever they typed.. then allow that text to be entered on the UI.

jsonrpc.go
```diff
@@ -293,6 +299,28 @@ type SSHKeyState struct {
...
+	}
+
+	RestartMDNS()
```
IDisposable, line 312: Should this happen _after_ the `SaveConfig()` succeeds?

network.go
```diff
@@ -113,6 +113,13 @@ func checkNetworkState() {
 	}
 }
 
+func RestartMDNS() {
+	err := startMDNS()
+	if err != nil {
+		return
```
IDisposable, line 119: Do we care if the MDNS doesn't restart? Should we be checking that before calling `SaveConfig()` above? Seems like we're just eating the error here but carrying on with the save there (error or not).

ui/src/routes/devices.$id.settings.appearance.tsx
```diff
@@ -48,6 +77,24 @@ export default function SettingsAppearanceRoute() {
...
+      <SettingsItem title="Device Name" description="Set your device name">
+        <InputFieldWithLabel
+          required
+          label=""
+          placeholder="Enter Device Name"
+          description={`DNS: ${nameConfigSettings.dns}`}
+          defaultValue={nameConfigSettings.name}
+          onChange={e => handleNameChange(e.target.value)}
+        />
+      </SettingsItem>
```
IDisposable, line 89: Add ability to edit DNS with something like this?
```tsx
 <SettingsItem title="DNS Name" description="Set the DNS name for the device">
    <InputFieldWithLabel
      required
      label=""
      placeholder="Enter DNS Name"
      defaultValue={nameConfigSettings.dns}
      onChange={e => handleDNSChange(e.target.value)}
    />
  </SettingsItem>
```
If we leave off the `required` then they could leave it blank to allow the API to compute the default suffix of `'.local'` or we could just do that in the `setDNS` state manager.

```tsx
const [dns, setDNS] = useState("");

const handleDNSChange = (value: string) => {
  // to do client-side defaulting uncomment
  // if (value.len() == 0) {
  //   value = name + ".local"
  // }
  setDNS(value);
};
```

IDisposable (commented):
Looks pretty solid, perhaps more freedom to supply the full DNS name?

adamshiervani:
This has now been implemented and launched in [0.4.0](https://github.com/jetkvm/kvm/releases/tag/release%2F0.4.0) through the new Network settings page.

PR #242, opened 2025-03-09
feat: Add scroll sensitivity configuration and improved wheel event h…

adamshiervani:
Improve scrolling and add a setting for slow, default and High

Added a new store called DeviceSettingsStore. We should migrate all settings there and get all the relevant ones on page load. Currently, some settings are in Localstorage, and some stored on device - causes both confusion and bugs. We should move all the settings to the DeviceSettingsStore moving forward. Didn't want to blow this PR up, so migrating other settings will be done later.


ym (approved):
LGTM

PR #250, opened 2025-03-11
fix: Logging cleanup

SuperQ:
* Cleanup additional `fmt.Println()` that should call logger.
* Use `%v` for logging errors.


SuperQ:
@ym Followup to https://github.com/jetkvm/kvm/pull/216

ym (approved):

PR #299, opened 2025-03-26
feat: disable metrics endpoint by default

ym:


SuperQ:
Please do not do this. There is no reason to disable the metrics endpoint by default.

ym:
> Please do not do this. There is no reason to disable the metrics endpoint by default.

The metrics endpoint is currently unauthenticated, and please don't get me wrong, I have no intention of disabling or removing it, the goal is to make it more secure. Right now it's a security risk that we can't ignore as we continue to add more metrics. But for now, I'm going to move this PR to draft since the UI and password part is not yet finished.

SuperQ:
Sorry, but your assumption is incorrect. There is no security risk with metrics.

There is nothing "more secure" by disabling metrics.

If there is a security issue, please describe it in detail. Otherwise you are making statements without basis.

IDisposable:
Circling back to this, I am beginning to see the metrics GATHERING consume a significant slice of the device's CPU...we might be well served to allow folks to opt into (or out of) that overhead.

SuperQ:
@IDisposable Do you have a [pprof](https://pprof.me/) output that I can see? Gathering should be very low overhead. It would be interesting to know what's going on.

Also, Gathering metrics only happens when a user is _using_ metrics. So, opting out would disable intentionally gathered metrics. That doesn't make a lot of sense.

Issue #290, opened 2025-03-25
Jiggler "jiggles" whether the mouse is used or not


lwsrbrts:
## Versions
App: 0.3.8
System: 0.2.3

## Issue
Since upgrading to the latest App version, I've found that when the jiggler is enabled, even during active use of the mouse, the jiggler is sending jiggles to the device causing windows to be repositioned unexpectedly.

## Reproduction steps
To reproduce on Windows, enable the jiggler and open an Explorer window. Left-click and select the title bar as though to reposition the window, you may need to continue to keep the left mouse button clicked and move around the window (which should mean the jiggler is not activated). After some period of time, less than 30 seconds, the window will get repositioned/maximised etc. to the top left (0,0).

## Why is this an issue?
For me, this could happen during a file copy drag/drop operation and cause the file to be copied/moved to the wrong location but it's also quite disorienting when moving a window causes it to be resized or pinned.

## Workaround
The workaround is to only enable jiggler when it's needed and no active input is expected but this isn't convenient.

jackislanding:
I'll take a look at this. I noticed that the lastUsed variable is never assigned again once it is initialized.

jackislanding:
@lwsrbrts see https://github.com/jetkvm/kvm/pull/316

lwsrbrts:
Great! Thanks for looking in to this - definitely going to be a good improvement to have that much more control. 👍

lwsrbrts:
Still see this issue on App: 0.4.5 and System: 0.2.5.

PR #316, opened 2025-04-02
Added crontab scheduler for jiggler

jackislanding:
Explanation of fields:

- Schedule 
- - Crontab that allows user to set the schedule of when the jiggler is active
- Jitter Percentage
- - Variance to the jiggler schedule.
- - Example: Jiggler is set to fire every 30 seconds. If using a jitter of 50%, this adds +/- 50% of 30 seconds, resulting in a random range from 15 - 45 seconds.
- Inactivity Limit
- - How long the mouse/keyboard can be inactive before the jiggler starts triggering.

**Note:** The jiggler was broken since the lastUserInput variable was never reset, resulting in the jiggler firing nonstop after 20 seconds. This is now fixed by using the built-in gadget GetLastUserInputTime function.

Jiggler Config View:

image

Custom Crontab Schedule:

image

Custom Jitter Percentage:

image

Notification when setting the Jigger Config Successfully:

image


adrianmeraz:
> Any update on this? The current jiggler impl is lacking. Also would it be possible to have the jiggle be relative to where the cursor is instead of using absolute positioning?
> 
> Or maybe another option is to send a keyboard key press and release instead

Sorry, there's not an update. 

The issue is the PRs are left to sit for a long time, and then we have to rebase on dev constantly, and hope someone eventually reviews the PR.

Hopefully, the JetKVM team implements this, but I wouldn't hold my breath.

lwsrbrts:
I'd settle for the issue in #290 being resolved but this would be _much_ better!

When selecting text, the fact the jiggler jiggles when it shouldn't is really beginning to grind my gears.

adamshiervani:
Looks good! I’d suggest not exposing the underlying data model to the user initially. Similar to how we handle USB device and EDID-settings, we should present a simple dropdown with common Jiggler options while also providing an escape hatch for advanced users who understand the underlying details and want full control.

I don’t have any great suggestions for the preset names right now - maybe someone else can chime in - but offering three standard options plus a ‘Custom’ aligns well with the overall JetKVM UI philosophy.

mszczewski:
Perhaps options like 1 hour, 4 hours, 8 hours, and the cron configuration?

dakjr:
Any progress on getting this merged and incorporated?  The current mouse jiggler is SEVERELY lacking and frustrating.

adamshiervani:
Added this to be scheduled for the next release. I will pull the PR and just make some sensible default decisions and merge it with `dev`.

ym (approved):

PR #324, opened 2025-04-03
feat: Add support for remove KVM switches

kirillmakhonin-brt:
This PR adds feature to setup remove KVM switch so JetKVM can be remotely switch source inputs.

New features:

1. Adds section in the Settings interface to enable remote KVM switch and manage connections.
image

2. Each connection can be HTTP(s), TCP or UDP. For TCP or UDP messages can be set in HEX, ASCII or Base64

3. Adds ability to select channel from the main interface
image

### DIFF
--- a/config.go
+++ b/config.go
@@ -34,6 +34,10 @@ type Config struct {
 	TLSMode              string             `json:"tls_mode"`
 	UsbConfig            *usbgadget.Config  `json:"usb_config"`
 	UsbDevices           *usbgadget.Devices `json:"usb_devices"`
+	// Remote KVM
+	RemoteKvmEnabled         bool            `json:"remote_kvm_enabled"`
+	RemoteKvmSelectedChannel string          `json:"remote_kvm_selected_channel"`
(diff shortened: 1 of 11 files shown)


IDisposable:
This is very well laid out, can't wait for it to be merged so I can add on the serial port support I'm working on for this [MLEEDA 10 Port  with RS-232 Wired Remote](https://www.amazon.com/dp/B0DL2DB1XJ)

kirillmakhonin-brt:
@ym  could you please check this and let me know if you are willing to merge it? I can resolve conflicts for sure but I'm worried that I did not get any feedback on this PR for a last month

adamshiervani:
This isn’t something we want to support right now. We’re about to implement a feature tracker, and if there’s more demand for this in the future, we can consider it then. Once we settle on a plugin system, this could even be a plugin.

That said, the JetKVM UI generally prioritizes ease of use with optional customization, as seen in the EDID and USB config settings, where we offer common presets alongside customizable options for people who know what they want. If this feature were to be added to the UI, it should follow the same principle.

PR #331, opened 2025-04-07
feat: cursor and keyboard lock

ym:


kirillmakhonin:
@ym looks like this PR contains pieces from #330

ym:
> @ym looks like this PR contains pieces from #330

Yes, that's true because the feature requires TLS to work. I'll rebase it once the TLS PR is merged.

PR #359, opened 2025-04-15
feat: set hardware clock after NTP sync, fix#319

tucksaun:
Not synchronising the hardware clock causes the time to be lost on next reboot, making features such as OTA updates check fail.

### DIFF
--- a/ntp.go
+++ b/ntp.go
@@ -193,5 +193,9 @@ func setSystemTime(now time.Time) error {
 	if err != nil {
 		return fmt.Errorf("failed to run date -s: %w, %s", err, string(output))
 	}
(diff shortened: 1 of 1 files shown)


ym:
Thanks so much for your contribution! I'm happy to let you know that this feature has already been implemented in our latest dev branch (you can see it [here](https://github.com/jetkvm/kvm/blob/189b84380b6e9b54166c897f2d5e2592a01dca1e/internal/timesync/rtc_linux.go)).

Since we already have this covered, I'll go ahead and close this PR. We really appreciate your interest in improving the project, and we'd love to see more contributions from you in the future!

Issue #313, opened 2025-04-01
Not able to send "@" or "\" to the client


bambi-is-evil:
Hi there,

just received my JetKVM and it works really great. One issue I have is if I want to use "@" or "/", that are really important to me. Since I'm german, I'm working with a german keyboard and layout. This keys have to be pressed together with the key "Alt GR" in this layout. This just does not work with JetKVM. Also virtual keybord and c&p do not work. It is really important for me to get this issue solved or a workaround.

Sorry for my bad English and thanks in advance!

TrevorDunne:
Same here cannot type \ on English keyboard layout # appears instead. Remote desktop is fine, I  tried different classes of keyboard under hardware still not working.

williamjohnstone:
#116 will resolve this issue, for now try to use Alt codes or switch your host machines keyboard layout to US English.

dlorch:
Alt-Gr issue fixed in #399.

dlorch:
German keyboard layout for "Paste text" available in #405.

Issue #118, opened 2025-01-30
Cannot type characters with AltGr key


puigru:
Hello! I just received my JetKVM and I'm pretty pleased with it. However, I'm running into a bit of a bothersome issue with regards to keyboard layout. I can't type certain characters like "@" or "#", which in my Spanish keyboard are entered by pressing AltGr + a number key. I just end up typing the number instead of the special character on the receiving end.

This is especially annoying as you need "@" to type e-mails or mention people in many chat clients, which forces me to resort to using Windows' virtual keyboard. JetKVMs' built-in doesn't cut it because, the layout is wrong there too and I can't seem to find a way to change it.

![Image](https://github.com/user-attachments/assets/8bba775a-0dc3-453d-ad12-051029e336da)
This is my keyboard layout (characters in blue are typed in combination with AltGr)

JetKVM recognises the key combination of AltGr + 2 (to type @), as such:
![Image](https://github.com/user-attachments/assets/6d5af690-c199-4ab6-bbd8-5f7b7f8fd71b)

Nevexo:
Unfortunately, JetKVM currently only supports the US keymap.

Issue #30 in progress as PR #116 

Doesn't really matter what your "controlling" machine has it's keymap set to, the JetKVM tries to type them on it's US keyboard. We'll hopefully be able to test out the wonderful #116 soon and get this resolved.

dlorch:
Fixed in #399 and #405.

PR #399, opened 2025-04-30
Fix: Alt Gr not recognized

### LINKED ISSUE #30: Can't write/paste certain special characters using key combinations
I'm not sure if it's related to me using a non-US keyboard layout (AZERTY, both on host and the remote computer) but I can't seem to be able to write/paste certain special characters using key combinations such as AltGr+3 for '#' or AltGr+0 for '@', it instead just outputs numbers or the default non-combination character.

I have included a GIF of the issue below.

Browser: Firefox 133.0.3 on Windows 10
JetKVM App: 0.3.4
JetKVM System: 0.2.0
Remote computer: Linux Mint 21.3 Cinnamon

firefox_4LqMd68tEF

dlorch:
This PR fixes the Alt Graph key not being recognized.

Provides fixes for #30, #118, #313.

This can be tested by adding following log statements to `internal/usbgadget/hid_keyboard.go`:

```golang
func (u *UsbGadget) keyboardWriteHidFile(data []byte) error {
        if u.keyboardHidFile == nil {
                var err error
                u.keyboardHidFile, err = os.OpenFile("/dev/hidg0", os.O_RDWR, 0666)
                if err != nil {
                        return fmt.Errorf("failed to open hidg0: %w", err)
                }
        }

        fmt.Println("-- hid_keyboard --")
        fmt.Printf("modf: 0x%x\n", data[0])
        fmt.Printf("key1: 0x%x\n", data[2])
        fmt.Printf("key2: 0x%x\n", data[3])
        fmt.Printf("key3: 0x%x\n", data[4])
        fmt.Printf("key4: 0x%x\n", data[5])
        fmt.Printf("key5: 0x%x\n", data[6])
        fmt.Printf("key6: 0x%x\n", data[7])
        fmt.Println("------------------")

        _, err := u.keyboardHidFile.Write(data)
        if err != nil {
                u.log.Error().Err(err).Msg("failed to write to hidg0")
                u.keyboardHidFile.Close()
                u.keyboardHidFile = nil
                return err
        }

        return nil
}
```

The correct output after applying this PR for pressing Alt-Gr + 2 is:

```
-- hid_keyboard --
modf: 0x40
key1: 0x1f
key2: 0x0
key3: 0x0
key4: 0x0
key5: 0x0
key6: 0x0
------------------
```

Previously, JetKVM would output this instead:

```
-- hid_keyboard --
modf: 0x0
key1: 0xe6
key2: 0x0
key3: 0x0
key4: 0x0
key5: 0x0
key6: 0x0
------------------
```

Alternatively, the `keyboardReport` log entries can be observed (I did not initially know about it when investigating this issue).

The key codes correspond to definitions in the [Linux USB HID gadget driver](https://www.kernel.org/doc/Documentation/usb/gadget_hid.txt).

Steps to validate:

1. Use [Swiss German](https://kbdlayout.info/KBDSG/) keyboard on your host computer
2. Start Debian Linux with Gnome on the computer attached to JetKVM
3. Configure `German (Switzerland)` as keyboard layout on the computer attached to JetKVM
4. Open LibreOffice Writer
5. Type Alt Gr + 2 (for the `@` sign). Previously: this would output `2`, now it outputs `@`

image

### DIFF
--- a/ui/src/components/WebRTCVideo.tsx
+++ b/ui/src/components/WebRTCVideo.tsx
@@ -332,18 +332,19 @@ export default function WebRTCVideo() {
           // Example: If altKey is true, keep all modifiers
           // If altKey is false, filter out 0x04 (AltLeft)
           //
-          // But intentionally do not filter out 0x40 (AltRight) to enable Alt Gr (Alt Graph)
-          // as a modifier. The altKey attribute is not set on key combinations involving the
-          // Alt Gr key, which means the modifier would otherwise unintentionally disappear
-          // from the filteredModifiers list.
+          // But intentionally do not filter out 0x40 (AltRight) to enable Alt Gr
+          // (Alt Graph) as a modifier. The `altKey` attribute is set to `false` on
+          // key combinations involving the Alt Gr key, which means the modifier
+          // would otherwise be unintentionally removed from the filteredModifiers
+          // list.
           //
           // For example, the KeyboardEvent for Alt Gr + 2 has the following structure:
           // - altKey: false
           // - code:   "Digit2"
           // - type:   ["keydown" | "keyup"]
           //
-          // Adding and removing 0x40 (AltRight) from and to the list of active modifiers is
-          // taken care of by the respective logic in the keyUpHandler an keyDownHandler.
+          // Adding and removing 0x40 (AltRight) from and to the list of active
+          // modifiers is handled by keyUpHandler an keyDownHandler.
           .filter(
             modifier =>
               altKey ||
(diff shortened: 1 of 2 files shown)


dlorch:
@Nevexo could you review this PR? `Alt Gr` was indeed not recognized by JetKVM, due to the way the app was keeping track of which modifier was being pressed on the physical keyboard at any given time, and the fact that `Alt Gr` does *not* identify itself as an `Alt Key`.

dlorch:
@Nevexo @adamshiervani could this be looked into and be acknowledged as an open issue? Alt Gr not being recognized is a reproducible problem. It has been raised at least 7 times (if I counted right) in #30, and independently in #118 and #313. Personally, I am unable to type the `@` sign (Alt Gr + 2) on my Swiss German keyboard, among other essential keys like `|` (Alt Gr + 1) that are unlocked by this modifier.

I am open for other solution approaches, but that would require taking a bit of a broader view on how the logic behind `filteredModifiers`. Any input from you on this would be greatly appreciated.

adamshiervani (approved):

PR #432, opened 2025-05-09
Upgrade npm packages for the UI

IDisposable:
## Moving to current  as of 2025-05-09 for almost everything.

- Remove the erroneous extra dependency to old xterm package since the correct @xterm/xterm package was already included (suspect a bad merge) and it was causing issues with _react 19.1_.
- Switched to using the hooks exposed in the `usehooks-ts` package (this package was already referenced, suspect a bad merge) removing our private copies of `useInterval`, `useIsMounted`, `useResizeObserver` which are identical.
- Added import of `JSX` from react now needed because **NPX** is not in global scope in _react 19.x_.
- Explicitly cast the `ref` of included elements due to change in _react 19.x_


IDisposable:
Still maybe to upgrade (I need to test more).

Runtime dependencies:
- @vitejs/plugin-basic-ssl: 1.2.0 -> 2.0.0
- cva: 1.0.0-beta.1 -> 1.0.0-beta.3
- focus-trap-react: 10.2.3 -> 11.0.3
- framer-motion: 11.15.0 -> 12.10.5
- react-router-dom: 6.22.3 -> 7.6.0
- tailwind-merge: 2.5.5 -> 3.2.0
- zustand: 4.5.2 => 5.0.4

Dev dependencies:
tailwindcss: 3.4.17 -> 4.1.6
vite: 5.2.0 -> 6.3.5

ui/src/components/Terminal.tsx
```diff
@@ -79,10 +79,11 @@ function Terminal({
     return () => {
       setDisableKeyboardFocusTrap(false);
     };
-  }, [enableTerminal, instance, ref, setDisableKeyboardFocusTrap, type]);
+  }, [ref, instance, enableTerminal, setDisableKeyboardFocusTrap, type]);
 
   const readyState = dataChannel.readyState;
   useEffect(() => {
+    if (!instance) return;
```
IDisposable, line 86: test instance and bail earlier

ui/src/components/Terminal.tsx
```diff
@@ -79,10 +79,11 @@ function Terminal({
     return () => {
       setDisableKeyboardFocusTrap(false);
     };
-  }, [enableTerminal, instance, ref, setDisableKeyboardFocusTrap, type]);
+  }, [ref, instance, enableTerminal, setDisableKeyboardFocusTrap, type]);
```
IDisposable, line 82: style nit: always put the monitorees in least-to-most changing order

ui/src/components/Terminal.tsx
```diff
@@ -158,13 +158,11 @@ function Terminal({
     return () => {
       window.removeEventListener("resize", handleResize);
     };
-  }, [ref, instance, dataChannel]);
+  }, [ref, instance]);
```
IDisposable, line 161: doesn't actually depend on dataChannel

ui/src/components/WebRTCVideo.tsx
```diff
@@ -67,7 +67,7 @@ export default function WebRTCVideo() {
 
   // Video-related
   useResizeObserver({
-    ref: videoElm,
+    ref: videoElm as React.RefObject<HTMLElement>,
```
IDisposable, line 70: react 19.x added Ref<T | null> so this we need to explicitly cast now

ui/src/components/WebRTCVideo.tsx
```diff
@@ -10,7 +10,7 @@ import {
   useVideoStore,
 } from "@/hooks/stores";
 import { keys, modifiers } from "@/keyboardMappings";
-import { useResizeObserver } from "@/hooks/useResizeObserver";
+import { useResizeObserver } from "usehooks-ts";
```
IDisposable, line 13: use the **already included** `usehook-ts` implementations and kill off our versions

ui/src/routes/devices.$id.settings.tsx
```diff
@@ -19,7 +19,7 @@ import { LinkButton } from "../components/Button";
 import { cx } from "../cva.config";
 import { useUiStore } from "../hooks/stores";
 import useKeyboard from "../hooks/useKeyboard";
-import { useResizeObserver } from "../hooks/useResizeObserver";
+import { useResizeObserver } from "usehooks-ts";
```
IDisposable, line 22: use the **already included** `usehook-ts` implementations and kill off our versions

ui/src/routes/devices.$id.settings.tsx
```diff
@@ -30,7 +30,7 @@ export default function SettingsRoute() {
   const scrollContainerRef = useRef<HTMLDivElement>(null);
   const [showLeftGradient, setShowLeftGradient] = useState(false);
   const [showRightGradient, setShowRightGradient] = useState(false);
-  const { width } = useResizeObserver({ ref: scrollContainerRef });
+  const { width = 0 } = useResizeObserver({ ref: scrollContainerRef as React.RefObject<HTMLDivElement> });
```
IDisposable, line 33: react 19.x added Ref<T | null> so this we need to explicitly cast now

Also, best to avoid width possibly being null

ui/src/routes/devices.tsx
```diff
@@ -5,7 +5,7 @@ import { ArrowRightIcon } from "@heroicons/react/16/solid";
 import DashboardNavbar from "@components/Header";
 import { LinkButton } from "@components/Button";
 import KvmCard from "@components/KvmCard";
-import useInterval from "@/hooks/useInterval";
+import { useInterval } from "usehooks-ts";
```
IDisposable, line 8: use the **already included** `usehook-ts` implementations and kill off our versions

adamshiervani (approved):

Issue #358, opened 2025-04-15
Scroll is extremely aggressive, even on low sensitivity


lucianmoldovanu:
# Context
I connect to a JetKVM device either locally or via JetJVM cloud, from a desktop host (Windows 11). In settings, I set the Mouse > Scroll Sensitivity to Low.

# Expectation
A scroll that is in line with the settings of the mouse (i.e. relatively similar to the experience on the local host) or with the mouse scroll settings on the remote machine (in my case, also a Windows 11 with Mouse > Lines to scroll at a time set to the minimum value (1).

# Observed behavior
Scroll is extremely aggressive, thus unusable. In a VSCode window with 1200 lines for example, one scroll point on the wheel gets translated to 1/3 of the screen - thus starting on the first line, the first down scroll point gets me to line ~400, the second scroll gets me to line ~800 and the last gets me to the end of the file. A similar behavior is observed in other apps too (Google Sheets, Microsoft Word etc).

lucianmoldovanu:
Duplicate issue of #317

john--:
This was my comment on #317 : I think the issue with Macs are that they send exponentially more "wheel" events than other OS'. It sounds like the authors of JetKVM might not own a Mac if they're not able to address this issue, so maybe in a future version they could offer an advanced mode where we can enter a numeric value for the scroll speed?
Because I'll be honest, as impressed as I am with this device, the scrolling issue renders it unusable for my needs.

lucianmoldovanu:
Understand, but this is equally unusable on Windows as well, nothing to do with Mac specifically (just to be clear).

john--:
I see.
FWIW I use PiKVM and its out of the box settings were too sensitive as well. I had to dial my settings to the absolute minimum to get a usable experience:

![Image](https://github.com/user-attachments/assets/6295fa1a-c091-4585-b07c-c4133e338898)

If the authors of JetKVM could make an equivalent setting to this one, it would be fantastic

Geo4TH:
Oddly when I scroll, the first "click" does nothing, and the second click will scroll from cell 1 in google sheets to cell 210. This is on the lowest sensitivity, but it doesn't change on other sensitivity settings.

(Windows 10, i7-8700k, 48GB, UHD Graphics 630) Mouse on the remote machine is a Logitech G604

john--:
[...]

Issue #317, opened 2025-04-02
Scrolling speed still broken


jackislanding:
When scrolling, the speed is still too fast. Changing the settings does not appear to do anything.

STR4NG3RdotSH:
Confirmed.
For example, in VSCode, in a script thats 835 lines it only takes 3 notches/clicks to hit bottom of the document. Jumps from line 1, to 342, to 686 and finally 835.
I also adjusted the scroll sensitivity, still jumped the same amount of lines.

STR4NG3RdotSH:
I just noticed this only impacts some applications on the host. VSCode still jumps hundreds of lines, but notepad (windows) scrolls fine/smooth. I can't seem to find an exact setting to tweak in VSCode that might help, so I'm thinking its VSCode interpreting the scroll command coming from JetKVM that may be the issue. Maybe the scroll command coming from JetKVM can be tweaked.
Recreate issue:
Open long doc in notepad, observe beautiful scroll
Open long doc in VSCode (or other affected app, maybe OP has example), observe hundreds of lines jumped when scroll wheel touched

Tenmanichan:
I tried it in Outlook, but the scrolling still flies off to the other side of the world.

rollerbabe:
For me, it doesn't seem application-specific.   For example a large file in notepad still jumps all over the place. (Windows 11)

devinbergin:
Yeah having the same issue in VSCode going from macOS to Windows 11. The smallest scroll I do on the trackpad jumps about 300 lines.

sravanrox:
yeah doing from mac to windows outlook scroll speed is way too fast..

john--:
Me again. I think the issue with Macs are that they send exponentially more "wheel" events than other OS'. It sounds like the authors of JetKVM might not own a Mac if they're not able to address this issue, so maybe in a future version they could offer an advanced mode where we can enter a numeric value for the scroll speed?
Because I'll be honest, as impressed as I am with this device, the scrolling issue renders it unusable for my needs.

sravanrox:
yep i feel the same.. Not able to use it due to this scrolling issue. Hopefully they fix it soon

john--:
I posted some findings in #182  . My hunch is that there's a bug in the device emulator, as the scroll amounts are not linear as youd expect.

john--:
[...]

Issue #182, opened 2025-02-21
Mouse wheel does not scroll - it jumps


briangeek:
App: 0.3.8-dev202502191413
System: 0.2.2

Also happened on:
App: 0.3.7
System: 0.2.2

When trying to use scroll wheel on mouse, scroll bar will jump huge intervals instead of scrolling smoothly.

Tenmanichan:
I am experiencing the same phenomenon.

If I rotate the scroll by just one notch, the display area scrolls as if it had been rotated 3000 notches.

MehdyDriouech:
Same here

Yoshifresh:
Same here as well.

CForChrisProooo:
Same, any ideas?

rollerbabe:
I have this issue as well. Very annoying.

regg00:
Same issue here with or without smooth scrolling activated on my Logitech mouse.
The slightest touch to the scroll wheel jumps the screen to the top or bottom of the scroll area.

droach282:
Same here.

CCSnell:
Ditto. Janky Scroll

skotikus:
Same here

App: 0.3.9
System: 0.2.3

john--:
Seems related to #358 and #317 
While this doesn't seem like a high priority to the maintainers, a lot of us consider this to be a dealbreaker.
If the application code is ever released (and not just the UI), I'd gladly take this one on myself.

rmschooley:
I just received my jetkvm a few days ago. Like others, I love the device but the current scrolling behavior is a bit annoying. Although it's been a few years, I have some knowledge about this kind of mouse behavior and would like to offer a few comments...

A standard usb mouse will report a positive value when the wheel is moved in one direction and a negative value when the wheel is moved in the opposite direction. If the wheel was moved a single "click" since the last report, the value will be '1' in magnitude. If the wheel was moved two "clicks" since the last report, the value will be '2' in magnitude, and so on, and so forth. I think a positive value indicates a scroll down and a negative value indicates a scroll up. But, I might have that mixed up.

I am not sure there is anything symmetric in the html api (which i think is similar if not identical to what is being used by jetkvm). In the html api, I believe the wheel event will only indicate the amount of wheel movement depending on the browser -- either 1) the number of pixels, or 2) the number of lines of content, or 3) the number of pages of content.

[...]

PR #434, opened 2025-05-10
Fix mouse scrolling issues

john--:
This defect was discovered in #182 and should address the issue there as well as #317 and #358.

Between @rmschooley and me, we narrowed the issue down to the /dev/hidg1 device, and @rmschooley suggested this change to the HID descriptor, which in my testing successfully resolved the erratic scrolling behaviour.


ym (approved):

john--:
Thank you!

PR #440, opened 2025-05-13
feat: Reset optionally reset USB HID in dev-deploy

### LINKED ISSUE #438: HID devices unavailable when running dev_deploy.sh (temporary issue until next release)
This is merely a temporary issue, but I figured I'd create a GH issue about if for visibility. Just in case any developers out there have rebased their code from dev since Sunday and are wondering why things suddenly stopped working for them.

Essentially in #182 we changed the device descriptor for /dev/hidg1 to fix a scrolling bug.

The issue occurs when you load the current 0.3.9 version of JetKVM prior to running the newer code from the dev branch -- this is fairly typical since the release version is the one that starts up on boot. The 0.3.9 version attempts to create the device with the old descriptor while the newer code tries to create the device with the newer descriptor. The system doesn't like this.

You'll notice behaviour like `/dev/hidg0`, `/dev/hidg1`, and `/dev/hidg2` no longer existing, and all keyboard/mouse input failing to be recognized.

This issue will be resolved once the next version (after 0.3.9) is released -- which will allow the descriptions in both the release and debug versions to match.

However, in the interim you can get around this by executing the following prior to executing `dev_deploy.sh`:

    rm -rf /sys/kernel/config/usb_gadget/jetkvm/configs/c.1/hid.usb1/

Thanks and sorry for the inconvenience.

EDIT: @dlorch seems to have a more elegant solution [here](https://github.com/jetkvm/kvm/issues/438#issuecomment-2877283351).

IDisposable:
Adds `--reset-usb-hid` command to delete the configured USB HID device before running.

Fixes #438

### DIFF
--- a/dev_deploy.sh
+++ b/dev_deploy.sh
@@ -24,6 +24,7 @@ show_help() {
 REMOTE_USER="root"
 REMOTE_PATH="/userdata/jetkvm/bin"
 SKIP_UI_BUILD=false
+RESET_USB_HID_DEVICE=false
 LOG_TRACE_SCOPES="${LOG_TRACE_SCOPES:-jetkvm,cloud,websocket,native,jsonrpc}"
 
 # Parse command line arguments
@@ -39,6 +40,10 @@ while [[ $# -gt 0 ]]; do
             ;;
         --skip-ui-build)
             SKIP_UI_BUILD=true
+            shift
+            ;;
+        --reset-usb-hid)
+            RESET_USB_HID_DEVICE=true
             shift
             ;;
         --help)
@@ -73,6 +78,12 @@ ssh "${REMOTE_USER}@${REMOTE_HOST}" "killall jetkvm_app_debug || true"
 
 # Copy the binary to the remote host
 ssh "${REMOTE_USER}@${REMOTE_HOST}" "cat > ${REMOTE_PATH}/jetkvm_app_debug" < jetkvm_app
+
+if [ "$RESET_USB_HID_DEVICE" = true ]; then
+    # Remove the old USB gadget configuration
+    ssh "${REMOTE_USER}@${REMOTE_HOST}" "rm -rf /sys/kernel/config/usb_gadget/jetkvm/configs/c.1/hid.usb*"
+    ssh "${REMOTE_USER}@${REMOTE_HOST}" "ls /sys/class/udc > /sys/kernel/config/usb_gadget/jetkvm/UDC"
+fi
 
 # Deploy and run the application on the remote host
 ssh "${REMOTE_USER}@${REMOTE_HOST}" ash << EOF


dlorch:
I'm running into the same issue with `/de/hidg*` not being available. I think rebinding UDC should be done like this:

```
# ls /dev/hidg*
ls: /dev/hidg*: No such file or directory
# ls /sys/class/udc > /sys/kernel/config/usb_gadget/jetkvm/UDC
# cat /sys/kernel/config/usb_gadget/jetkvm/UDC
ffb00000.usb
# ls /dev/hidg*
/dev/hidg0  /dev/hidg1  /dev/hidg2
# echo > /sys/kernel/config/usb_gadget/jetkvm/UDC
# ls /dev/hidg*
ls: /dev/hidg*: No such file or directory
```

dlorch:
https://github.com/jetkvm/kvm/issues/438#issuecomment-2877283351

IDisposable:
I think I've grokked what you are recommending @dlorch, please advise

dlorch:
@IDisposable you *shouldn't* need to modify the USB gadget configuration in `/sys/kernel/config/usb_gadget/jetkvm/configs/`.

Instead, the desired logic would be:
- if `/sys/kernel/config/usb_gadget/jetkvm/UDC` is empty
  - echo "Rebinding USB Gadget UDC"
  - `ls /sys/class/udc > /sys/kernel/config/usb_gadget/jetkvm/UDC`
  - echo -n "Available HIDs: "
  - `ls /dev/hidg*`
- fi 

For reference:
* `/sys/kernel/config/usb_gadget/jetkvm/functions/` (folder) contains the *created* USB gadget function configurations, managed by JetKVM in [internal/usbgadget/config.go](https://github.com/jetkvm/kvm/blob/dev/internal/usbgadget/config.go)
* `/sys/kernel/config/usb_gadget/jetkvm/configs/` (folder) contains the *enabled* USB gadget functions (softlinks to the functions in `functions`), also managed by JetKVM in [internal/usbgadget/config.go](https://github.com/jetkvm/kvm/blob/dev/internal/usbgadget/config.go)
* `/sys/kernel/config/usb_gadget/jetkvm/UDC` (file) contains the *bound* USB Device Controllers (UDC), managed by JetKVM in [internal/usbgadget/udc.go](https://github.com/jetkvm/kvm/blob/dev/internal/usbgadget/udc.go)

The scope of this PR should be to re-bind the UDC, which will make the `/dev/hidg*` devices re-appear.

Before scenario (empty `UDC` file):

```
# cat /sys/kernel/config/usb_gadget/jetkvm/UDC
# ls /dev/hidg*
ls: /dev/hidg*: No such file or directory
```

After scenario:

```
# cat /sys/kernel/config/usb_gadget/jetkvm/UDC
ffb00000.usb
# ls /dev/hidg*
/dev/hidg0  /dev/hidg1  /dev/hidg2
```

Of course, these are all not real files and folders, they are maintained by an in-memory structure managed by ConfigFS, just as a side note.

IDisposable:
I will revisit this tomorrow then, sorry that I got it wrong.

dlorch:
> I will revisit this tomorrow then, sorry that I got it wrong.

Well I *think* that's what needs to happen at least, to be cross-checked in the context of what is happening in https://github.com/jetkvm/kvm/issues/438. Let me know if I need to investigate anything on my end to validate that also.

dlorch:
> I will revisit this tomorrow then, sorry that I got it wrong.

I had a closer look at @john--'s PR #434. The suggestion that John had is the correct one, mine solves for a similar but different problem. Apologies!

What seems not yet solved however is when a user upgrades to the latest app version. This would require a reboot of the JetKVM device or a "reset" like the one you added in the shell script, or else the users would might see a breaking change. Has this been taken into consideration?

john--:
> > I will revisit this tomorrow then, sorry that I got it wrong.
> 
> I had a closer look at @john--'s PR #434. The suggestion that John had is the correct one, mine solves for a similar but different problem. Apologies!
> 
> What seems not yet solved however is when a user upgrades to the latest app version. This would require a reboot of the JetKVM device or a "reset" like the one you added in the shell script, or else the users might see a breaking change. Would this not be an issue? I am unsure how the upgrade procedure works.

I'm a very new user to JetKVM, so tell me: how do uprades normally work? I assumed it would pull down the latest versions of everything and reboot, making this a non-issue -- but maybe this isn't the case?

john--:
I'm thinking the HIDs might need to be reset in the application to solve for these upgrade scenarios. Something like:
https://github.com/john--/kvm/commit/558dd43a9378d9c0e61eaec246eced307c63dbf9

Any opinions?

PR #447, opened 2025-05-14
feat(ui): Add Ctrl+Alt-Backspace combination key to Virtual Keyboard

### LINKED ISSUE #445: Please add more common hotkeys
Hi,

Please add more common hotkeys:
- Ctrl+Alt+Fx
- Alt+PrintScreen+letter
- Ctrl+Alt+backspace

These are usually needed when KVM is connected to Linux machine

Thanks

IDisposable:
Fixes #445 (somewhat)

### DIFF
--- a/ui/src/components/VirtualKeyboard.tsx
+++ b/ui/src/components/VirtualKeyboard.tsx
@@ -143,6 +143,16 @@ function KeyboardWrapper() {
         return;
       }
 
+      if (key === "CtrlAltBackspace") {
+        sendKeyboardEvent(
+          [keys["Backspace"]],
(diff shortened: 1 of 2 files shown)


IDisposable:
Ready forr review @ym @adamshiervani

adamshiervani:
I'd like to get rid of the Keyboard Shortcuts in the Virtual Keyboard and only lean into Keyboard Macros.

IDisposable:
> I'd like to get rid of the Keyboard Shortcuts in the Virtual Keyboard and only lean into Keyboard Macros.

Since shortcuts don't work yet (throws errors when I save), I can't see how that works on the UI... but I really like having those three shortcuts... but it's your UI ;)

PR #449, opened 2025-05-14
feat(ui): Enhance Virtual Keyboard for US

IDisposable:
Fix the display problems with shifted keys (should show capitals and symbols, but was showing (KeyX) and the state management was losing the Shift modifer.

Add the missing keys from the directional keys section, added Insert, added Print Screen/System Request, added Scroll Lock, added Pause/Break. Fixed display for the Delete key.

Added (but not exposed on normal US keyboard yet) the Numpad Equal

Builds on #447


IDisposable:
Ready for review @ym @adamshiervani

PR #452, opened 2025-05-15
fix(ui): Adjust EmptyCard icon size and tweak SettingsMacros

adamshiervani:


ui/src/components/EmptyCard.tsx
```diff
@@ -30,7 +30,7 @@ export default function EmptyCard({
         <div className="max-w-[90%] space-y-1.5 text-center md:max-w-[60%]">
           <div className="space-y-2">
             {IconElm && (
-              <IconElm className="mx-auto h-6 w-6 text-blue-600 dark:text-blue-400" />
+              <IconElm className="mx-auto h-5 w-5 text-blue-600 dark:text-blue-600" />
```
IDisposable, line 33: Just something I noticed working on the Tailwind upgrade... the `h-5 w-5` can be written `size-5` if you prefer :)

ui/src/routes/devices.$id.settings.macros.tsx
```diff
@@ -168,15 +168,15 @@ export default function SettingsMacrosRoute() {
...
+                <p className="mt-1 ml-4 overflow-hidden text-xs text-slate-500 dark:text-slate-400">
                   <span className="flex flex-col items-start gap-1">
                     {macro.steps.map((step, stepIndex) => {
                       const StepIcon = stepIndex === 0 ? LuMoveRight : LuCornerDownRight;
 
                       return (
                         <span key={stepIndex} className="inline-flex items-center">
                           <StepIcon className="mr-1 h-3 w-3 shrink-0 text-slate-400 dark:text-slate-500" />
-                          <span className="px-2 py-0.5 rounded-md border border-slate-200/50 dark:border-slate-700/50 bg-slate-50 dark:bg-slate-800">
+                          <span className="rounded-md border border-slate-200/50 bg-slate-50 px-2 py-0.5 dark:border-slate-700/50 dark:bg-slate-800">
```
IDisposable, line 179: Quick question... what is your preferred order of classes? I would rather get it right the first time :)

IDisposable:
Love it.

ui/src/components/EmptyCard.tsx
adamshiervani: Yeah, I saw that too. Let's use that moving forward or when we see it.

ui/src/routes/devices.$id.settings.macros.tsx
adamshiervani: Actually, this is sorted by https://github.com/tailwindlabs/prettier-plugin-tailwindcss

PR #470, opened 2025-05-17
Improve/Simplify Mouse Wheel Scroll Behavior

rmschooley:
This pull request makes several small but significant changes to the mouse wheel scroll behavior. In addition to guaranteeing that every mouse wheel click produces a scroll, these changes also maintain most if not all of the acceleration scroll values provided by many popular browsers. These changes simplify the code and make the various user scroll sensitivity settings completely unnecessary. However, these settings were not removed in this request.

These changes have been tested under the following environments. Unless otherwise noted, the scrolling behavior seems to work perfectly fine in these environments.

1. Windows 10 - Chrome 136 - Mouse (Delta = 100)
2. Windows 10 - Edge 136 - Mouse (Delta = 100)
3. Windows 10 - Firefox 138 - Mouse (Delta = 108)
4. Windows 10 - Opera 119 - Mouse (Delta = 8)
5. Linux Ubuntu 24.04.2 - Chrome 136 - Mouse (Delta = 120) + Trackpad
6. Linux Ubuntu 24.04.2 - Edge 136 - Mouse (Delta = 120) + Trackpad
7. Linux Ubuntu 24.04.2 - Firefox 138 - Mouse (Delta = 138) + Trackpad
8. Linux Ubuntu 24.04.2 - Opera 119 - THIS BROWSER CANNOT START JETKVM!
9. Android 15 - Chrome 136 - Mouse (Delta = 64)
10. Android 15 - Edge 136 - Mouse (Delta = 64)
11. Android 15 - Firefox 138 - Mouse (Delta = 63) - SCROLLING IS BACKWARDS!
12. Android 15 - Opera 89 - THIS BROWSER CANNOT START JETKVM!

This pull request corrects a scrolling polarity issue and fixes some invalid scrolling values not properly handled by my previous closed pull request on this issue.

Thank you very much for your consideration!

### DIFF
--- a/internal/usbgadget/hid_mouse_absolute.go
+++ b/internal/usbgadget/hid_mouse_absolute.go
@@ -113,7 +113,7 @@ func (u *UsbGadget) AbsMouseWheelReport(wheelY int8) error {
 	}
 
 	err := u.absMouseWriteHidFile([]byte{
-		2,                  // Report ID 2
+		2,            // Report ID 2
 		byte(wheelY), // Wheel Y (signed)
 	})
 
--- a/internal/usbgadget/hid_mouse_absolute.go
+++ b/internal/usbgadget/hid_mouse_absolute.go
@@ -107,23 +107,15 @@ func (u *UsbGadget) AbsMouseWheelReport(wheelY int8) error {
 	u.absMouseLock.Lock()
 	defer u.absMouseLock.Unlock()
 
-	// Accumulate the wheelY value
-	u.absMouseAccumulatedWheelY += float64(wheelY) / 8.0
-
-	// Only send a report if the accumulated value is significant
-	if abs(u.absMouseAccumulatedWheelY) < 1.0 {
+	// Only send a report if the value is non-zero
+	if wheelY == 0 {
 		return nil
 	}
 
-	scaledWheelY := int8(u.absMouseAccumulatedWheelY)
-
 	err := u.absMouseWriteHidFile([]byte{
 		2,                  // Report ID 2
-		byte(scaledWheelY), // Scaled Wheel Y (signed)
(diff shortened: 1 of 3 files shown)


adamshiervani (approved):
Thanks for the extensive testing and PR!

I just tested this on my dev machine and it works great. I think there is space to make the Low, Default and High settings more distinct, but let's get this into the next release first.

john--:
> Thanks for the extensive testing and PR!
> 
> I just tested this on my dev machine and it works great. I think there is space to make the Low, Default and High settings more distinct, but let's get this into the next release first.

Does High/Low/Default even do anything after this change?

rmschooley:
No -- This pull request does not reference those settings. As I previously mentioned, I am not sure how meaningful those settings were since the scrolling was not working quite right when they were added to the code. I did not remove those settings in case we wanted to add some variation of them back into the code. I would guess the vast majority of users will be very happy with the scrolling behavior of the new code. But, I guess we shall see whenever the new firmware is released.

I did see that someone asked for a setting to reverse the scrolling direction. However, I am not sure if they were requesting that feature based on the current broken firmware or not.

adamshiervani:
The only setting that is used is the `blockDelay`, and we don't really need that either. I'll make a PR simply removing the setting, as the current implementation is a good default. We can always add the Scroll Sensitivity back if there are riots.

robertbudding:
> Tested and verified that the scroll improvements are much better. The previous settings didn't appear to make any difference. Thanks for this much-needed update.

Did you test this PR by itself to compare the differences?
PR #434 resolved the main scrolling problem for most users. It was included in the same build. Many of us noticed differences between the sensitivity options. I run a custom build with this change reverted because I prefer the "Low" sensitivity on my equipment.

PR #488, opened 2025-05-20
fix: Handle any panic during the callRPCHandler

IDisposable:
Added a wrapper around the `callRPCHandler` function to recover from panic and translate that to and error so that the RPC thread doesn't just die when something malformed comes in. This should keep the Jet flying.

### DIFF
--- a/jsonrpc.go
+++ b/jsonrpc.go
@@ -466,7 +466,31 @@ func rpcSetTLSState(state TLSState) error {
 	return nil
 }
 
-func callRPCHandler(handler RPCHandler, params map[string]interface{}) (interface{}, error) {
+type RPCHandler struct {
+	Func   interface{}
+	Params []string
+}
+
+// call the handler but recover from a panic to ensure our RPC thread doesn't collapse on malformed calls
+func callRPCHandler(handler RPCHandler, params map[string]interface{}) (result interface{}, err error) {
+	// Use defer to recover from a panic
+	defer func() {
+		if r := recover(); r != nil {
+			// Convert the panic to an error
+			if e, ok := r.(error); ok {
+				err = e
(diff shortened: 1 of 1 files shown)


IDisposable:
Hey @ym, looks like we're getting really close to a release... while finishing up @dlorch's paste keyboard cleanup, I found that malformed RPC calls were crashing the JetKVM's RPC thread so it became unresponsive. I wrapped it in a `func` that uses `recover()` to morph _panics_ into plain errors. This keeps the RPC thread flying.

This is a harmless change but with HUGE stability improvements

ym (approved):

dlorch:
Nice one, I confirm I was also able to crash the JetKVM process with (unintentionally) malformed RPC input.

IDisposable:
This really should make it into 0.4.0 @adamshiervani

ym:
> Nice one, I confirm I was also able to crash the JetKVM process with (unintentionally) malformed RPC input.

Could you also share the error message (stack trace) and payload after masking sensitive information? The functions it called should be robust enough to handle exceptions, but it is also necessary to have a 'recover()' function as a last resort to avoid crashing the application.

Many thanks for the contribution!

dlorch:
@IDisposable did your crashes occur on `keyboardReport` possibly and did you capture the inputs + stack trace?

IDisposable:
I didn't save any traces, but it was on a malformed keyboard report message, I think with a _null_ for the keys instead of an empty array. It couldn't got a NPE on this line of _jsonrpc.go_ calling `.Type()` so the `elemValue` was `nil`
```go
  if !elemValue.Type().ConvertibleTo(paramType.Elem()) {
```

PR #405, opened 2025-05-02
feat(ui): enable multiple keyboard layouts for "paste text" to remote host

dlorch:
Pasting text in the "Paste text" dialog sends individual key strokes to the target device using an USB keyboard emulation. Which numerical key codes are being sent depend on a mapping. This mapping was based on US English until now.

This PR provides the mechanism to enable multiple keyboard layouts for different languages and comes included with keyboard layouts for:
* Czech
* English (UK)
* English (US)
* French
* German
* Italian
* Norwegian
* Spanish
* Swedish
* Swiss French
* Swiss German

Let me know if you would like to see further keyboard layouts and I see what I can do.

I've seen #116, but the scope of that PR is much bigger and aiming at overhauling the entire keyboard functionality. For this PR here, I wrote the minimal set of changes to enable keyboard layouts in the "Paste text" dialog. We can then later align the initiatives.

This provides fixes for #22, #30, #65, #104, #47.

Could potentially address #436.

paste-modal-spanish

keyboard-layout-config

PR #495, opened 2025-05-21
chore(ui): Clean new keyboard option

IDisposable:
Fixed the Tailwind CSS syntax for `in` (nested) selector
Added missing React dependency for `useEffect`

### DIFF
--- a/ui/src/routes/devices.$id.settings.keyboard.tsx
+++ b/ui/src/routes/devices.$id.settings.keyboard.tsx
@@ -46,7 +46,7 @@ export default function SettingsKeyboardRoute() {
       if ("error" in resp) return;
       setKeyboardLayout(resp.result as string);
     });
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [send, setKeyboardLayout]);
 
   const onKeyboardLayoutChange = useCallback(
     (e: React.ChangeEvent<HTMLSelectElement>) => {
--- a/ui/src/routes/devices.$id.settings.tsx
+++ b/ui/src/routes/devices.$id.settings.tsx
@@ -151,7 +151,6 @@ export default function SettingsRoute() {
                     className={({ isActive }) => (isActive ? "active" : "")}
                   >
                     <div className="flex items-center gap-x-2 rounded-md px-2.5 py-2.5 text-sm transition-colors hover:bg-slate-100 dark:hover:bg-slate-700 in-[.active]:bg-blue-50 in-[.active]:text-blue-700! md:in-[.active]:bg-transparent dark:in-[.active]:bg-blue-900 dark:in-[.active]:text-blue-200! dark:md:in-[.active]:bg-transparent">
-
                       <LuMouse className="h-4 w-4 shrink-0" />
                       <h1>Mouse</h1>
                     </div>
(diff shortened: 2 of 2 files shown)


IDisposable:
@dlorch @ym this is a tiny change to cover the new Tailwind syntax (which changed between when #405 was started):
`[.active_&]:` -> `in-[.active]:`

Also adds the missing dependency for the `useEffect`

ui/src/routes/devices.$id.settings.keyboard.tsx
```diff
@@ -51,7 +51,6 @@ export default function SettingsKeyboardRoute() {
       />
 
       <div className="space-y-4">
-        { /* this menu item could be renamed to plain "Keyboard layout" in the future, when also the virtual keyboard layout mappings are being implemented */ }
```
dlorch, line 54: Is there a reason to remove this? I felt it served a purpose, as a reminder that `title="Paste text"` can be renamed to `title="Keyboard layout"` once virtual keyboards have caught up with their keyboard layouts.

`title="Paste text"` only exists as an intermediary solution, since there are different development speeds for "paste text" and "virtual keyboard" functionality. This should be cleaned up later. Maybe there is a better place to persist this information?

IDisposable: I've come around to thinking it is clearer the way it is.. the setting is and should be related to the the remote's OS keyboard layout or things just won't work.

IDisposable: ... btw, working on the Virtual Keyboard again... getting the Ctrl/Alt to track "down" states like it does for Shift so it's more like a touch keypad... which will also allow sending things like Ctrl+Ctrl+1 from the Virtual Keyboard as well as the physical keyboard.

ui/src/routes/devices.$id.settings.keyboard.tsx
```diff
@@ -51,7 +51,6 @@ export default function SettingsKeyboardRoute() {
...
         <SettingsItem
           title="Paste text"
```
IDisposable, line 55: Oh... I see, the comment refers to the `title="Paste text"` and blurb in the `<p>` below... when I get VK working with keyboard layouts, indeed we will need to come back here.

ui/src/routes/devices.$id.settings.keyboard.tsx
dlorch: Exactly! It would be confusing for users to have a keyboard layout setting for "paste box" and another keyboard layout setting for "virtual keyboard", that should be the same eventually, when development on layouts for "virtual keyboards" has caught up. Hence the reminder above.

ui/src/routes/devices.$id.settings.keyboard.tsx
dlorch: While working on keyboard layouts for virtual keyboards, you might be introducing another setting _just_ for virtual keyboard layouts. But later this really should be merged into one common "Keyboard layout" setting, IMO.

ui/src/routes/devices.$id.settings.keyboard.tsx
IDisposable: I'll restore the comment until such time as it needs to change to something more valuable

IDisposable:
Reverted the comment.

ui/src/routes/devices.$id.settings.keyboard.tsx
```diff
@@ -25,7 +25,7 @@ export default function SettingsKeyboardRoute() {
       if ("error" in resp) return;
       setKeyboardLayout(resp.result as string);
     });
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [send, setKeyboardLayout]);
```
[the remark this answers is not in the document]
IDisposable: @ym, this really needs to be merged... tons of React rerenders until we do...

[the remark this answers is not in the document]
ym: Empty dependencies mean that it will behave like `componentDidMount`, running only once when the component mounts. I can also confirm that it does not re-render outside the case mentioned above.

adamshiervani (approved):

ym (changes requested):

PR #498, opened 2025-05-22
feat(ui): Add Ctrl+Alt+Del to the action bar

IDisposable:
Since this is the sort of thing we do all the time, make it one-click away


IDisposable:
Rebased and ready @ym

IDisposable:
Adam commented that he would leverage macros, but that wasn't well received and the PR went stale and became unmergeable, so I reimplemented it because it was a great addition that shouldn't die.

PR #501, opened 2025-05-22
feat(ui): Add a clickable link to the header logo

IDisposable:
Might as well drive traffic to you :)


IDisposable:
Would targeting app.jetkvm.com be better @adamshiervani ?

adamshiervani:
I appreciate the gesture, but any links to external sources - even if they point to our own content - should be added with a clear and deliberate purpose. We should avoid including them too freely.

IDisposable:
Fine by me :)

PR #502, opened 2025-05-22
feat: sync keyboard led status

ym:


IDisposable:
This is going to be nice, I'll rebase my Virtual Keyboard PR once this is merged (since I'm mucking about with that state on the client as well)

Issue #421, opened 2025-05-05
Remote Reboot of JetKVM Device?


brianoconnorgaming:
I have a remotely deployed JetKVM Device that is acting up. after digging around in the Docs and settings, I havnt been able to find a way to remote reboot the device short of driving over there and doing a hard power cycle. This seems like a very overlooked feature for these devices!

john--:
you can ssh into it and reboot

bshogeman:
SSH only works for if you are in the same network. Not if you are real remote, outside the network.

john--:
> SSH only works for if you are in the same network. Not if you are real remote, outside the network.

The web terminal should do the trick in this case

jackstown:
It would be great to have a simple button in the settings

yanncd:
Hello, how to open web terminal ? I could not find it

yanncd:
> > SSH only works for if you are in the same network. Not if you are real remote, outside the network.
> 
> The web terminal should do the trick in this case

What do you mean by web terminal ? Is it a specific terminal ?

richard-parker:
I've had mine for about a month, and actually today was the first time I needed to use it remotely. It looks like I've hit the same issue as you guys - video not loading. However, I have a VPN to my network which enabled me to access the JetKVM via SSH. I can confirm that after logging into that, and issuing the `reboot` command, it came back up and worked.

A simple 'reboot device' button the JetKVM UI would be really helpful.

And it would also be very nice to know what causes this apparently time-related instability in the device's video connection.

axcro7:
I was experiencing issues with sending keystrokes (mouse was still working fine) and had to power cycle the device to get it working again.
A reboot button at least somewhere in the advanced settings would be super helpful

memesalot:
Thank you!! Much appreciated :)

PR #505, opened 2025-05-22
feat(ui): Enable reboot of device (#421)

adammkelly:
Basic reboot for the JetKVM device within the settings page.

No handling of after the reboot happens from the UI due to it being a disruptive path.
This could be a future improvement were it polls for the device to be responsive again if known to be on the same IP address.

image
image


IDisposable (approved):
Looks great

ui/src/routes/devices.$id.settings.general._index.tsx
```diff
@@ -92,6 +92,21 @@ export default function SettingsGeneralRoute() {
               />
             </SettingsItem>
           </div>
+
+          <div className="mt-2 flex items-center justify-between gap-x-2">
+            <SettingsItem
+              title="Reboot Device"
+              description="Power cycle the JetKVM"
```
dseven, line 99: It's not actually power-cycling, though - it's only rebooting the OS - not the same.

Rhys1210:
How do we get this function as from the screen shot above mine doesn't have the reboot option. 

Screenshot 2025-07-14 223754

dseven:
> How do we get this function as from the screen shot above mine doesn't have the reboot option.

Build from source, or wait for a future release that includes this feature.

PR #511, opened 2025-05-23
feat: add local web server loopback mode configuration

agoodkind:
# Add option to restrict local web server to loopback interface only

This PR adds a configuration option to restrict the JetKVM web server to only listen on the loopback interface (`127.0.0.1`/`localhost`).

**Use cases:**
- SSH tunneling access
- Cloud connectivity without local network exposure
- Enhanced security through reduced attack surface
- Behind reverse proxies

**Security benefits:** Improves device hardening by limiting network exposure and implementing defense-in-depth principles.

**Implementation:**
- `LocalLoopbackOnly` config option (default: false)
- Accessible via JSON-RPC: `getLocalLoopbackOnly`/`setLocalLoopbackOnly`
- Requires reboot to apply changes

This maintains full functionality via SSH or cloud while significantly reducing potential security risks. 

### Demo
Screenshot 2025-05-25 at 16 32 13
Screenshot 2025-05-25 at 16 32 04
image


IDisposable (approved):
Looks like a lot of whitespace changes... please check your tab settings

agoodkind:
> Looks like a lot of whitespace changes... please check your tab settings

I believe in go there is only one formatting setting, see https://go.dev/doc/effective_go#formatting

In this case the majority of the changes were additions to `type` `struct`s which according to that doc will get formatted to have the columns always aligned. It just so happens that these structs are quite big already so when `gofmt` is run it has to change all the previous lines. I can try to make the config key shorter but at the cost of it being less obvious what it does

web.go
```diff
@@ -52,8 +52,9 @@ type ChangePasswordRequest struct {
 }
 
 type LocalDevice struct {
-	AuthMode *string `json:"authMode"`
-	DeviceID string  `json:"deviceId"`
+	AuthMode     *string `json:"authMode"`
+	DeviceID     string  `json:"deviceId"`
+	LoopbackOnly bool    `json:"loopbackOnly"`
```
agoodkind, line 57: `gofmt` aligned the columns here since `LoopbackOnly` is longer than `AuthMode` and `DeviceID`

IDisposable (approved):

IDisposable:
Just clarifying, I am just a random developer that loves the JetKVM, as such my Approval means nothing except I like it ... I don't have write-rights here...

agoodkind:
Ah gotcha

agoodkind:
@ym

Issue #515, opened 2025-05-23
Support for connecting to the Internet via a proxy


solemnwarning:
Support for the JetKVM device to connect to the Internet (e.g. for downloading updates) via a HTTP proxy as required in some network environments.

solemnwarning:
Another point of consideration: should proxy auto-config (PAC) files be supported (requires a JS interpreter on the device)?

PR #521, opened 2025-05-24
Implement HTTP proxy option (#515).

solemnwarning:
This commit adds a "Proxy" field to the network settings screen, which can be used to specify a HTTP proxy for any outgoing requests from the device.

For testing, I have confirmed direct connections to check for updates work when the device is on an unrestricted network with no proxy set, and proxied connections to check for updates work when the device is on a network without a direct route to the Internet.

I haven't tested the HTTP time syncing or actual download/install of updates as I'm not sure how to exercise them.

### DIFF
--- a/internal/confparser/confparser.go
+++ b/internal/confparser/confparser.go
@@ -3,6 +3,7 @@ package confparser
 import (
 	"fmt"
 	"net"
+	"net/url"
 	"reflect"
 	"slices"
 	"strconv"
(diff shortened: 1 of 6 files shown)


ym (changes requested):
Could you please fix the linter errors and rename 'proxy' to 'http_proxy'? 'Proxy' sounds too generic, and doesn't include SOCKS5 or other types of proxy.

solemnwarning:
@ym if you can tell me what exactly the linter wants I'll get right on it (all I found when googling the error string was other linter runs).

FYI SOCKS proxies could be supported in the future (the Go HTTP client does in fact have built-in support), I just limited it to plain HTTP for the time being to keep it simple, but I'll do the rename anyway since its specifically for the HTTP client (unless you think there may be other connections made by the JetKVM in the future which _should_ be able to be proxied).

IDisposable (changes requested):
Looks great!

IDisposable (approved):

solemnwarning:
@ym is there anything still blocking this change?

ym (approved):

Issue #238, opened 2025-03-07
Num Lock key out of sync with remote


briangeek:
When connecting to a remote PC where NUM LOCK is off, but it is on locally, the status of the NUM LOCK is out of sync.

MehdyDriouech:
Same here, also with capslock

phatlewt:
I had this issue too, had to connect a keyboard and physically toggle the numlock on the remote pc for it to work.

ym:
Fixed in https://github.com/jetkvm/kvm/pull/502

PR #530, opened 2025-05-27
Remove Out Endpoint Descriptors from Keyboard, Absolute Mouse, and Relative Mouse

rmschooley:
Added attribute to remove unnecessary and potentially confusing out endpoint descriptors from the keyboard, absolute mouse, and relative mouse.

### DIFF
--- a/internal/usbgadget/hid_keyboard.go
+++ b/internal/usbgadget/hid_keyboard.go
@@ -14,9 +14,10 @@ var keyboardConfig = gadgetConfigItem{
 	path:       []string{"functions", "hid.usb0"},
 	configPath: []string{"hid.usb0"},
 	attrs: gadgetAttributes{
-		"protocol":      "1",
-		"subclass":      "1",
-		"report_length": "8",
+		"protocol":        "1",
+		"subclass":        "1",
+		"report_length":   "8",
+		"no_out_endpoint": "1",
 	},
 	reportDesc: keyboardReportDesc,
 }
--- a/internal/usbgadget/hid_mouse_absolute.go
+++ b/internal/usbgadget/hid_mouse_absolute.go
@@ -11,9 +11,10 @@ var absoluteMouseConfig = gadgetConfigItem{
 	path:       []string{"functions", "hid.usb1"},
 	configPath: []string{"hid.usb1"},
 	attrs: gadgetAttributes{
-		"protocol":      "2",
-		"subclass":      "1",
-		"report_length": "6",
+		"protocol":        "2",
+		"subclass":        "1",
+		"report_length":   "6",
+		"no_out_endpoint": "1",
 	},
 	reportDesc: absoluteMouseCombinedReportDesc,
 }
--- a/internal/usbgadget/hid_mouse_relative.go
+++ b/internal/usbgadget/hid_mouse_relative.go
@@ -11,9 +11,10 @@ var relativeMouseConfig = gadgetConfigItem{
 	path:       []string{"functions", "hid.usb2"},
 	configPath: []string{"hid.usb2"},
 	attrs: gadgetAttributes{
-		"protocol":      "2",
-		"subclass":      "1",
-		"report_length": "4",
+		"protocol":        "2",
+		"subclass":        "1",
+		"report_length":   "4",
+		"no_out_endpoint": "1",
 	},
 	reportDesc: relativeMouseCombinedReportDesc,
 }


rmschooley:
> We do need the keyboard endpoint for the LED indicators.

Unless JetKVM has some other special need for it, I do *not* think we need the keyboard out endpoint for the led indicators. Although it is true a keyboard *may* use an out endpoint to communicate the state of its led indicators, it is extremely non-standard for one to do so. Over the years, I have seen dozens if not hundreds of different keyboard implementations and have never seen an out endpoint be implemented to send the state of its led indicators prior to JetKVM. Most if not all other keyboards communicate the state of the led indicators via the Set_Report command across the device's default endpoint. In fact, this is *required* for boot keyboards.

**USB HID 1.11 Spec, Section 4.4**
The Interrupt Out pipe is optional. If a device declares an Interrupt Out endpoint then Output reports are transmitted by the host to the device through the Interrupt Out endpoint. If no Interrupt Out endpoint is declared then Output reports are transmitted to a device through the Control endpoint, using Set_Report(Output) requests.

Note Endpoint 0 is a Control pipe always present in USB devices. Therefore, only the Interrupt In pipe is described for the Interface descriptor using an Endpoint descriptor. In fact, several Interface descriptors may share Endpoint 0. An Interrupt Out pipe is optional and requires an additional Endpoint descriptor if declared.

**USB HID 1.11 Spec, Section 6.2.2.5**
Output items make Output reports accessible via the Control pipe with a Set_Report (Output) command.

**USB HID 1.11 Spec, Appendix C**
Synchronization between LED states and CAPS LOCK, NUM LOCK, SCROLL LOCK, COMPOSE, and KANA events is maintained by the host and NOT the keyboard. If using the keyboard descriptor in Appendix B, LED states are set by sending a 5-bit absolute report to the keyboard via a Set_Report(Output) request.

**USB HID 1.11 Spec, Appendix F.3**
The Boot Keyboard shall not maintain CAPS LOCK, NUM LOCK, SCROLL LOCK, COMPOSE, or KANA LED states without explicit Set_Report (Output) requests from the system.

**USB HID 1.11 Spec, Appendix F.4**
The system shall maintain synchronization between LED states the Caps Lock, Num Lock, or Scroll Lock events. The system sets LED states by sending a 5-bit absolute report to the keyboard via a Set_Report (specifying Output report) request.

rmschooley:
FYI: I have tested this change on my JetKVM. When the out endpoint *is* included, JetKVM sends the state of its led indicators via that endpoint. When the out endpoint *is not* included, JetKVM sends the state of its led indicators via its default endpoint using Set_Report().

In summary -- It seems like the linux hid keyboard usbgadget driver already contains all of the code necessary to properly send the state of its led indicators to a keyboard whether or not the keyboard implements an out endpoint.

As I mentioned in my previous post, boot keyboards are required to maintain the state of their led indicators via its default endpoint using Set_Report(). The keyboard examples in the specification also do *not* employ an out endpoint.

internal/usbgadget/hid_keyboard.go
```diff
@@ -14,9 +14,10 @@ var keyboardConfig = gadgetConfigItem{
 	path:       []string{"functions", "hid.usb0"},
 	configPath: []string{"hid.usb0"},
 	attrs: gadgetAttributes{
-		"protocol":      "1",
-		"subclass":      "1",
-		"report_length": "8",
+		"protocol":        "1",
+		"subclass":        "1",
+		"report_length":   "8",
+		"no_out_endpoint": "1",
```
IDisposable, line 20: Don't we need the out endpoint to get the keyboard LED messages from the computer? These are what drive the LEDs on the keyboard and a recent PR #502 added code to "trust" those messages so the remote is authoritative to the NumLock/CapsLock/ScrollLock/Compose/Kana.

See [hid_keyboard.go](https://github.com/jetkvm/kvm/blob/dev/internal/usbgadget/hid_keyboard.go#L96)

IDisposable (changes requested):

internal/usbgadget/hid_keyboard.go
rmschooley: Hmmm....I must admit to being a bit perplexed about PR #502.

1. What specific issue is PR #502 trying to solve?
2. Can you provide a small description on how it solves this issue?

Also - Are you saying that PR #502 depends on the led states being communicated over the non-default endpoint? If PR #502 does depend on this, do you realize it is unlikely to work within a boot operating environment where the software will be attempting to use the keyboard as a boot keyboard and will be sending these led states via Set_Report()?

I am able to get the led states between the remote computer and the local computer "out-of-sync" quite easily using the latest code on the development branch. So, I may just not be understanding what this new code in PR #502 is trying to accomplish.

If I am completely misunderstanding something here, I definitely apologize in advance! :)

rmschooley: I am still confused about the functionality of PR #502.

Is PR#502 supposed to keep the state of the keyboard leds on the remote computer (i.e. the computer to which JetKVM is attached) in sync with the state of the keyboard leds that are displayed in the browser which is browsing JetKVM?

If so - How do the different states of the new "LED State Synchronization" setting (i.e. Automatic, Browser Only, and Host Only" impact this behavior?

Also - Is it understood that it is still possible to get the keyboard leds on the keyboard that is attached to the system which is running the browser which is browsing JetKVM out-of-sync with all of the other keyboards?

Finally - Is it also understood that PR #502 will not work when JetKVM is attached to a system that is running an operating environment that is attempting to use the JetKVM keyboard as a boot keyboard?

P.S. I'm also confused about Issue #238. When I originally read this issue, I thought the submitter was stating that the state of the keyboard leds on the remote computer were out-of-sync with the state of the keyboard leds on the local computer (as opposed to the state of the keyboards leds displayed in the browser). I am not sure how anyone determined exactly what keyboard the submitter (or any of the other commenters on that issue) was talking about.

IDisposable: > Is PR#502 supposed to keep the state of the keyboard leds on the remote computer (i.e. the computer to which JetKVM is attached) in sync with the state of the keyboard leds that are displayed in the browser which is browsing JetKVM?

I think we need to start at the beginning... for USB keyboards:

1. In a PC/Mac, the **keyboard** doesn't modify the state of the latching indicators (CapsLock, NumLock, etc.) it merely keeps track of what it was last told they should be. No pressing of keys will change the LEDs until computer tells the keyboard to change.
2. At boot time, the PC uses default states, and the setting of any BIOS options to decide what the current state of the keyboard is (likely NumLock on, CapsLock off, etc)
3. When the PC initializes the USB port, and finds a keyboard attached, it **SENDS** a message to the keyboard to tell it what LEDs to light up. The keyboard does as it is told.
4. When OS loads, it may decide to _set_ the state of these things to a configured default. For instance in Windows it's possible to FORCE the NumLock on at boot (or login) time.
5. Let's say, for the sake of discussion, that the OS is in "caps lock is off" state.
6. The OS then **SENDS** a message to the keyboard to turn on whatever LEDs should be on.
7. The user types something on the keyboard and the appropriate scan code is sent to the PC... that code *DOES NOT CHANGE* based on the LEDs. It sends the code associated with the key you pressed. If you happen to be holding the Ctrl or Shift or Alt or Meta key, then that fact is sent in the modifiers... but note that when you pressed those modifier keys, the keyboard _sent the corresponding key code_ **with** the modifier state reflecting what it thinks you are holding down.
8. The computer receives the modifiers and key code and decides what character that maps to _based on the keyboard layout_ active. If you hit an `f` key, it'll send KeyF, which is key code 0x09. If you happen to be holding down the Left Shift key, then the modifier will be 0x02 (assuming no other modifier keys are being held).
9. The OS will (usually) interpret that to be a _capital F_.
10. If you press the `CapsLock` key, then the keyboard will send a CapsLock key code, which is 0x39.
11. The OS sees that key code and toggles it's internal state of caps lock (if at boot it was on, then this keypress turns it off, and vice versa).
12. The OS _then **SENDS**_ a message to the keyboard that says that the CapsLock should _now_ be **ON** so the LEDs message will have that bit on... and the keyboard will loyally turn on that LED.
13. At this point, if the user presses the `z` key, it'll send the KeyZ, which is key code 0x1d. The modifiers will be all zero (assuming you're not holding a Shift key down _too_), so the OS says... just a `z`... but wait, my caps lock internally is on, so this is a `Z`.

So that's why the PC tells the keyboard what LEDs are on.. _because it's the one tracking that!_

(Note: the keyboard doesn't care what LEDs, and does NOT change what it reports back, ever)

> If so - How do the different states of the new "LED State Synchronization" setting (i.e. Automatic, Browser Only, and Host Only" impact this behavior?

Short version:
- Browser Only: The application running on the end-user's browser assumes a starting state, and then toggles back and forth as the user presses _their own_ CapsLock key (and the others). When you press a key it tells the PC that you did... and hopefully you're on the same page
- Host Only: The JetKVM tracks the _last received_ LEDs setting and runs with it... so the JetKVM has to _listen_ on the USB for those messages and set internal state _as if it were a keyboard!_
- Automatic: Ummm... @ym? Help?

> Also - Is it understood that it is still possible to get the keyboard leds on the keyboard that is attached to the system which is running the browser which is browsing JetKVM out-of-sync with all of the other keyboards?

Ah, so there's the rub... as far as the browser+JetKVM _should be_ concerned, it should _always_ trust the PC... where thing get confusing is that (_at this time_) there's NOTHING on the browser end toggling CapsLock state/LEDs of the controlling person's LEDs... so at connection time we will show the right state at the bottom of the browser window, but it may not match the controlling person's keyboard state _for their OS_... so yep, that can be out of sync.

What would be a nice follow on to what we got with 502 is figuring out to have the controller's browser _set_ the CapsLock state on the controlling users' PC (and thus it will update the physical keyboard LEDs). I'm going to figure that out ASAP.

> Finally - Is it also understood that PR https://github.com/jetkvm/kvm/pull/502 will not work when JetKVM is attached to a system that is running an operating environment that is attempting to use the JetKVM keyboard as a boot keyboard?

Actually, it's MUCH better off  in this case to be in Host Only mode, because it should be seeing the OS/BIOS's view of whether the CapsLock (et al) are on (and thus the LED should be lit).

rmschooley: Thank you very much for the quick response and the detailed answers! Outside of the option for "Automatic", the description of the various settings for "LED State Synchronization" now makes perfect sense to me. Since I have some experience in the design of usb devices (including keyboards and mice) as well as usb host system software, I was aware and agree with most of the other things you mentioned. Nevertheless, it was an excellent summary for those who may not be as familiar.

_**HOWEVER...I think Step #12 in your above response is missing one important caveat that is essential to this discussion.**_

12. The OS then SENDS a message to the keyboard that says that the CapsLock should now be ON so the LEDs message will have that bit on... and the keyboard will loyally turn on that LED. _If the OS (or BIOS) is treating the keyboard as a BOOT keyboard *or* as a HID keyboard without an OUT Interrupt Endpoint, it will send this message to the keyboard's Default Control Endpoint via the Set_Report(Output) request. Otherwise, the OS (or BIOS) will send this message to the keyboard's OUT Interrupt Endpoint (as opposed to its Default Control Endpoint)._

Based on my interpretation of the code (as well as the previous comments made by @ym), PR #502 seems to be "listening" to JetKVM's OUT Interrupt Endpoint to determine when the LEDs are changing state. But, it does not seem to be listening to the keyboard's Default Control Endpoint. As such, it will *not* work when the OS/BIOS is treating the keyboard as a boot keyboard. 

For what it is worth, I have tested this new code (PR #502) in various different operating environments.  When the OS/BIOS is treating the keyboard as a HID keyboard with an OUT Interrupt Endpoint, it works exactly as you expect. When the OS/BIOS is treating the keyboard as a BOOT keyboard (i.e. a keyboard without an OUT Interrupt Endpoint), it does *not* work. Specifically, the LEDs on any keyboards attached to the remote computer change state just fine, but the LED states displayed on the browser on the local system never change state.

Please be aware that most BIOS operating environments in the past always treated USB keyboards as BOOT Keyboards. I suspect this is still probably true today, but don't know with absolute certainty.

IDisposable: That's some excellent context I was unaware of...so the JetKVM needs to listen on both channels?

rmschooley:
Since I don't think the issue between setting the led's on the boot keyboard and the non-boot keyboard (i.e. the hid keyboard) will be resolved any time soon (and perhaps not ever), I guess it probably makes sense to just close this pull-request and open a new issue describing that problem at least for the sake of history. Hopefully, we can get a request approved and merged deleting the unnecessary mouse out endpoints at that time.

PR #535, opened 2025-05-28
fix(ui): Fullscreen keyboard lock logic

IDisposable:
When the keyboard lock is supposed to be active (in full-screen mode), hitting the escape key (NOT long-pressing) **should NOT** dismiss the full-screen mode, and **should** send the Escape key through to the remote.

- Added awaits to the browser calls that need to complete in order.
- Cleaned up (mostly) duplicate code in the Absolute/Relative mouse handling.
- Ensure we don't overrun any existing keyboard lock or pointer lock.
- Release the keyboard lock when leaving full-screen.
- Per standards, we need to acquire the keyboard and pointer locks before entering full-screen or the user may get multiple messages about exiting.
- Fixed all the missing/excess React dependencies.
- Moved the PointerLockBar up so it is visible.
- Change PointerLockBar animations to ensure it is visible when needed.
- Fixed `readonly` warnings.

### DIFF
--- a/ui/src/components/VideoOverlay.tsx
+++ b/ui/src/components/VideoOverlay.tsx
@@ -10,7 +10,7 @@ import LoadingSpinner from "@components/LoadingSpinner";
 import Card, { GridCard } from "@components/Card";
 
 interface OverlayContentProps {
-  children: React.ReactNode;
+  readonly children: React.ReactNode;
 }
 function OverlayContent({ children }: OverlayContentProps) {
   return (
@@ -23,7 +23,7 @@ function OverlayContent({ children }: OverlayContentProps) {
 }
 
 interface LoadingOverlayProps {
-  show: boolean;
+  readonly show: boolean;
 }
 
 export function LoadingVideoOverlay({ show }: LoadingOverlayProps) {
@@ -57,8 +57,8 @@ export function LoadingVideoOverlay({ show }: LoadingOverlayProps) {
 }
 
 interface LoadingConnectionOverlayProps {
-  show: boolean;
-  text: string;
+  readonly show: boolean;
+  readonly text: string;
 }
 export function LoadingConnectionOverlay({ show, text }: LoadingConnectionOverlayProps) {
   return (
@@ -91,8 +91,8 @@ export function LoadingConnectionOverlay({ show, text }: LoadingConnectionOverla
 }
 
 interface ConnectionErrorOverlayProps {
-  show: boolean;
-  setupPeerConnection: () => Promise<void>;
+  readonly show: boolean;
+  readonly setupPeerConnection: () => Promise<void>;
 }
 
 export function ConnectionFailedOverlay({
@@ -153,7 +153,7 @@ export function ConnectionFailedOverlay({
 }
 
 interface PeerConnectionDisconnectedOverlay {
-  show: boolean;
+  readonly show: boolean;
 }
 
 export function PeerConnectionDisconnectedOverlay({
@@ -207,8 +207,8 @@ export function PeerConnectionDisconnectedOverlay({
 }
 
 interface HDMIErrorOverlayProps {
-  show: boolean;
-  hdmiState: string;
+  readonly show: boolean;
+  readonly hdmiState: string;
 }
 
 export function HDMIErrorOverlay({ show, hdmiState }: HDMIErrorOverlayProps) {
@@ -310,8 +310,8 @@ export function HDMIErrorOverlay({ show, hdmiState }: HDMIErrorOverlayProps) {
 }
 
 interface NoAutoplayPermissionsOverlayProps {
-  show: boolean;
-  onPlayClick: () => void;
+  readonly show: boolean;
+  readonly onPlayClick: () => void;
 }
 
 export function NoAutoplayPermissionsOverlay({
@@ -361,7 +361,7 @@ export function NoAutoplayPermissionsOverlay({
 }
 
 interface PointerLockBarProps {
-  show: boolean;
+  readonly show: boolean;
 }
 
 export function PointerLockBar({ show }: PointerLockBarProps) {
@@ -369,10 +369,10 @@ export function PointerLockBar({ show }: PointerLockBarProps) {
     <AnimatePresence mode="wait">
       {show ? (
         <motion.div
-          className="absolute -top-[36px] left-0 right-0 z-20 bg-white"
-          initial={{ y: 20, opacity: 0, zIndex: 0 }}
-          animate={{ opacity: 1, y: 0 }}
-          exit={{ y: 43, zIndex: 0 }}
+          className="flex w-full items-center justify-between bg-transparent"
+          initial={{ opacity: 0, zIndex: 0 }}
+          animate={{ opacity: 1, zIndex: 20 }}
+          exit={{ opacity: 0, zIndex: 0 }}
           transition={{ duration: 0.5, ease: "easeInOut", delay: 0.5 }}
         >
           <div>
--- a/ui/src/components/WebRTCVideo.tsx
+++ b/ui/src/components/WebRTCVideo.tsx
@@ -1,6 +1,15 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useResizeObserver } from "usehooks-ts";
 
+import VirtualKeyboard from "@components/VirtualKeyboard";
+import Actionbar from "@components/ActionBar";
+import MacroBar from "@/components/MacroBar";
+import InfoBar from "@components/InfoBar";
+import notifications from "@/notifications";
+import useKeyboard from "@/hooks/useKeyboard";
+import { useJsonRpc } from "@/hooks/useJsonRpc";
+import { cx } from "@/cva.config";
+import { keys, modifiers } from "@/keyboardMappings";
 import {
   useHidStore,
   useMouseStore,
@@ -9,15 +18,6 @@ import {
   useUiStore,
   useVideoStore,
 } from "@/hooks/stores";
-import { keys, modifiers } from "@/keyboardMappings";
-import { cx } from "@/cva.config";
-import VirtualKeyboard from "@components/VirtualKeyboard";
-import Actionbar from "@components/ActionBar";
-import MacroBar from "@/components/MacroBar";
-import InfoBar from "@components/InfoBar";
-import useKeyboard from "@/hooks/useKeyboard";
-import { useJsonRpc } from "@/hooks/useJsonRpc";
-import notifications from "@/notifications";
 
 import {
   HDMIErrorOverlay,
@@ -106,8 +106,9 @@ export default function WebRTCVideo() {
   );
 
   // Pointer lock and keyboard lock related
-  const isPointerLockPossible = window.location.protocol === "https:";
-
+  const isPointerLockPossible = window.location.protocol === "https:" || window.location.hostname === "localhost";
+  const isFullscreenEnabled = document.fullscreenEnabled;
+ 
   const checkNavigatorPermissions = useCallback(async (permissionName: string) => {
     const name = permissionName as PermissionName;
     const { state } = await navigator.permissions.query({ name });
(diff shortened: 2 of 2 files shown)


ui/src/components/VideoOverlay.tsx
```diff
@@ -10,7 +10,7 @@ import LoadingSpinner from "@components/LoadingSpinner";
 import Card, { GridCard } from "@components/Card";
 
 interface OverlayContentProps {
-  children: React.ReactNode;
+  readonly children: React.ReactNode;
```
IDisposable, line 13: There's a warning in VS Code that all these properties should be declared `readonly` (which they should), so while I was in here I updated them. We need to do a sweep through the UI to get the rest... another PR will arrive for ONLY warning-fixes.

ui/src/components/VideoOverlay.tsx
```diff
@@ -361,18 +361,18 @@ export function NoAutoplayPermissionsOverlay({
...
 export function PointerLockBar({ show }: PointerLockBarProps) {
   return (
     <AnimatePresence mode="wait">
       {show ? (
         <motion.div
-          className="absolute -top-[36px] left-0 right-0 z-20 bg-white"
-          initial={{ y: 20, opacity: 0, zIndex: 0 }}
-          animate={{ opacity: 1, y: 0 }}
-          exit={{ y: 43, zIndex: 0 }}
+          className="flex w-full items-center justify-between bg-transparent"
```
IDisposable, line 372: Animating the location just plain broke the layout and resulted in this button/action just being behind and invisible to the user. I couldn't get the thing centered the way I wanted, but in this form it animates into and out of visibility correctly. Also, setting the `motion.div` to`bg-transparent` helps ensure it doesn't look ugly.

ui/src/components/VideoOverlay.tsx
```diff
@@ -361,18 +361,18 @@ export function NoAutoplayPermissionsOverlay({
...
+          initial={{ opacity: 0, zIndex: 0 }}
+          animate={{ opacity: 1, zIndex: 20 }}
```
IDisposable, line 374: Even though the `div` was supposed to be `z-20`, not animating _to_ that destination left the div as `z-index: 0` and thus invisible. Just changing this was NOT enough because it was still positioned all wonky relative to the action bar.

ui/src/components/VideoOverlay.tsx
```diff
@@ -361,18 +361,18 @@ export function NoAutoplayPermissionsOverlay({
...
+          exit={{ opacity: 0, zIndex: 0 }}
```
IDisposable, line 375: If we fade in, we have to fade out :)

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
   );
 
   // Pointer lock and keyboard lock related
-  const isPointerLockPossible = window.location.protocol === "https:";
-
+  const isPointerLockPossible = window.location.protocol === "https:" || window.location.hostname === "localhost";
```
IDisposable, line 109: Also allowing pointer lock on "localhost" so we can test during development (this will not be an issue in deployed code)

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
+  const isFullscreenEnabled = document.fullscreenEnabled;
```
IDisposable, line 110: Let's have a single source of truth about _if_ we can offer/execute full-screen.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
+ 
   const checkNavigatorPermissions = useCallback(async (permissionName: string) => {
     const name = permissionName as PermissionName;
     const { state } = await navigator.permissions.query({ name });
     return state === "granted";
   }, []);
 
   const requestPointerLock = useCallback(async () => {
-    if (document.pointerLockElement) return;
+    if (!isPointerLockPossible
```
IDisposable, line 119: Just safer

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
+      || videoElm.current === null
+      || document.pointerLockElement) return;
 
     const isPointerLockGranted = await checkNavigatorPermissions("pointer-lock");
+
     if (isPointerLockGranted && settings.mouseMode === "relative") {
-      videoElm.current?.requestPointerLock();
+      await videoElm.current.requestPointerLock();
```
IDisposable, line 126: We _absolutely_ have to wait for the browser to honor the request for _pointer_ lock or things get wonky

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
     }
-  }, [checkNavigatorPermissions, settings.mouseMode]);
+  }, [checkNavigatorPermissions, isPointerLockPossible, settings.mouseMode]);
+
+  const requestKeyboardLock = useCallback(async () => {
```
IDisposable, line 130: Follow the same pattern for keyboard lock, so we're sure to get it right everywhere

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
+      }
+    }
+  }, [checkNavigatorPermissions]);
+
+  const releaseKeyboardLock = useCallback(async () => {
+    if (videoElm.current === null || document.fullscreenElement !== videoElm.current) return;
+
+    if ("keyboard" in navigator) {
+        // @ts-expect-error - keyboard unlock is not supported in all browsers
+        await navigator.keyboard.unlock();
```
IDisposable, line 147: When we leave full-screen we _absolutely_ have to `unlock()` the keyboard or subsequent attempts to `lock()` will throw an error and then we'll not be able to catch the `Escape` key

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
+    }
+  }, []);
 
   useEffect(() => {
     if (!isPointerLockPossible || !videoElm.current) return;
 
     const handlePointerLockChange = () => {
       if (document.pointerLockElement) {
-        notifications.success("Pointer lock Enabled, hold escape to exit");
+        notifications.success("Pointer lock Enabled, press escape to unlock");
```
IDisposable, line 156: This instruction is NOT correct. The press and hold is only relevant for _keyboard lock_, not pointer lock. Clarified.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
         setIsPointerLockActive(true);
       } else {
-        notifications.success("Pointer lock disabled");
+        notifications.success("Pointer lock Disabled");
```
IDisposable, line 159: If we **E**nabled we should **D**isabled

ui/src/components/WebRTCVideo.tsx
```diff
@@ -144,27 +169,39 @@ export default function WebRTCVideo() {
     return () => {
       abortController.abort();
     };
-  }, [isPointerLockPossible, videoElm]);
+  }, [isPointerLockPossible]);
 
   const requestFullscreen = useCallback(async () => {
-    videoElm.current?.requestFullscreen({
+     if (!isFullscreenEnabled || !videoElm.current) return;
```
IDisposable, line 175: This is hugely important... we need to request both the keyboard lock and pointer lock _before_ we transition to full-screen or things are just not stable.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -144,27 +169,39 @@ export default function WebRTCVideo() {
...
+
+    // per https://wicg.github.io/keyboard-lock/#system-key-press-handler
+    // If keyboard lock is activated after fullscreen is already in effect, then the user my 
+    // see multiple messages about how to exit fullscreen. For this reason, we recommend that
+    // developers call lock() before they enter fullscreen:
+    await requestKeyboardLock();
+    await requestPointerLock();
+
+    await videoElm.current.requestFullscreen({
```
IDisposable, line 184: ... and we have to wait for the transition to full-screen before moving on to other things...

ui/src/components/WebRTCVideo.tsx
```diff
@@ -144,27 +169,39 @@ export default function WebRTCVideo() {
...
       navigationUI: "show",
     });
+  }, [isFullscreenEnabled, requestKeyboardLock, requestPointerLock]);
 
-    // we do not care about pointer lock if it's for fullscreen
-    await requestPointerLock();
+  // setup to release the keyboard lock anytime the fullscreen ends
```
IDisposable, line 189: This block ensures that when the browser fires the event in response to the full-screen mode changing, we will release the keyboard lock.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -144,27 +169,39 @@ export default function WebRTCVideo() {
...
+  useEffect(() => {
+    if (!videoElm.current) return;
 
-    const isKeyboardLockGranted = await checkNavigatorPermissions("keyboard-lock");
-    if (isKeyboardLockGranted) {
-      if ("keyboard" in navigator) {
-        // @ts-expect-error - keyboard lock is not supported in all browsers
-        await navigator.keyboard.lock();
+    const handleFullscreenChange = () => {
+      if (!document.fullscreenElement) {
```
IDisposable, line 194: Per spec, by the time the event fires, the transition has _already_ completed.

IDisposable:
Before this change, on entering full screen you get the message `To exit full screen, press Esc` and if you hit an Escape key it _will_ leave full screen mode.
Before

After this change, on entering full screen you get the message `To exit full screen, press and hold Esc` and if you hit an Escape key it _will_ **NOT** leave full screen mode. 
After

ui/src/components/WebRTCVideo.tsx
ym: Good catch! I didn't realize that it could work on localhost. :-)

ui/src/components/WebRTCVideo.tsx
```diff
@@ -509,17 +526,14 @@ export default function WebRTCVideo() {
       document.addEventListener("keydown", keyDownHandler, { signal });
       document.addEventListener("keyup", keyUpHandler, { signal });
 
-      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-      // @ts-expect-error
-      window.clearKeys = () => sendKeyboardEvent([], []);
```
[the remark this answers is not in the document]
ym: We can probably use useRef or other methods to expose the API.@adamshiervani Any better idea?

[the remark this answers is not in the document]
IDisposable: If you have a use-case, I can restore that line... but there are NO references to `clearKeys` anywhere else in the code-base

ui/src/components/WebRTCVideo.tsx
```diff
@@ -106,32 +106,57 @@ export default function WebRTCVideo() {
...
 
   // Pointer lock and keyboard lock related
-  const isPointerLockPossible = window.location.protocol === "https:";
-
+  const isPointerLockPossible = window.location.protocol === "https:" || window.location.hostname === "localhost";
+  const isFullscreenEnabled = document.fullscreenEnabled;
+ 
   const checkNavigatorPermissions = useCallback(async (permissionName: string) => {
     const name = permissionName as PermissionName;
     const { state } = await navigator.permissions.query({ name });
```
literallylara, line 114: This will throw an error in browsers that do understand a permission, e.g. Firefox, causing the fullscreen button to not do anything. See: https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query

> Thrown if retrieving the PermissionDescriptor information failed in some way, or the permission doesn't exist or is unsupported by the user agent.

Screenshot 2025-06-19 at 22 11 23

IDisposable: I don't use anything but Chrome and Edge, so thanks for finding this. I can wrap that with a check.

IDisposable: Addressed in #631

literallylara: > I don't use anything but Chrome and Edge, so thanks for finding this. I can wrap that with a check.

Unfortunately it is still necessary to do cross-browser and cross-device testing in the world of web development today. I can recommend [caniuse.com](https://caniuse.com/).

IDisposable: PR #631 submitted

adrianmeraz:
@IDisposable Did this really get merged into the latest update? I don't see the changes, doesn't required a long press ESC

IDisposable:
It can only do a keyboard lock if you're hitting it in HTTPS, so enable that or use app.jetkvm.com to access the JetKVM

amatriain:
Still not working for me with Firefox on arch linux. App version 0.4.5, System version 0.2.5. Accessing jetKVM through the LAN, not cloud, with HTTPS mode enabled, using the self-signed cert option.

literallylara:
@amatriain Keyboard locking is in experimental state and not currently suppported by Firefox, see: https://caniuse.com/mdn-api_keyboard_lock

PR #600, opened 2025-06-12
chore: upgrade go from 1.24.3 to 1.24.4

ym:


SuperQ:
You can specify `1.24.x` and it will always use the latest Go patch release.

PR #604, opened 2025-06-12
feat: add command to show version

ym:


IDisposable (approved):
This looks good :)

SuperQ:
The `github.com/prometheus/common/version` package also includes a nice version output.

https://github.com/prometheus/common/blob/main/version/info.go#L40-L48

That's part of why I put the package in main.go, so it could be used later. :)

ym:
> The `github.com/prometheus/common/version` package also includes a nice version output.
> 
> https://github.com/prometheus/common/blob/main/version/info.go#L40-L48
> 
> That's part of why I put the package in main.go, so it could be used later. :)

Yea, I wanted to use this function, but it doesn't allow overriding `versionInfoTmpl`. I ended up creating something similar, though.

PR #606, opened 2025-06-12
chore: indicate jetkvm version from user-agent

ym:
TODO:

- [ ] Allow users to override the user agent and choose whether to send the version number.


IDisposable (approved):
This is helpful to using the POOL.NTP.ORG servers, and doesn't _really_ need to be configurable. Can we ship this as is and add the user-agent configurability to the network settings enhancements?

PR #609, opened 2025-06-13
fix/timesync: Make timesync  work with fallbacks and repolling

IDisposable:
Resolving issues with JetKVM devices not being able to contact any NTP servers to get time (and thus the clock is wrong, and thus the SSL negotiations fail, and thus auto update checks fail). This change can target DEV and MAIN.

- Added check to not attempt auto update if time sync is needed and not yet successful (delays 30 second to recheck).
- Added resync of time when DHCP or link state changes if online
- Added conditional* fallback from configured* NTP servers to the IP-named NTP servers, and then to the DNS named ones if that fails
- Added conditional* fallback from the configured* HTTP servers to the default DNS named ones.
- Uses the configuration* option for how many queries to run in parallel
- Added known static IPs for time servers (in case DNS resolution isn't up yet)
- Added time.cloudflare.com to fall-back NTP servers

* Note: The UI for configuring many of these things doesn't exist yet, but the defaults are reasonable

### DIFF
--- a/internal/timesync/timesync.go
+++ b/internal/timesync/timesync.go
@@ -158,6 +158,7 @@ func (t *TimeSync) Sync() error {
 	var (
 		now    *time.Time
 		offset *time.Duration
+		log    zerolog.Logger
 	)
 
 	metricTimeSyncCount.Inc()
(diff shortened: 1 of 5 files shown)


internal/timesync/timesync.go
```diff
@@ -159,11 +159,28 @@ func (t *TimeSync) Sync() error {
 	metricTimeSyncCount.Inc()
 
 	if syncMode.Ntp {
-		now, offset = t.queryNetworkTime()
+		// try the configured servers first
+		now, offset = t.queryNetworkTime(t.ntpServers)
+
+		if syncMode.NtpUseFallback && now == nil {
+			// now try the default hard-coded IP servers in case DNS is down
+			now, offset = t.queryNetworkTime(defaultNTPServerIPs)
```
IDisposable: Until we have configuration option for the preferred order  (and a UI to set it), we're just going to have to choose to either do IP or hostname first. Since IPs will always work even if DNS isn't up. This seems the best course.

See #625

network.go
```diff
@@ -53,7 +60,7 @@ func initNetwork() error {
 		},
 		OnConfigChange: func(networkConfig *network.NetworkConfig) {
 			config.NetworkConfig = networkConfig
-			networkStateChanged()
+			networkStateChanged(false)
```
IDisposable, line 63: If we're just reconfiguring things, likely the network isn't fully up yet, so we can wait for other events to fire.

IDisposable:
Rebased and tested

internal/timesync/timesync.go
```diff
@@ -188,7 +188,7 @@ Orders:
 		case "ntp":
 			if syncMode.Ntp && syncMode.NtpUseFallback {
 				t.l.Info().Msg("using NTP fallback")
-				now, offset = t.queryNetworkTime(defaultNTPServers)
+				now, offset = t.queryNetworkTime(defaultNTPServerIPs)
```
IDisposable, line 191: Renamed here because we now have the IP and DNS-named listed separately

IDisposable:
Fixed the two go lint issues formatting (trailing tabs on merged line) and not using the NTP fallback hostnames. Also now logs out the final new time (and mode that it was fetched via)

IDisposable:
Rebased to dev :shipit:

IDisposable:
Rebased :)

ym (approved):

PR #611, opened 2025-06-13
feat(ui): Add deviceName or deviceID to the browser tab title

IDisposable:
Addresses #304. In local mode will show the deviceId, in cloud connection will show the deviceName.

Would be nice to show the hostname, but that would require device-side code changes to the /device endpoint.

image

### DIFF
--- a/ui/src/routes/devices.$id.tsx
+++ b/ui/src/routes/devices.$id.tsx
@@ -58,9 +58,11 @@ import { SystemVersionInfo } from "./devices.$id.settings.general.update";
 
 interface LocalLoaderResp {
   authMode: "password" | "noPassword" | null;
+  deviceId: string;
 }
 
 interface CloudLoaderResp {
+  deviceId: string;
   deviceName: string;
   user: User | null;
   iceConfig: {
@@ -85,7 +87,7 @@ const deviceLoader = async () => {
   if (deviceRes.status === 401) return redirect("/login-local");
   if (deviceRes.ok) {
     const device = (await deviceRes.json()) as LocalDevice;
-    return { authMode: device.authMode };
+    return { authMode: device.authMode, deviceId: device.deviceId };
   }
 
   throw new Error("Error fetching device");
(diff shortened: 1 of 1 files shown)


IDisposable:
To make it show the hostname would require a device-side change (which I can easily do) for the /device endpoint, but that would modify the RPC responses, so we need someone like @ym or @adamshiervani to give their blessing.

 I did NOT change the top-left as suggested in #304 as it's actually an SVG image, not an H1... we certainly could insert the name right after that JetKVM logo as whatever, but would need a blessing and CSS styling from the same fine JetKVM employees.

adamshiervani:
Totally agree on the need to differentiate the device UI when managing multiple devices - it’s essential for clarity.

In cloud environments, this is relatively straightforward. Each cloud device already has a name, which we can start displaying in the UI - for example, in document.title and the menu bar.

The challenge is with local environments. What should we display there — the cloud name? The device ID? The hostname? Introducing a new property could help, but using the hostname as-is isn’t ideal, especially since it often needs to be a valid FQDN.

Things get more complicated when we have multiple data sources: one from the Cloud DB and one from the Local device. If there’s a collision between them, it can be quite confusing.

IDisposable:
You make the call what you want it to be, and which places (to right of logo and/or in connection status chip) and I'll happily pivot.

My _personal_ suggestion is that the cloud name be something that can be set in the _device_ config... as it is now, we have to delete/disassociate the JetKVM and readd in the cloud to change the name...

IDisposable:
@adamshiervani can we revisit this and decide what we want to do?

nsict-crj:
From my perspective (as someone not using the cloud) the three key places I want to see a configured hostname show up are, in decreasing importance:
- In the page title (which appears in the browser tab)
- On the login page
- In the DHCP lease name
- Somewhere prominent on the page when connected

I believe that currently (0.4.8) it only turns up in the lease name?

It feels to me, perhaps naïvely, that these ought all to be easy, and encompass everything people are asking for?

PR #616, opened 2025-06-14
Update hid_mouse_absolute.go

ronskvm:
Changed absolute mouse usb interface descriptor's subclass field to zero. If a usb mouse advertises itself as a boot device, it is *required* to reports its Button Status Value in Byte 0, its X Displacement Value in Byte 1, and its Y Displacement Value in Byte 2 when it is configured to operate in boot mode.

Although JetKVM's absolute mouse does always report its Button Status Value in Byte 0, it always reports its X Absolute Value in Byte 1 and Byte 2, and its Y Absolute Value in Byte 3 and Byte 4. As such, this incorrect code will definitely cause problems for operating environments that attempt to configure the absolute mouse to operate in boot mode.

P.S. I believe it is acceptable for the relative mouse to advertise itself as as boot device because it does indeed always report its Button Status Value in Byte 0, its X Displacement Value in Byte 1, and its Y Displacement Value in Byte 2 under all conditions at least in the current code.

### DIFF
--- a/internal/usbgadget/hid_mouse_absolute.go
+++ b/internal/usbgadget/hid_mouse_absolute.go
@@ -12,7 +12,7 @@ var absoluteMouseConfig = gadgetConfigItem{
 	configPath: []string{"hid.usb1"},
 	attrs: gadgetAttributes{
 		"protocol":      "2",
-		"subclass":      "1",
+		"subclass":      "0",
 		"report_length": "6",
 	},
 	reportDesc: absoluteMouseCombinedReportDesc,


ym (approved):

IDisposable:
I see the reference in [Section B.2 of the HID Spec](https://www.usb.org/sites/default/files/hid1_11.pdf)

That said, making only this change would be really wrong (if it really restricts X and Y to a single byte) would not work with the current logic that sends a scaled value between 0-32767 (which of course needs two bytes) from the UI.

See https://github.com/jetkvm/kvm/blob/dev/ui/src/components/WebRTCVideo.tsx#L243-L285

rmschooley:
No - You are looking at this issue from the wrong perspective. When the mouse advertises itself as a boot device, an operating environment that attempts to use the mouse as a boot device assumes the first byte returned by the mouse is its button status, the second byte is the x displacement, and the third byte is the y displacement. Since the absolute mouse does not send its data in this format, it most definitely should not be advertising itself as a boot device.

The relative mouse can and should advertise itself as a boot device since it does always send its data in the format required by a boot device. The relative mouse is the only mouse that has any chance of working in an operating environment that wants to use the mouse as a boot device.

IDisposable:
That makes perfect sense, thanks... so the relative mouse will be available to BOOT protocol and when  it switches over, either will be available. :)

rmschooley:
Correct.

Needless to say, this obviously means the relative mouse needs to be enabled and selected for use in the settings in order for a mouse to function in an operating environment that wants to use a mouse as a boot device.

In the future, it would be nice to add the code necessary to detect whether the operating environment wants to use a mouse as a boot device and adjust what the mouse reports based on this so we could advertise the absolute mouse as a boot device and use it as such in such an environment. But, I doubt anyone is going to make these non-trivial code adjustments any time soon.

IDisposable (approved):

ym (approved):

PR #361, opened 2025-04-15
network enhanecment / refactor

ym:
## UI/UX Improvements
- Added cloud connection status indicators on the display screen (NotConfigured, Connecting, Connected, Disconnected)
- Implemented a new Networking Tab

image

## Core Features
- **mDNS**: Refactored into a separate package with user configuration options
- **Time Synchronization**:
  - Moved to a separate package with user configuration options
  - Added performance metrics to monitor synchronization status
  - Implemented parallel request capability to improve boot speed
  - Added RTC clock support for time persistence during reboots (Note: current hardware lacks battery backup, so time will reset after power loss)

## Developer Tools
- Integrated Google's pprof endpoints for performance profiling
- Created a streaming log viewer using Server-Sent Events (SSE) at `/developer/log-stream`

PR #625, opened 2025-06-19
feat/DHCP NTP Servers - Enhances time sync with DHCP NTP and custom servers

IDisposable:
Extends time synchronization capabilities by:

- Adding support for obtaining NTP server addresses from DHCP leases.
- Enabling the use of user-defined NTP and HTTP servers for time synchronization.
- Introduces a configurable time sync ordering, allowing users to specify the preferred order of time sources.

### DIFF
--- a/internal/confparser/confparser_test.go
+++ b/internal/confparser/confparser_test.go
@@ -46,8 +46,8 @@ type testNetworkConfig struct {
 	TimeSyncOrdering        []string    `json:"time_sync_ordering,omitempty" one_of:"http,ntp,ntp_dhcp,ntp_user_provided,http_user_provided" default:"ntp,http"`
 	TimeSyncDisableFallback null.Bool   `json:"time_sync_disable_fallback,omitempty" default:"false"`
 	TimeSyncParallel        null.Int    `json:"time_sync_parallel,omitempty" default:"4"`
-	TimeSyncNTPServers      []string    `json:"time_sync_ntp_servers,omitempty" validate_type:"ipv4_or_ipv6" required_if:"TimeSyncOrdering=ntp_custom"`
-	TimeSyncHTTPUrls        []string    `json:"time_sync_http_urls,omitempty" validate_type:"url" required_if:"TimeSyncOrdering=http_custom"`
+	TimeSyncNTPServers      []string    `json:"time_sync_ntp_servers,omitempty" validate_type:"ipv4_or_ipv6" required_if:"TimeSyncOrdering=ntp_user_provided"`
+	TimeSyncHTTPUrls        []string    `json:"time_sync_http_urls,omitempty" validate_type:"url" required_if:"TimeSyncOrdering=http_user_provided"`
 }
 
 func TestValidateConfig(t *testing.T) {
--- a/internal/confparser/confparser_test.go
+++ b/internal/confparser/confparser_test.go
@@ -43,9 +43,11 @@ type testNetworkConfig struct {
 	LLDPTxTLVs              []string    `json:"lldp_tx_tlvs,omitempty" one_of:"chassis,port,system,vlan" default:"chassis,port,system,vlan"`
 	MDNSMode                null.String `json:"mdns_mode,omitempty" one_of:"disabled,auto,ipv4_only,ipv6_only" default:"auto"`
 	TimeSyncMode            null.String `json:"time_sync_mode,omitempty" one_of:"ntp_only,ntp_and_http,http_only,custom" default:"ntp_and_http"`
-	TimeSyncOrdering        []string    `json:"time_sync_ordering,omitempty" one_of:"http,ntp,ntp_dhcp,ntp_user_provided,ntp_fallback" default:"ntp,http"`
+	TimeSyncOrdering        []string    `json:"time_sync_ordering,omitempty" one_of:"http,ntp,ntp_dhcp,ntp_user_provided,http_user_provided" default:"ntp,http"`
 	TimeSyncDisableFallback null.Bool   `json:"time_sync_disable_fallback,omitempty" default:"false"`
 	TimeSyncParallel        null.Int    `json:"time_sync_parallel,omitempty" default:"4"`
+	TimeSyncNTPServers      []string    `json:"time_sync_ntp_servers,omitempty" validate_type:"ipv4_or_ipv6" required_if:"TimeSyncOrdering=ntp_custom"`
+	TimeSyncHTTPUrls        []string    `json:"time_sync_http_urls,omitempty" validate_type:"url" required_if:"TimeSyncOrdering=http_custom"`
 }
 
 func TestValidateConfig(t *testing.T) {
--- a/internal/network/netif.go
+++ b/internal/network/netif.go
@@ -21,6 +21,7 @@ type NetworkInterfaceState struct {
 	ipv6Addr      *net.IP
 	ipv6Addresses []IPv6Address
 	ipv6LinkLocal *net.IP
+	ntpAddresses  []*net.IP
 	macAddr       *net.HardwareAddr
 
 	l         *zerolog.Logger
@@ -76,6 +77,7 @@ func NewNetworkInterfaceState(opts *NetworkInterfaceOptions) (*NetworkInterfaceS
 		onInitialCheck:  opts.OnInitialCheck,
 		cbConfigChange:  opts.OnConfigChange,
 		config:          opts.NetworkConfig,
+		ntpAddresses:    make([]*net.IP, 0),
 	}
 
 	// create the dhcp client
@@ -89,7 +91,7 @@ func NewNetworkInterfaceState(opts *NetworkInterfaceOptions) (*NetworkInterfaceS
 				opts.Logger.Error().Err(err).Msg("failed to update network state")
 				return
 			}
-
+			_ = s.updateNtpServersFromLease(lease)
 			_ = s.setHostnameIfNotSame()
 
 			opts.OnDhcpLeaseChange(lease)
@@ -133,6 +135,27 @@ func (s *NetworkInterfaceState) IPv6String() string {
 		return "..."
 	}
 	return s.ipv6Addr.String()
+}
+
+func (s *NetworkInterfaceState) NtpAddresses() []*net.IP {
+	return s.ntpAddresses
+}
+
+func (s *NetworkInterfaceState) NtpAddressesString() []string {
+	ntpServers := []string{}
+
+	if s != nil {
+		s.l.Debug().Any("s", s).Msg("getting NTP address strings")
+
+		if len(s.ntpAddresses) > 0 {
+			for _, server := range s.ntpAddresses {
+				s.l.Debug().IPAddr("server", *server).Msg("converting NTP address")
+				ntpServers = append(ntpServers, server.String())
+			}
+		}
+	}
+
+	return ntpServers
 }
 
 func (s *NetworkInterfaceState) MAC() *net.HardwareAddr {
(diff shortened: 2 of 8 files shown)


IDisposable:
Fleshing out the device-side support for the configuration options added in #361

Still need the configuration UI to be finished, but this respects all the settings allowed in the config file https://github.com/jetkvm/kvm/pull/361/files#diff-5850db209b26ffd9fd3630b5255df6e30442d6cdfd7bcb6174c60924608dcc45R47-R50

internal/timesync/http.go
```diff
@@ -19,9 +19,9 @@ var defaultHTTPUrls = []string{
 	// "http://www.msftconnecttest.com/connecttest.txt",
 }
 
-func (t *TimeSync) queryAllHttpTime() (now *time.Time) {
-	chunkSize := 4
-	httpUrls := t.httpUrls
+func (t *TimeSync) queryAllHttpTime(httpUrls []string) (now *time.Time) {
+	chunkSize := int(t.networkConfig.TimeSyncParallel.ValueOr(4))
```
IDisposable, line 23: Honors the setting of the number of parallel time-sync requests (defaults to 4)

internal/network/netif.go
```diff
@@ -89,7 +91,7 @@ func NewNetworkInterfaceState(opts *NetworkInterfaceOptions) (*NetworkInterfaceS
 				opts.Logger.Error().Err(err).Msg("failed to update network state")
 				return
 			}
-
+			_ = s.updateNtpServersFromLease(lease)
```
IDisposable, line 94: Every time we get a new DHCP lease, we need to copy over the NTP servers (if any) to the `NetworkInterfaceState` for access inside the time-sync logic.

internal/confparser/confparser_test.go
```diff
@@ -43,9 +43,11 @@ type testNetworkConfig struct {
 	LLDPTxTLVs              []string    `json:"lldp_tx_tlvs,omitempty" one_of:"chassis,port,system,vlan" default:"chassis,port,system,vlan"`
 	MDNSMode                null.String `json:"mdns_mode,omitempty" one_of:"disabled,auto,ipv4_only,ipv6_only" default:"auto"`
 	TimeSyncMode            null.String `json:"time_sync_mode,omitempty" one_of:"ntp_only,ntp_and_http,http_only,custom" default:"ntp_and_http"`
-	TimeSyncOrdering        []string    `json:"time_sync_ordering,omitempty" one_of:"http,ntp,ntp_dhcp,ntp_user_provided,ntp_fallback" default:"ntp,http"`
+	TimeSyncOrdering        []string    `json:"time_sync_ordering,omitempty" one_of:"http,ntp,ntp_dhcp,ntp_user_provided,http_user_provided" default:"ntp,http"`
```
IDisposable, line 46: Since anyone who ran a version that included the previous code would have gotten the defaults written to their device's config file, we should behave the same way as before until they change the setting (there's still no UI for that, but that's a different PR).

So, the old behavior was to try NTP on the default servers, then HTTP on the default servers. This means that we should interpret the saved config line for `ntp,http` as if they requested the default servers. 

Thus `ntp` means the _prior_ default NTP servers and we need `ntp_user_provided` to specify user-supplied values.
The same logic applies to the HTTP URLs, so `http` refers to the _prior_ default list of HTTP URLs.

Ideally, we probably would like the default ordering to be `ntp_dhcp`. then `ntp`, then `http` but that ship has sailed for anyone running 0.4.x. We _might_ want to change the defaults here for future users.

Thus we added the `http_user_provided` option and deleted the redundant `ntp_fallback` option here to match

internal/confparser/confparser_test.go
```diff
@@ -43,9 +43,11 @@ type testNetworkConfig struct {
...
 	TimeSyncDisableFallback null.Bool   `json:"time_sync_disable_fallback,omitempty" default:"false"`
 	TimeSyncParallel        null.Int    `json:"time_sync_parallel,omitempty" default:"4"`
+	TimeSyncNTPServers      []string    `json:"time_sync_ntp_servers,omitempty" validate_type:"ipv4_or_ipv6" required_if:"TimeSyncOrdering=ntp_user_provided"`
```
IDisposable, line 49: These are where the user's user-provided NTP Server and HTTP URL values can be store. The `required_if` is likely not completely correct but changing the confparser seems way out of scope here. It will do the right thing _iff_ the user selected only `ntp_user_provided` in the `TimeSyncOrdering` but not if they have it included in the string somewhere. I think @ym might know how to express that (includes not equals), but not there now.

internal/timesync/ntp.go
```diff
@@ -46,6 +47,10 @@ type ntpResult struct {
 
 func (t *TimeSync) queryMultipleNTP(servers []string, timeout time.Duration) (now *time.Time, offset *time.Duration) {
 	results := make(chan *ntpResult, len(servers))
+
+	_, cancel := context.WithTimeout(context.Background(), timeout)
```
IDisposable, line 51: Add logic to cancel all outstanding NTP queries when we process a valid reply or end this chunk. This parallels logic in the HTTP path, and might not be desired.

internal/timesync/ntp.go
```diff
@@ -66,15 +71,25 @@ func (t *TimeSync) queryMultipleNTP(servers []string, timeout time.Duration) (no
 				return
 			}
 
+			if response.IsKissOfDeath() {
```
IDisposable, line 74: We really need to make sure the NTP server we were talking to didn't tell us to go away. https://support.ntp.org/Dev/KoDResponses

internal/timesync/timesync.go
```diff
@@ -64,14 +63,13 @@ func NewTimeSync(opts *TimeSyncOptions) *TimeSync {
 	}
 
 	t := &TimeSync{
-		syncLock:      &sync.Mutex{},
-		l:             opts.Logger,
-		rtcDevicePath: rtcDevice,
-		rtcLock:       &sync.Mutex{},
-		preCheckFunc:  opts.PreCheckFunc,
-		ntpServers:    defaultNTPServers,
```
IDisposable, line 72: We don't need these two `ntpServers` and `httpUrls` in the `TimeSync` struct as they're available to the time-sync code.

internal/timesync/timesync.go
```diff
@@ -82,34 +80,42 @@ func NewTimeSync(opts *TimeSyncOptions) *TimeSync {
...
 
+func (t *TimeSync) SetDhcpNtpAddresses(addresses []string) {
+	t.dhcpNtpAddresses = addresses
+}
+
 func (t *TimeSync) getSyncMode() SyncMode {
 	syncMode := SyncMode{
+		Ntp:             true,
+		Http:            true,
+		Ordering:        []string{"ntp_dhcp", "ntp", "http"},
```
IDisposable, line 91: This default ordering is a lie, unless there was no stored config, because that will overwrite this defaulting... but just in case, use the DHCP provided NTP servers, then the NTP _default_ servers, then the _default_ HTTP URLs

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
 
-	if syncMode.Ntp {
-		now, offset = t.queryNetworkTime()
-	}
+	syncMode := t.getSyncMode()
 
-	if syncMode.Http && now == nil {
-		now = t.queryAllHttpTime()
+Orders:
+	for _, mode := range syncMode.Ordering {
```
IDisposable, line 168: Do the options in the order they specified.

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+		switch mode {
+		case "ntp_user_provided":
+			if syncMode.Ntp {
+				t.l.Info().Msg("using NTP custom servers")
+				now, offset = t.queryNetworkTime(t.networkConfig.TimeSyncNTPServers)
```
IDisposable, line 173: NTP servers from the config (will be empty until the UI is coded)

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+				if now != nil {
+					t.l.Info().Str("source", "NTP").Time("now", *now).Msg("time obtained")
+					break Orders
+				}
+			}
+		case "ntp_dhcp":
+			if syncMode.Ntp {
+				t.l.Info().Msg("using NTP servers from DHCP")
+				now, offset = t.queryNetworkTime(t.dhcpNtpAddresses)
```
IDisposable, line 182: NTP servers from the DHCP lease

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+				if now != nil {
+					t.l.Info().Str("source", "NTP DHCP").Time("now", *now).Msg("time obtained")
+					break Orders
+				}
+			}
+		case "ntp":
+			if syncMode.Ntp && syncMode.NtpUseFallback {
+				t.l.Info().Msg("using NTP fallback")
+				now, offset = t.queryNetworkTime(defaultNTPServers)
```
IDisposable, line 191: NTP servers from the code-defined defaults (fallbacks)

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+				now = t.queryAllHttpTime(t.networkConfig.TimeSyncHTTPUrls)
+				if now != nil {
+					t.l.Info().Str("source", "HTTP").Time("now", *now).Msg("time obtained")
+					break Orders
+				}
+			}
+		case "http":
+			if syncMode.Http && syncMode.HttpUseFallback {
+				t.l.Info().Msg("using HTTP fallback")
+				now = t.queryAllHttpTime(defaultHTTPUrls)
```
IDisposable, line 209: HTTP URLs from the code-defined defaults (fallbacks)

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+				now, offset = t.queryNetworkTime(defaultNTPServers)
+				if now != nil {
+					t.l.Info().Str("source", "NTP fallback").Time("now", *now).Msg("time obtained")
+					break Orders
+				}
+			}
+		case "http_user_provided":
+			if syncMode.Http {
+				t.l.Info().Msg("using HTTP custom URLs")
+				now = t.queryAllHttpTime(t.networkConfig.TimeSyncHTTPUrls)
```
IDisposable, line 200: HTTP URLs from the config (will be empty until the UI is coded)

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+			if syncMode.Ntp && syncMode.NtpUseFallback {
+				t.l.Info().Msg("using NTP fallback")
+				now, offset = t.queryNetworkTime(defaultNTPServers)
+				if now != nil {
+					t.l.Info().Str("source", "NTP fallback").Time("now", *now).Msg("time obtained")
+					break Orders
+				}
+			}
+		case "http_user_provided":
+			if syncMode.Http {
```
IDisposable, line 198: Note, that `http` means the _prior_ HTTP URL default list.

internal/timesync/timesync.go
```diff
@@ -154,16 +160,61 @@ func (t *TimeSync) Sync() error {
...
+		case "ntp_dhcp":
+			if syncMode.Ntp {
+				t.l.Info().Msg("using NTP servers from DHCP")
+				now, offset = t.queryNetworkTime(t.dhcpNtpAddresses)
+				if now != nil {
+					t.l.Info().Str("source", "NTP DHCP").Time("now", *now).Msg("time obtained")
+					break Orders
+				}
+			}
+		case "ntp":
```
IDisposable, line 188: Note that `ntp` means the _prior_ default NTP server list.

network.go
```diff
@@ -19,8 +19,17 @@ func networkStateChanged() {
 	// do not block the main thread
 	go waitCtrlAndRequestDisplayUpdate(true)
 
+	if timeSync != nil {
```
IDisposable, line 22: These `nil` checks are necessary because we can hit `networkStateChanged` before we've even finished setting up.

IDisposable:
This is step one of two to enable all the rest of the NTP configurable behavior that was started in #361. Hopefully this is helpful.

IDisposable:
If this is merged first, PR #609 will need to be adjusted (or just closed)

PR #631, opened 2025-06-19
fix(ui)/firefox permissions error handling

IDisposable:
Per https://github.com/jetkvm/kvm/pull/535#discussion_r2157613450 we need to handle Firefox/Mozilla browsers not having these pointer-lock or keyboard-lock permissions checks. Also might as well not throw if something goes wrong when trying to check permissions or apply for locks.

### DIFF
--- a/ui/src/components/WebRTCVideo.tsx
+++ b/ui/src/components/WebRTCVideo.tsx
@@ -115,9 +115,18 @@ export default function WebRTCVideo() {
   const isFullscreenEnabled = document.fullscreenEnabled;
  
   const checkNavigatorPermissions = useCallback(async (permissionName: string) => {
-    const name = permissionName as PermissionName;
-    const { state } = await navigator.permissions.query({ name });
-    return state === "granted";
+    if (!navigator.permissions || !navigator.permissions.query) {
+      return false; // if can't query permissions, assume NOT granted
+    }
+
+    try {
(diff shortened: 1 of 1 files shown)


ui/src/components/WebRTCVideo.tsx
```diff
@@ -115,9 +115,18 @@ export default function WebRTCVideo() {
...
+    }
+
+    try {
+      const name = permissionName as PermissionName;
+      const { state } = await navigator.permissions.query({ name });
+      return state === "granted";
+    } catch {
+      // ignore errors
+    }
+    return false; // if query fails, assume NOT granted
```
literallylara, line 129: Thanks for the quick fix! It may be a matter of preference, but this could be simplified to:

```ts
if (!navigator.permissions?.query) return false;

const name = permissionName as PermissionName;
const { state } = await navigator.permissions.query({ name }).catch(_ => ({ state: null }));

return state === "granted";
```

As well as further down:

```ts
try { await somePromiseReturningFunction() } catch {}
// could be simplified to:
await somePromiseReturningFunction().catch(_=>null)
```

See: 
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

PR #639, opened 2025-06-23
fix: ensure certStore is initialised when adding custom cert

### LINKED ISSUE #612: Issue trying to add custom certificate to TLS settings (Experimental)
Updated to latest version of JetKVM - 0.4.3.  I noticed under Settings -> Access there is a new setting for HTTPS Mode (Experimental).

Just wanted to report that I tried to test with my own custom certificate (it's a wildcard cert for my domain).  I put in the certificate information and the private key, but when I press "Update TLS Settings" I get the following error:

Image

One of the things I did notice that in the "sample" text for teh Private Key, you have 

START PRIVATE KEY
END PRIVATE KEY

When my key file starts and ends with 

START RSA PRIVATE KEY
END RSA PRIVATE KEY

Not sure if that's the issue, but not sure what's causing the issue.  I can't save or update the certificate.

dseven:
Fixes: #612

### DIFF
--- a/web_tls.go
+++ b/web_tls.go
@@ -108,6 +108,9 @@ func setTLSState(s TLSState) error {
 			isChanged = true
 		}
 		// parse pem to cert and key
+		if certStore == nil {
+			initCertStore()
+		}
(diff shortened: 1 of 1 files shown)


IDisposable (approved):
:shipit:

PR #672, opened 2025-07-06
feat(dc-power-extension) Power restore mode in DCPowerControl component

PH89:
With this feature its possible to set a power restore mode for the DC Power extension module. 
There are three options that can be applied by the module in case the device comes back after a power loss:

1. **Power OFF:** which is the default and leaves the DC output turned off. 
2. **Power ON:** which turns the DC output always on.
3. **Last state:** which restores the state that the DC output was before the power loss took place.

To make this work, there is a firmware update for the DC module necessary. See https://github.com/jetkvm/dc-extension-firmware/pull/3  

**This is how its look like:**
DC Restore on Power Loss

Feedback always welcome

### DIFF
--- a/jsonrpc.go
+++ b/jsonrpc.go
@@ -681,10 +681,11 @@ func rpcResetConfig() error {
 }
 
 type DCPowerState struct {
-	IsOn    bool    `json:"isOn"`
-	Voltage float64 `json:"voltage"`
-	Current float64 `json:"current"`
-	Power   float64 `json:"power"`
+	IsOn         bool    `json:"isOn"`
+	Voltage      float64 `json:"voltage"`
+	Current      float64 `json:"current"`
+	Power        float64 `json:"power"`
+	RestoreState int     `json:"restoreState"`
 }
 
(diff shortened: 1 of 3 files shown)


cdalvaro:
Please, @jetkvm team, take care of https://github.com/jetkvm/dc-extension-firmware/pull/3. It is necessary for this PR to work.

I think this feature is really necessary.

kromsam:
I'm on...
System: 0.2.7
App: 0.4.9

But I don't have this feature yet. Is it already rolled out?

PH89:
@kromsam Please make sure you installed the latest firmware update for the DC Power Control Extension. You need to flash this firmware update manually by using a usb-c cable.

Find the update uf2 file here:
https://github.com/jetkvm/dc-extension-firmware/releases/tag/0.2.0

walle89:
If anyone else, like me, was looking for the instructions for updating the JetKVM DC Power Control Extension, they’re available here: [Updates - DC Extension](https://jetkvm.com/docs/advanced-usage/ota-updates#dc-extension)

PR #686, opened 2025-07-14
chore: add issue templates

ym:


cdccentral-sourcecontrol:
Yes, the Dynatrace OneAgent installer specified with the arch=x86 parameter
in the download URL is designed for 64-bit x86 architectures, commonly
referred to as x86-64 or amd64. This parameter does not support 32-bit x86
systems, as Dynatrace focuses on modern 64-bit environments. For instance,
the download command typically uses a format such as
https://your-environment-id.live.dynatrace.com/api/v2/deployment/installer/agent/unix/default/latest?arch=x86,
which targets amd64-compatible processors from both Intel and AMD.


Matthew Rutledge
Cloud Systems Architect

CDCCentral
Ballston Lake, NY
Voice: 845-481-0097
www.cdccentral.com

cdccentral-sourcecontrol:
sorry


Matthew Rutledge
Cloud Systems Architect

CDCCentral
Ballston Lake, NY
Voice: 845-481-0097
www.cdccentral.com

PR #718, opened 2025-08-02
JetKVM Advanced, CGO-based 2-way Audio Support

pennycoders:
## Summary

This PR introduces **bidirectional audio support** to JetKVM, enabling both audio output (listening to the managed device) and audio input (microphone from browser to device). Audio is implemented using an **in-process CGO architecture** that directly calls C code for ALSA audio capture/playback and Opus encoding/decoding. The managed device presents itself as a **USB Audio Class 1 (UAC1) gadget** providing both stereo speakers and a stereo microphone interface over USB.

**Key Features:**
- Bidirectional stereo audio (48kHz, 16-bit, 2 channels)
- In-process CGO implementation for low latency and simplicity
- USB Audio Gadget (UAC1) integration
- WebRTC-based real-time streaming with Opus codec
- Frontend controls for enabling/disabling audio output and input
- HDMI or USB audio capture source selection
- SDP munging for proper stereo audio support in browsers

## Credits

- Thank you, @lqs as your #446 PR has actually encouraged me to pursue this

Thanks!
Alex

Issue #694, opened 2025-07-16
Add support for input type password for clipboard


xize:
### A note for the community

> [!NOTE]
> Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.

### Disclaimer

- [x] I have read and understood the disclaimer.
- [ ] I plan to implement the feature myself.

### Subsystem

None

### Feature description

Hi,

I would like to request a option for a checkbox to alter the textarea in the clipboard window with javascript to turn it into a password field, this could be quite usefull for when pasting a Bitlocker or Veracrypt bootloader password.

Currently I don't think it is easy to implement such feature when I look into ``/ui/src/components/popovers/PasteModal.tsx`` the javascript logic might be easier to implement in order to switch between element by DOM manipulation, but I think that a new component has to be written for the password field, using a textarea is a no go because the element doesn't support ``*`` masking.

My logic of GO isn't very good, but I can read it, there might be a few components of interest:
``/ui/src/components/AuthLayout.tsx`` <- likely not a good place.
``/ui/src/components/Fieldset.tsx`` <- this seem the base of input fields or textareas perhaps a new component needs to extend this.

Thank you very much!, have a nice day 👍

adamshiervani:
Thanks for the feature request! We've put up a draft fix in #1353 that adds a hide/show toggle to the paste modal. When enabled, the input switches to a password field so sensitive text like passwords isn't visible on screen. It'll go out with the next dev channel release and, once validated, promoted to stable.

PR #725, opened 2025-08-07
Rework keyboard management to allow device-side tracking of modifier states

IDisposable:
This is the beginnings of the rework of the keyboard support. 

## Goals:

1. Remove client/browser-side tracking of meta key state (Shift, Control, Alt, Super)
1. Add device-side (go code) to track the meta key state and communicate it back to the client (for Virtual Keyboard layout tracking)
1. Remove option to select LED tracking for latching key state (Caps Lock, Num Lock, Scroll Lock, Compose, Kana) since we have to trust the device
1. Ensure the device tracks the LED state at all times
1. Change the handling of keys in WebRTCVideo to use `reportKeypress` to pass all the keydown / keyup events so it can track the meta state and LED state
1. Allow for multiple-key-down (up to the standard 6 key-rollover of HID keyboards)


IDisposable:
@adamshiervani looks like this is mostly complete to get the state managed device-side. There was a _lot_ of work done to make the type-checking better because I had trouble working out some issues, so I just hardened up the code to get it to the point where the error was "obvious".  Sadly in the end it was some odd interaction between Go's default serialization of uint8/byte values to JSON, which if given an array slice of uint8/byte will emit the slice as a base-64 encoded blob... on the typescript side, that _doesn't_ get recognized as an array, so passing the _keys-down_  wasn't working. In the end, I added a `ByteSlice` type on the Go side, and forced it to serialize as an array... sigh...

Now that I'm mostly done, might want to chat through if I did this the way we want.  One thing I've realized just today is that if the _client UI_ tries to make the new rpc calls to the device and it's not current.. well.  Might have to restore the code I elided in commit 0c766343e4cd175c760cf3e77b28a992f6bf9151 and do some device-version detection?

internal/usbgadget/hid_keyboard.go
```diff
@@ -61,14 +61,18 @@ var keyboardReportDesc = []byte{
...
 	// https://www.usb.org/sites/default/files/documents/hid1_11.pdf
 	// https://www.usb.org/sites/default/files/hut1_2.pdf
 	KeyboardLedMaskNumLock    = 1 << 0
 	KeyboardLedMaskCapsLock   = 1 << 1
 	KeyboardLedMaskScrollLock = 1 << 2
 	KeyboardLedMaskCompose    = 1 << 3
 	KeyboardLedMaskKana       = 1 << 4
-	ValidKeyboardLedMasks     = KeyboardLedMaskNumLock | KeyboardLedMaskCapsLock | KeyboardLedMaskScrollLock | KeyboardLedMaskCompose | KeyboardLedMaskKana
+	// power on/off LED is 5
+	KeyboardLedMaskShift  = 1 << 6
```
IDisposable, line 74: Some keyboards are documented to have a _hardware_ shift LED, so there's support for that now

display.go
```diff
@@ -30,7 +30,7 @@ const (
 // do not call this function directly, use switchToScreenIfDifferent instead
 // this function is not thread safe
 func switchToScreen(screen string) {
-	_, err := CallCtrlAction("lv_scr_load", map[string]interface{}{"obj": screen})
+	_, err := CallCtrlAction("lv_scr_load", map[string]any{"obj": screen})
```
IDisposable, line 33: Here and all over, I ran go modernizer because I got tired of the warnings :)

internal/udhcpc/parser.go
```diff
@@ -101,7 +101,7 @@ func (l *Lease) SetLeaseExpiry() (time.Time, error) {
 func UnmarshalDHCPCLease(lease *Lease, str string) error {
 	// parse the lease file as a map
 	data := make(map[string]string)
-	for _, line := range strings.Split(str, "\n") {
+	for line := range strings.SplitSeq(str, "\n") {
```
IDisposable, line 104: ... also a Modernizer update

internal/udhcpc/parser.go
```diff
@@ -165,7 +165,7 @@ func UnmarshalDHCPCLease(lease *Lease, str string) error {
 			field.Set(reflect.ValueOf(ip))
 		case []net.IP:
 			val := make([]net.IP, 0)
-			for _, ipStr := range strings.Fields(value) {
+			for ipStr := range strings.FieldsSeq(value) {
```
IDisposable, line 168: ... also a Modernizer update

internal/usbgadget/hid_keyboard.go
```diff
@@ -91,27 +96,27 @@ func getKeyboardState(b byte) KeyboardState {
...
 	}
 
-	newState := getKeyboardState(b)
-	if reflect.DeepEqual(u.keyboardState, newState) {
+	if u.keyboardState == state {
 		return
 	}
-	u.log.Info().Interface("old", u.keyboardState).Interface("new", newState).Msg("keyboardState updated")
-	u.keyboardState = newState
+	u.log.Trace().Interface("old", u.keyboardState).Interface("new", state).Msg("keyboardState updated")
```
IDisposable: ... a bit too chatty ;)

internal/usbgadget/hid_keyboard.go
```diff
@@ -159,7 +209,7 @@ func (u *UsbGadget) listenKeyboardEvents() {
 				}
 				u.resetLogSuppressionCounter("keyboardHidFileRead")
 
-				l.Trace().Int("n", n).Bytes("buf", buf).Msg("got data from keyboard")
+				l.Trace().Int("n", n).Uints8("buf", buf).Msg("got data from keyboard")
```
IDisposable, line 195: `Bytes` emits this as a string, ugh... use Uints8 to see the array of bytes.

internal/usbgadget/hid_keyboard.go
```diff
@@ -195,12 +245,12 @@ func (u *UsbGadget) OpenKeyboardHidFile() error {
 	return u.openKeyboardHidFile()
 }
 
-func (u *UsbGadget) keyboardWriteHidFile(data []byte) error {
+func (u *UsbGadget) keyboardWriteHidFile(modifier byte, keys []byte) error {
```
IDisposable, line 231: Made this explicit because we have the state split up anyway

internal/usbgadget/hid_keyboard.go
```diff
@@ -211,22 +261,112 @@ func (u *UsbGadget) keyboardWriteHidFile(data []byte) error {
...
+		keys = append(keys, make([]byte, hidKeyBufferSize-len(keys))...)
 	}
 
-	err := u.keyboardWriteHidFile([]byte{modifier, 0, keys[0], keys[1], keys[2], keys[3], keys[4], keys[5]})
-	if err != nil {
-		return err
+	return u.keyboardWriteHidFile(modifier, keys)
+}
+
+const (
```
IDisposable, line 283: the device is now responsible for tracking the modifier state, so it needs the keycodes for those (from the HID spec)

internal/usbgadget/hid_keyboard.go
```diff
@@ -211,22 +261,112 @@ func (u *UsbGadget) keyboardWriteHidFile(data []byte) error {
...
+		}
 	}
 
-	u.resetUserInputTime()
-	return nil
+	if err := u.keyboardWriteHidFile(modifier, keys); err != nil {
+		u.log.Warn().Uint8("modifier", modifier).Uints8("keys", keys).Msg("Could not write keypress report to hidg0")
+	}
+
+	var result = KeysDownState{
```
IDisposable: We intentionally construct a whole new KeyDownState so the modification detector sees things change!

internal/usbgadget/usbgadget.go
```diff
@@ -41,6 +41,11 @@ var defaultUsbGadgetDevices = Devices{
 	MassStorage:   true,
 }
 
+type KeysDownState struct {
+	Modifier byte      `json:"modifier"`
+	Keys     ByteSlice `json:"keys"`
```
IDisposable, line 46: A `ByteSlice` is just a `[]byte` slice that has explicit json marshalling

jsonrpc.go
```diff
@@ -561,30 +566,40 @@ func riskyCallRPCHandler(handler RPCHandler, params map[string]interface{}) (int
...
+			}
 		}
 		return results[0].Interface(), nil
 	}
 
-	return nil, errors.New("unexpected return values from handler")
+	return nil, fmt.Errorf("too many return values from handler: %d", len(results))
+}
+
+func asError(value reflect.Value) (bool, error) {
```
IDisposable, line 594: Refactored out some "largely identical" error extraction

jsonrpc.go
```diff
@@ -1047,6 +1062,8 @@ var rpcHandlers = map[string]RPCHandler{
 	"renewDHCPLease":         {Func: rpcRenewDHCPLease},
 	"keyboardReport":         {Func: rpcKeyboardReport, Params: []string{"modifier", "keys"}},
 	"getKeyboardLedState":    {Func: rpcGetKeyboardLedState},
+	"keypressReport":         {Func: rpcKeypressReport, Params: []string{"key", "press"}},
```
IDisposable, line 1064: The two new functions! `keypressReport` reports when a key is pressed or released, and returns the current KeysDownState (with up to 6 key-rollover, and current modifier state)

`getKeyDownState` fetches the device's current state so we can sync up the UI at session startup.

remote_mount.go
```diff
@@ -27,10 +27,7 @@ func (w *WebRTCDiskReader) Read(ctx context.Context, offset int64, size int64) (
 	}
 	mountedImageSize := currentVirtualMediaState.Size
 	virtualMediaStateMutex.RUnlock()
-	end := offset + size
-	if end > mountedImageSize {
-		end = mountedImageSize
-	}
+	end := min(offset+size, mountedImageSize)
```
IDisposable, line 30: ... also a Modernizer update

ui/src/components/InfoBar.tsx
```diff
@@ -175,6 +170,11 @@ export default function InfoBar() {
               Kana
             </div>
           ) : null}
+          {keyboardLedState?.shift ? (
```
IDisposable: This is the LED shift state on some keyboards

ui/src/components/VirtualKeyboard.tsx
```diff
@@ -44,15 +44,19 @@ function KeyboardWrapper() {
...
   const isCapsLockActive = useHidStore(useShallow(state => state.keyboardLedState?.caps_lock));
 
-  // HID related states
-  const keyboardLedStateSyncAvailable = useHidStore(state => state.keyboardLedStateSyncAvailable);
-  const keyboardLedSync = useSettingsStore(state => state.keyboardLedSync);
-  const isKeyboardLedManagedByHost = useMemo(() =>
-    keyboardLedSync !== "browser" && keyboardLedStateSyncAvailable,
-    [keyboardLedSync, keyboardLedStateSyncAvailable],
-  );
+  /*
```
IDisposable: These are going to be used to show what keys are currently down for the VirtualKeyboard (still to be finished)

ui/src/components/VirtualKeyboard.tsx
```diff
@@ -168,19 +172,11 @@ function KeyboardWrapper() {
         toggleLayout();
 
         if (isCapsLockActive) {
-          if (!isKeyboardLedManagedByHost) {
```
IDisposable, line 171: Removed all the browser-side stuff about LED management

ui/src/components/WebRTCVideo.tsx
```diff
@@ -344,153 +335,45 @@ export default function WebRTCVideo() {
     sendAbsMouseMovement(0, 0, 0);
   }, [sendAbsMouseMovement]);
 
-  // Keyboard-related
```
IDisposable, line 347: This code may have to be restored and guarded to be used if the browser is talking to an old device... can that happen with the online (hosted) UI?

ui/src/components/WebRTCVideo.tsx
```diff
@@ -344,153 +335,45 @@ export default function WebRTCVideo() {
...
-        ...prev.activeModifiers,
-        modifiers[code],
-      ]);
-
       // When pressing the meta key + another key, the key will never trigger a keyup
       // event, so we need to clear the keys after a short delay
       // https://bugs.chromium.org/p/chromium/issues/detail?id=28089
       // https://bugzilla.mozilla.org/show_bug.cgi?id=1299553
-      if (e.metaKey) {
+      if (e.metaKey && hidKey < 0xE0) {
```
IDisposable, line 352: We only "fake the up" on the non-modifier key (not any of the modifiers)... so we get the key-up on the X for example when Meta-X is pressed.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -344,153 +335,45 @@ export default function WebRTCVideo() {
...
-      setIsNumLockActive,
-      setIsCapsLockActive,
-      setIsScrollLockActive,
-    ],
+    [sendKeypressEvent],
   );
 
   const keyUpHandler = useCallback(
-    (e: KeyboardEvent) => {
+    async (e: KeyboardEvent) => {
```
IDisposable, line 375: I wonder is all/more of our callbacks should be `async`?

ui/src/components/WebRTCVideo.tsx
```diff
@@ -667,6 +550,18 @@ export default function WebRTCVideo() {
         };
   }, [videoSaturation, videoBrightness, videoContrast]);
 
+  function getAdjustedKeyCode(e: KeyboardEvent) {
```
IDisposable, line 559: I'm not sure this is still needed, but if it _is_ then we have to do it for both key down and key up... so

ui/src/keyboardMappings.ts
```diff
@@ -14,7 +14,7 @@ export const keys = {
   CapsLock: 0x39,
   Comma: 0x36,
   Compose: 0x65,
-  ContextMenu: 0,
+  ContextMenu: 0x65,  // same as Compose
```
IDisposable: ooops.

ui/src/keyboardMappings.ts
```diff
@@ -42,6 +42,7 @@ export const keys = {
   F10: 0x43,
   F11: 0x44,
   F12: 0x45,
+  F13: 0x68,
```
IDisposable, line 66: missed it

ui/src/keyboardMappings.ts
```diff
@@ -120,6 +121,14 @@ export const keys = {
   Space: 0x2c,
   SystemRequest: 0x9a,
   Tab: 0x2b,
+  ControlLeft: 0xe0,
```
IDisposable: These all have to be in the keys collection because we need to get their names when we pass them around

ui/src/routes/devices.$id.settings.keyboard.tsx
```diff
@@ -33,11 +29,6 @@ export default function SettingsKeyboardRoute() {
   }, [keyboardLayout]);
 
   const layoutOptions = keyboardOptions();
-  const ledSyncOptions = [
```
IDisposable, line 36: Killed the  LED sync stuff.

wol.go
```diff
@@ -65,7 +65,7 @@ func createMagicPacket(mac net.HardwareAddr) []byte {
 	buf.Write(bytes.Repeat([]byte{0xFF}, 6))
 
 	// Write the target MAC address 16 times
-	for i := 0; i < 16; i++ {
+	for range 16 {
```
IDisposable, line 68: ... also a Modernizer update

IDisposable:
Once this is finalized (and we've answered the questions on support for old-school), then I will clean up VirtualKeyboard to track the down-keys and start on the revamp of layouts for that.

adamshiervani:
Awesome work -- will go through all of it.

Regarding the old devices in cloud mode, we have code for this, either as a hook `useFeatureFlag()` or as a component `<FeatureFlag />`.

ui/src/components/InfoBar.tsx
```diff
@@ -1,50 +1,65 @@
...
-  const hdmiState = useVideoStore(state => state.hdmiState);
+  const displayKeys = useMemo(() => {
+    if (!showPressedKeys || !keysDownState) 
+      return "";
+
+    const state = keysDownState as KeysDownState;
+    const activeModifierMask = state.modifier || 0;
+    const keysDown = state.keys || [];
+    const modifierNames = Object.entries(modifiers).filter(([_, mask]) => (activeModifierMask & mask) !== 0).map(([name, _]) => name);
+    const keyNames = Object.entries(keys).filter(([_, value]) => keysDown.includes(value)).map(([name, _]) => name);
```
adamshiervani, line 47: Filter the keys to remove "None"

adamshiervani:
Found an issue with Caps Lock. A few observations:

- Starting with Caps Lock being off, it takes three Caps Lock clicks, to Enable and then Disable Caps Lock.
- It doesn't seem to be a release event for it
- Weird logging loop until Caps Lock back to disabled.

```2025-08-14T11:13:00Z TRC jetkvm usb keysDownState updated new={"keys":[57,0,0,0,0,0],"modifier":0} old={"keys":[0,0,0,0,0,0],"modifier":0}
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb keyboardState updated new=5 old=7
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:00Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:01Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb got data from keyboard buf=[5,0,0,0,0,0,0,0] listener=keyboardEvents n=1 path=/dev/hidg0
2025-08-14T11:13:02Z TRC jetkvm usb reading from keyboard for LED state changes listener=keyboardEvents path=/dev/hidg0
2025-08-14T11:13:02Z INF jetkvm usb rpcKeypressReport called key=57 msg=HELLLOOOOHELLLOOOOHELLLOOOOHELLLOOOOHELLLOOOO press=false
2025-08-14T11:13:02Z TRC jetkvm usb keysDownState updated new={"keys":[0,0,0,0,0,0],"modifier":0} old={"keys":[57,0,0,0,0,0],"modifier":0}```

**Comment by IDisposable** [comment]
Interesting... so the "logging loop" is not a loop, this is the normal LED status being pushed from the host to the device, which then gets forwarded to the browser from the device. I think we could maybe not log that if it's unchanged to decrease the noise.

As for CapsLock... is the host device a MacOS or Windows machine? I ask because I've noticed that the _mac_ will track CapsLock state for the internal keyboard independently from the CapsLock on the device-simulated keyboard.  By this I mean that if you hit CapsLock on the mac's physical keyboard it will turn on that keyboard's LED, but NOT the remote/device keyboard (and toggles back and forth as expected). When you hit CapsLock on the remote/browser keyboard, it will send the key down to the device, which will forward to the host and then _trust_ the LED status that comes back (and of course forward to the browser).

As for the release event, when CapsLock is pressed, there should be a message with CapsLock keypress code  0x39 / 57 `keys` of `[57,0,0,0,0,0]` and a `press` of `true` which should get sent from device to the host and then it should note that the CapsLock is selected/toggled and then send the current LED State back with that bit set but I don't see that in the traces. Once you let go of the CapsLock key, then browser will send a keypress message with all the `key` `0x39` and a `press` of `false`, which will be relayed to the host with a HID write of zero modifier and all zeros for the keys-down... so it should see that the CapsLock key has been released (but of course, that doesn't change the toggle state until the next press/release pair).

I will do some more testing and see why we're not seeing the LED state coming back correctly.

I might want to unify the messages for keys down to _also_ include the current LED states as that seems likely better than polling.

If you're seeing some other behavior, perhaps we can do a screen share?

File: internal/usbgadget/hid_keyboard.go [hunk 27]
```diff
@@ -61,14 +61,18 @@ var keyboardReportDesc = []byte{
 
 const (
 	hidReadBufferSize = 8
+	hidKeyBufferSize  = 6
+	hidErrorRollOver  = 0x01
```
IDisposable, line 65: used when they hold down more than 6 keys

IDisposable:
> Regarding the old devices in cloud mode, we have code for this, either as a hook `useFeatureFlag()` or as a component `<FeatureFlag />`.

So I need to restore (logically) the [old handling code](https://github.com/jetkvm/kvm/pull/725/files#diff-503edceb49a3a14452eb4663ab40040a5ff04a2a89f13e48be02539e2c939f08L347-L493) and keep both things available and gate it based on if the remote accepts the keypressReport call [like we did for LED state management](https://github.com/jetkvm/kvm/blob/608f69db13b5862aa78f53057930bd7a82719767/ui/src/routes/devices.%24id.tsx#L669), or should we just have an option flag in settings?

IDisposable:
Restored the ability to run with either new devices or old devices (uses keypressReport when possible)

internal/usbgadget/hid_keyboard.go
```diff
@@ -211,22 +261,127 @@ func (u *UsbGadget) keyboardWriteHidFile(data []byte) error {
 	return nil
 }
 
-func (u *UsbGadget) KeyboardReport(modifier uint8, keys []uint8) error {
+func (u *UsbGadget) UpdateKeysDown(modifier byte, keys []byte) KeysDownState {
```
IDisposable, line 247: This takes care of updating the KeysDownState with some logic to reset the "keyboard overflow" after it was sent on to the host.

jsonrpc.go
```diff
@@ -1047,6 +1061,8 @@ var rpcHandlers = map[string]RPCHandler{
 	"renewDHCPLease":         {Func: rpcRenewDHCPLease},
 	"keyboardReport":         {Func: rpcKeyboardReport, Params: []string{"modifier", "keys"}},
```
IDisposable, line 1062: Now returns the KeysDownState as well, browser clients that aren't expecting that response will silently ignore it.

ui/src/components/InfoBar.tsx
```diff
@@ -1,50 +1,63 @@
...
+  const keyboardLedState = useHidStore((state: HidState) => state.keyboardLedState);
+  const isTurnServerInUse = useRTCStore((state: RTCState) => state.isTurnServerInUse);
 
-  const isTurnServerInUse = useRTCStore(state => state.isTurnServerInUse);
+  const usbState = useHidStore((state: HidState) => state.usbState);
+  const hdmiState = useVideoStore((state: VideoState)  => state.hdmiState);
 
-  const usbState = useHidStore(state => state.usbState);
-  const hdmiState = useVideoStore(state => state.hdmiState);
+  const displayKeys = useMemo(() => {
```
IDisposable, line 40: Builds a string with all the "down" modifiers and keys listed.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -28,15 +31,16 @@ import {
 export default function WebRTCVideo() {
   // Video and stream related refs and states
   const videoElm = useRef<HTMLVideoElement>(null);
-  const mediaStream = useRTCStore(state => state.mediaStream);
+  const mediaStream = useRTCStore((state: RTCState) => state.mediaStream);
   const [isPlaying, setIsPlaying] = useState(false);
-  const peerConnectionState = useRTCStore(state => state.peerConnectionState);
+  const peerConnectionState = useRTCStore((state: RTCState) => state.peerConnectionState);
   const [isPointerLockActive, setIsPointerLockActive] = useState(false);
+  const [isKeyboardLockActive, setIsKeyboardLockActive] = useState(false);
```
IDisposable, line 33: We now track if we actually have the keyboard locked.

ui/src/components/WebRTCVideo.tsx
IDisposable: I moved all this handling into _useKeyboard.ts_

ui/src/components/WebRTCVideo.tsx
```diff
@@ -344,153 +339,58 @@ export default function WebRTCVideo() {
...
+          console.debug(`Forcing the meta key release of associated key: ${hidKey}`);
+          handleKeyPress(hidKey, false);
         }, 10);
       }
-
-      sendKeyboardEvent([...new Set(newKeys)], [...new Set(newModifiers)]);
+      console.debug(`Key down: ${hidKey}`);
+      handleKeyPress(hidKey, true);
+      
+      if (!isKeyboardLockActive && hidKey === keys.MetaLeft) {
```
IDisposable, line 361: This logic is needed to release the `MetaLeft` key _if_ we were not able to set a keyboard lock (in fullscreen mode) because in that case, we will never see the meta key **up** event... so we'll simulate one to ensure by the time we get back to the browser the state is correct. Note that if they are still holding the key down when/if we get focus back then the normal keyboard repeat messages will ensure we actually  receive the meta is down again.

usb.go
```diff
@@ -31,21 +31,31 @@ func initUsbGadget() {
...
+	})
+
 	// open the keyboard hid file to listen for keyboard events
 	if err := gadget.OpenKeyboardHidFile(); err != nil {
 		usbLogger.Error().Err(err).Msg("failed to open keyboard hid file")
 	}
 }
 
-func rpcKeyboardReport(modifier uint8, keys []uint8) error {
+func rpcKeyboardReport(modifier byte, keys []byte) (usbgadget.KeysDownState, error) {
```
IDisposable, line 46: We now return the KeysDownState from the keyboardReport, which will be ignored by old browser clients.

IDisposable:
Knocking off for today... everything except the latching keys status seems to be working great. Especially against a mac in full-screen mode.

ui/src/hooks/useKeyboard.ts
```diff
@@ -52,5 +99,78 @@ export default function useKeyboard() {
...
+  
+      if (keyPressAvailable) {
+        // if the keyPress api is available, we can just send the key press event
+        sendKeypressEvent(key, press);
+        // if keyPress api is STILL available, we don't need to handle the key locally
+        if (keyPressAvailable) return;
+      }
+
+      // if the keyPress api is not available, we need to handle the key locally
+      const downState = handleKeyLocally(keysDownState, key, press);
```
adamshiervani: Could you make the `Locally` suffix sound more dramatic? The code makes total sense, and the method names say what they do, but I feel that in a few months, when all the context is forgotten, this code won't come off as a fix for legacy devices using the cloud but rather as an implementation detail.

What do you think?

ui/src/hooks/useKeyboard.ts
```diff
@@ -52,5 +99,78 @@ export default function useKeyboard() {
...
+
+  const handleKeyPress = useCallback(
+    (key: number, press: boolean) => {
+      if (rpcDataChannel?.readyState !== "open") return;
+  
+      if (keyPressAvailable) {
+        // if the keyPress api is available, we can just send the key press event
+        sendKeypressEvent(key, press);
+        // if keyPress api is STILL available, we don't need to handle the key locally
+        if (keyPressAvailable) return;
```
adamshiervani: I feel the same sentiment here. Like, it's very appropriate to bind the condition to the ability of the device rather than the version, which doesn't technically mean anything. I do, however, feel that it makes it clear what the context of the code is. 

What do you think? Am I over-indexing on the future forgetfulness?

ui/src/hooks/useKeyboard.ts
```diff
@@ -1,42 +1,89 @@
...
-  );
+  const rpcDataChannel = useRTCStore((state: RTCState) => state.rpcDataChannel);
+
+  const keysDownState = useHidStore((state: HidState) => state.keysDownState);
+  const setKeysDownState = useHidStore((state: HidState) => state.setKeysDownState);
+
+  const keyPressAvailable = useHidStore((state: HidState) => state.keyPressAvailable);
+  const setKeyPressAvailable = useHidStore((state: HidState) => state.setKeyPressAvailable);
 
   const sendKeyboardEvent = useCallback(
```
adamshiervani, line 28: Same thing here. Simulating a forgetful future and just looking at the code, why are there two keyboard methods, one `sendKeypressEvent` and one `sendKeyboardEvent`?

ui/src/components/VirtualKeyboard.tsx
```diff
@@ -125,82 +130,56 @@ function KeyboardWrapper() {
...
+      const dynamicKeys = ["ShiftLeft", "ShiftRight", "ControlLeft", "ControlRight", "AltLeft", "AltRight", "MetaLeft", "MetaRight"];
 
+      // handle the fake key-macros we have defined for common combinations
       if (key === "CtrlAltDelete") {
-        sendKeyboardEvent(
-          [keys["Delete"]],
-          [modifiers["ControlLeft"], modifiers["AltLeft"]],
-        );
-        setTimeout(resetKeyboardState, 100);
+        executeMacro([ { keys: ["Delete"], modifiers: ["ControlLeft", "AltLeft"], delay: 100 } ]);
```
adamshiervani: That's such a good idea!

ui/src/hooks/useKeyboard.ts
adamshiervani: Only in legacy/local mode, if I press and hold `a` it will repeat as expected, but if I press and hold `a` and then click `s` - `a` will repeat indefinitely.

ui/src/components/VirtualKeyboard.tsx
```diff
@@ -125,82 +130,56 @@ function KeyboardWrapper() {
...
-
-      // Update current keys and modifiers
-      sendKeyboardEvent(newKeys, newModifiers);
-
-      // If shift was used as a modifier and caps lock is not active, revert to default layout
-      if (keyHasShiftModifier && !isCapsLockActive) {
-        setLayoutName("default");
+      // if they press any of the dynamic keys, we send a keypress down event but we don't release it until they click it again
+      if (dynamicKeys.includes(key)) {
+        const currentlyDown = keysDownState.keys.includes(keys[key]);
```
adamshiervani: Never true, so all the dynamic keys are pressed forever.

adamshiervani (changes requested):

ui/src/hooks/useKeyboard.ts
IDisposable: I'm really sensitive to the whole idea of having to handle both APIs supported (and it largely works, I think)... I need to run once with my device unchanged and just local-mode UI and test for sure.

ALL that said, I will add a whole bunch of documentation in the code related to that and then add some developer.md type doc, okay?

ui/src/hooks/useKeyboard.ts
IDisposable: Yes, I certainly can get more verbose. Luckily, it's mostly all focused into useKeyboard.ts so we can certainly put a long-winded description and make this name more pedantically.

ui/src/components/InfoBar.tsx
IDisposable: Oh, good point... I hated how that looked, thanks for the idea.

ui/src/hooks/useKeyboard.ts
IDisposable: I'll document each of those for their usage.

ui/src/components/VirtualKeyboard.tsx
IDisposable: It evolved to this because I wanted to eliminate keyboardReport being exposed outside of useKeyboard.ts but so much happier with this state.

ui/src/components/VirtualKeyboard.tsx
IDisposable: I haven't finished the VirtualKeyboard changes yet (still have to have the on-screen keys reflect the state of the modifiers), but the idea here is that if they pressed any of the dynamic modifier keys, we need "toggle" up or down with each on-screen key-down press (e.g. first time the click the left shift key, we send a keyPress DOWN, and then when they click that on-screen key the next time, we send a keyPress UP)

IDisposable: I might not be indexing with the right and I'll likely put that "is it down" hackery in useKeyboard.ts so all the oddities are in one place... in any case, will address this comment when I finish up VirtualKeyboard next.

ui/src/hooks/useKeyboard.ts
IDisposable: I haven't seen that behavior! Are you controlling a PC or Mac, what OS?

ui/src/routes/devices.$id.tsx
```diff
@@ -127,22 +128,22 @@ export default function KvmIdRoute() {
...
-  const setMediaMediaStream = useRTCStore(state => state.setMediaStream);
-  const setPeerConnection = useRTCStore(state => state.setPeerConnection);
-  const setDiskChannel = useRTCStore(state => state.setDiskChannel);
-  const setRpcDataChannel = useRTCStore(state => state.setRpcDataChannel);
-  const setTransceiver = useRTCStore(state => state.setTransceiver);
-  const location = useLocation();
+  const { sidebarView, setSidebarView, disableVideoFocusTrap } = useUiStore();
+  const [ queryParams, setQueryParams ] = useSearchParams();
+
+  const {
```
IDisposable, line 134: Here an elsewhere, switching to the simpler syntax because that'll make zustand upgrade to 5.x easier.

ui/src/routes/devices.$id.tsx
```diff
@@ -211,7 +212,7 @@ export default function KvmIdRoute() {
           clearInterval(checkInterval);
           setLoadingMessage("Connection established");
         } else if (attempts >= 10) {
-          console.log(
+          console.warn(
```
IDisposable, line 215: Reduced some logging noise and emit warnings/errors when appropriate so the console log filter is useful

config.go
```diff
@@ -114,7 +114,7 @@ var defaultConfig = &Config{
 	ActiveExtension:      "",
 	KeyboardMacros:       []KeyboardMacro{},
 	DisplayRotation:      "270",
-	KeyboardLayout:       "en_US",
+	KeyboardLayout:       "en-US",
```
IDisposable, line 117: ... once and for all fixing this for old and new devices with configs set or not set... so it should always work now (I tested by a factory reset and found there were still issues)

IDisposable:
Looks like the last couple of things I need to tweak are the VirtualKeyboard showing what dynamic keys are down, and whatever the repeating/not-repeating thing is going on. Will work on that this weekend.

ui/src/hooks/useKeyboard.ts
IDisposable: I think I've fixed this... wasn't handling n-key rollover correctly in legacy mode.

ui/src/components/VirtualKeyboard.tsx
IDisposable: ... and I might need to go one further step because playing some macros in legacy mode seems wrong... I may just have macros, work by pressing and releasing keys as-if... but that's a problem for this weekend.

ui/src/hooks/useKeyboard.ts
IDisposable: Well documented now.

ui/src/components/InfoBar.tsx
IDisposable: Done.

IDisposable:
Moved the virtual keyboard layout, the modifier display map, and the key display map into each language (for now the non-US are repackage the US ones).

I centralized all the layout and layout name logic into useKeyboardLayout.ts so we don't have identical code scattered about.

I've got the VirtualKeyboard working now in "sticky" fashion.

Will be working on the dead-key/accent/ctrl/alt next, looking to have keys look like those on kbdlayout.info eventually like https://kbdlayout.info/kbdpl1 shows (look at the Z key)

image

One thing I would appreciate would be a better styling for the "down" keys. Currently it's this, but I don't think the background is applying :(

```css
.hg-theme-default .hg-row .down-key {
  background: rgb(28, 28, 28);
  @apply text-white! font-bold!;
}
```

ui/src/components/VirtualKeyboard.tsx
IDisposable: I've cleaned that code up and it's working right now :)

adamshiervani:
I will review all the new code, but let's try to get this PR out the door. Let's feature freeze this PR, because it's already massive and difficult to properly review.

IDisposable:
That makes sense to me, I will pull the Polish keyboard put as that's not complete.

Sorry for the scope, I had to get Virtual Keyboard working to make it shippable, Which it now is!

IDisposable:
Would appreciate some input on the "down keys" CSS mentioned above, it's kind janky

ui/src/components/VirtualKeyboard.tsx
```diff
@@ -123,94 +129,69 @@ function KeyboardWrapper() {
...
+        handleKeyPress(keys[key], !currentlyDown)
+        return;
       }
 
-      setTimeout(resetKeyboardState, 100);
+      // otherwise, just treat it as a down+up pair
+      const cleanKey = key.replace(/[()]/g, "");
+      console.debug(`Regular key pressed: ${cleanKey} sending down and up pair`);
+      handleKeyPress(keys[cleanKey], true);
+      setTimeout(() => handleKeyPress(keys[cleanKey], false), 50);
```
adamshiervani, line 188: Leave it like this for now, but we need a more stable way to do this.

ui/src/components/MacroForm.tsx
```diff
@@ -234,6 +236,7 @@ export function MacroForm({
                   }
                   onDelayChange={delay => handleDelayChange(stepIndex, delay)}
                   isLastStep={stepIndex === (macro.steps?.length || 0) - 1}
+                  keyboard={keyboard}
```
adamshiervani: Something is wrong with the Keys ComboBox - it keeps losing focus, making it almost impossible to enter any value

ui/src/components/VirtualKeyboard.tsx
```diff
@@ -1,4 +1,3 @@
-import { useShallow } from "zustand/react/shallow";
```
adamshiervani, line 1: When the `VirtualKeyboard` is enabled, Ctrl does not trigger pressed false, only true.

ui/src/hooks/useKeyboardLayout.ts
```diff
@@ -0,0 +1,22 @@
...
+import { useSettingsStore } from "@/hooks/stores";
+import { KeyboardLayout, selectedKeyboard } from "@/keyboardLayouts";
+
+export function useKeyboardLayout(): { keyboard: KeyboardLayout } {
+  const { keyboardLayout } = useSettingsStore();
+
+  const isoCode = useMemo(() => {
+    console.log("Current keyboard layout from store:", keyboardLayout);
+    if (keyboardLayout && keyboardLayout.length > 0)
+      return keyboardLayout.replace("en_US", "en-US");
```
adamshiervani, line 24: Could you add a comment in the code explaining why?

ui/src/components/VirtualKeyboard.tsx
IDisposable: I'm not sure what you're aiming for, this "playing back the key" as an up/down pair works reliably

We _could_ use the `onKeyReleased={onKeyUp}` hook and only force the release if the they hang on too long?

ui/src/components/MacroForm.tsx
IDisposable: Fixed it, the state management was causing rerenders.

ui/src/hooks/useKeyboardLayout.ts
IDisposable: Added

IDisposable:
I think this is completed.

ui/src/components/VirtualKeyboard.tsx
IDisposable: Missed this comment... on it.

IDisposable: I can no longer reproduce this error.

adamshiervani: I don't have time to figure out exactly why this happens, but removing [this prop](https://hodgef.com/simple-keyboard/documentation/options/physicalkeyboardhighlightpreventdefault/) from the `Keyboard` component, it starts behaving as expected.

ui/src/components/VirtualKeyboard.tsx
adamshiervani: In bad network, this can cause issues. It's not isolated to just this code, just a general note. That's however, a task for another day. <3

adamshiervani:
Amazing work!  

I tested both with the `dev` branch frontend and this branch’s frontend - both work as expected, except for the virtual keyboard issue commented on above. 

Once this is merged, we should take a look at these keyboard issues:  
- Fix Paste Modal inconsistency
- Add more language keyboards and update the docs with setup instructions
- Ensure stability in poor network conditions  
  - WebRTC guarantees ordered delivery on the data channel, and with [this](https://github.com/jetkvm/kvm/pull/744), Golang will as well.  
  - The main issue is that on weak networks, keypresses may arrive late, so “long-press” characters appear when the key-up event isn’t triggered in time.  
- UI improvements to the Virtual Keyboard  
  - Styling  
  - Language indicator  
  - Header spacing  
  - Detach mode  
  - Mobile layout  

I'm sure there's more, but this is what I've got top of mind now.

ui/src/components/VirtualKeyboard.tsx
IDisposable: Oh that's a simple change... I was trying to get the browser-side keypresses to show up via simple-keyboard but that never worked anyway. I'll yank all that stuff and push back since _it does_ show the key down because of the down-keys css class now (though I still want more/better for that).

Pushed!

IDisposable:
> Amazing work!

Thanks! It's been a bit of a slog, but well worth it... a lot of cleanup on both sides is always nice.

> I tested both with the `dev` branch frontend and this branch’s frontend - both work as expected, except for the virtual keyboard issue commented on above.

I've pushed that fix, thanks for tracking it down. Please verify, since I never saw that issue in my tests (maybe because the JetKVM device and my client/browser is on the same switch ;) ) 

> Once this is merged, we should take a look at these keyboard issues:
> 
> * Fix Paste Modal inconsistency

Tell me more!

> * Add more language keyboards and update the docs with setup instructions

Excellent idea. I can do a trivial PR after we merge (would be nice for a simple one this time!)

> * Ensure stability in poor network conditions
>   * WebRTC guarantees ordered delivery on the data channel, and with [this](https://github.com/jetkvm/kvm/pull/744), Golang will as well.

That's a really cool find by @alexballas! I suggest you merge his right now (before this one) and I'll ensure there's no conflicts real fast.

>   * The main issue is that on weak networks, keypresses may arrive late, so “long-press” characters appear when the key-up event isn’t triggered in time.

This is so much better post this PR... I suspect with #744 and an equivalent change for the device-to-client originated messages (like LED State) we would be tons better. Also, I saw somewhere in [this comment](https://github.com/jetkvm/kvm/pull/718#issuecomment-3172605441) of PR #718 the idea that the general-user RPC channel should be refactored to split out the distinct messages for the audio... we would likely benefit from a distinct channel for HID messages (for mouse and keyboard).

> * UI improvements to the Virtual Keyboard
>   
>   * Styling

Yes, please! Not my skill set at all...

>   * Language indicator

Simple and smart... should be similar to the PasteModal so pick a form and I'll hack it.

>   * Header spacing
>   * Detach mode
>   * Mobile layout

> I'm sure there's more, but this is what I've got top of mind now.

#735, #694 are also mentioned... I just went through some of the open issues, and I REALLY didn't realize how many of these issues get fixed by this PR (and/or were in PR #500) sorry it dragged on so long :(

ui/src/components/VirtualKeyboard.tsx
IDisposable: It think #744 will largely resolve that issue.

IDisposable:
Rebased and conflicts resolved

ui/src/components/WebRTCVideo.tsx
IDisposable: Update: No, actually we already have too much `async` as we're decorating some methods that don't even have awaits. Stripped those out in #746

adamshiervani (approved):
Hell yeah, amazing work!!! Let's get this merged now.

IDisposable:
Whew...

Caedis:
Trying this out locally and seems like LMeta is not able to be held down. When I press and hold it down, the status bar only shows it pressed for a split second

IDisposable:
> Trying this out locally and seems like LMeta is not able to be held down. When I press and hold it down, the status bar only shows it pressed for a split second

LMeta is pretty hard to manage because it pops up the OS menus when the browser/remote is on Windows and MacOS... mostly the only way I can get things to work with LMeta+_something_ is to use Macros or Virtual Keyboard, but I would like to know more about how it acts on your system. Can you record a video or screen-share?

Caedis:
Before this pr, i was able to easily use Windows+Shift+S, but now I have to basically press all 3 at once and release quickly

Jetkvm connected to Windows  
Browser is on Linux

IDisposable:
> Before this pr, i was able to easily use Windows+Shift+S, but now I have to basically press all 3 at once and release quickly
> 
> Jetkvm connected to Windows Browser is on Linux

Trying to get to the bottom of this! Please bear with me...

Do you mean pressing keys on the physical keyboard?

What OS is the browser/client running on? I think you mean you've got

```mermaid
flowchart LR
    PC[Linux] -->|press physical key | LMeta
    LMeta --> Browser[fa:fa-browser Chrome]
    Browser -->|event| JetKVM[client]
    JetKVM -->|RPC| Device
    Device -->|HID| Host[Windows]
```

Also, are you keyboard trapped, fullscreen + running the JetKVM on an HTTPS (not HTTP)?

IDisposable:
> Before this pr, i was able to easily use Windows+Shift+S, but now I have to basically press all 3 at once and release quickly

Oh, do you mean that you before could press-and-release LMeta, press-and-release Shift, press S and it would be seen as a Windows+Shift+S on the host?

If that's the case I suspect that's because the old keyboard stack wasn't processing things in order or synchronously (which broke all kinds of input things and basically had to be fixed to be in-order processing.

Caedis:
No, i could press shift, then meta, then s, Windows would register it, then I would release all 3.

Yes, pressing and holding the physical key would only show as a quick press and release of the LMeta key.

I have 0 issues holding down LMeta in dwm or hyprland to move stuff around or executing other shortcuts that use LMeta.

Caedis:
Non-fullscreen and absolute mouse so not keyboard trapped. Wish that worked but it doesnt even with an https reverse proxy with an LE cert.

Caedis:
Client = Arch in Firefox
Host/Jetkvm is attached to Windows.

I have 0 shortcuts set to LMeta+Shift+S

Caedis:
Preview: 

https://github.com/user-attachments/assets/fbe35160-e3d8-4c1c-83fb-da4effb6fb56

IDisposable:
Great repro, thanks...will look into that ASAP

alexballas:
One thing I noticed:
Over https://app.jetkvm.com/ Holding LeftShift down and then pressing MetaLeft doesn't even trigger this momentary "MetaLeft" press. Nothing appears.

Over IP (without HTTPS) I do see this momentary press

alexballas:
```
      if (!isKeyboardLockActive && hidKey === keys.MetaLeft) {
        // If the left meta key was just pressed and we're not keyboard locked
        // we'll never see the keyup event because the browser is going to lose
        // focus so set a deferred keyup after a short delay
        setTimeout(() => {
          console.debug(`Forcing the left meta key release`);
          handleKeyPress(hidKey, false);
        }, 100);
      }
```
this part here is the one causing the momentary MetaLeft press

Caedis:
Nice find

alexballas:
On Linux when I press Alt key and change the focus of the window without releasing it, the browser will force release it. So I dont know if the following statement is correct.
```
        // If the left meta key was just pressed and we're not keyboard locked
        // we'll never see the keyup event because the browser is going to lose
        // focus so set a deferred keyup after a short delay
```

One for @IDisposable to confirm

Caedis:
Commented out that 2 blocks and deployed and everything works for me
(meta and shift being on at the start was due to my shortcut to record a selection being Meta + Shift + PrintSc)

https://github.com/user-attachments/assets/2bbd93c6-3381-4e4d-a06b-ed615b1a022b

internal/usbgadget/hid_keyboard.go
```diff
@@ -123,7 +128,35 @@ func (u *UsbGadget) GetKeyboardState() KeyboardState {
...
+	defer u.keyboardStateLock.Unlock()
+
+	if u.keysDownState.Modifier == state.Modifier &&
+		bytes.Equal(u.keysDownState.Keys, state.Keys) {
+		return // No change in key down state
+	}
+
+	u.log.Trace().Interface("old", u.keysDownState).Interface("new", state).Msg("keysDownState updated")
+	u.keysDownState = state
+
```
IDisposable, line 152: On review post folks beginning to test, I think I should have released the Lock here to ensure we're not holding through the entire callback for onKeysDownChange. Thoughts @adamshiervani ?

IDisposable:
> Over https://app.jetkvm.com/ Holding LeftShift down and then pressing MetaLeft doesn't even trigger this momentary "MetaLeft" press. Nothing appears.

As far as I know, the app.jetkvm.com has _not_ been updated yet since this isn't fully released. Thus it is still using the ReportKeys (old stack/Macros) instead of the ReportKeyPress (new stack)

> ```
>       if (!isKeyboardLockActive && hidKey === keys.MetaLeft) {
>         // If the left meta key was just pressed and we're not keyboard locked
>         // we'll never see the keyup event because the browser is going to lose
>         // focus so set a deferred keyup after a short delay
>         setTimeout(() => {
>           console.debug(`Forcing the left meta key release`);
>           handleKeyPress(hidKey, false);
>         }, 100);
>       }
> ```
> 
> this part here is the one causing the momentary MetaLeft press

Correct, not sure how to handle it now, but I certainly can remove that code in favor of a lazy timer. The comment remains true, though, because if someone is _not_ in fullscreen on an HTTPS connection, we can't grab a keyboard lock, and thus the LMeta kicks in for the browser's OS (and thus we would never see the key up event. This code (in a different form existed before) see https://github.com/jetkvm/kvm/pull/725/files#diff-503edceb49a3a14452eb4663ab40040a5ff04a2a89f13e48be02539e2c939f08L441-L450

I will try to find a better way that doesn't force the key-up event unless we've lost focus and some interval passes... but that's going to be fun because it's possible for the key state to get out of sync with the device and host... so have to protect that.

IDisposable:
> On Linux when I press Alt key and change the focus of the window without releasing it, the browser will force release it. So I dont know if the following statement is correct.
> 
> ```
>         // If the left meta key was just pressed and we're not keyboard locked
>         // we'll never see the keyup event because the browser is going to lose
>         // focus so set a deferred keyup after a short delay
> ```
> 
> One for @IDisposable to confirm

Interesting... maybe I should just not special-case the LMeta at all...

IDisposable:
> Over https://app.jetkvm.com/ Holding LeftShift down and then pressing MetaLeft doesn't even trigger this momentary "MetaLeft" press. Nothing appears.

As far as I know, the code on app.jetkvm.com is the old stack still (since 0.5 hasn't actually released). 

To get the new stack for now, you will have to be running against the device directly (via IP or your local DNS name)

In fact I had to rework some stuff in this PR to support the opposite case (when app has been upgraded but the device hasn't)... wheeee

IDisposable:
> Commented out that 2 blocks and deployed and everything works for me (meta and shift being on at the start was due to my shortcut to record a selection being Meta + Shift + PrintSc)

Awesome, thanks for confirming... I'm all in favor of just not special casing things :) 

What's the "other" block you deleted?

Caedis:
image

PR #739, opened 2025-08-17
Update mount list for new Debian 13 release

### LINKED ISSUE #738: Debian 13 has been released. Need update Image URL Mount list
### Disclaimer

### Application version

0.4.6

### System version

0.2.5

### Device model

JetKVM

### Extension model

None

### Remote device Hardware

Lenovo M720X

### Remote device OS

Debian 13

### Bug description

One week ago, there was released Debian 13.
So, need update Image URL Mount list.

Image

KayTeeFive:
Closes #738 (https://github.com/jetkvm/kvm/issues/738)

### DIFF
--- a/ui/src/routes/devices.$id.mount.tsx
+++ b/ui/src/routes/devices.$id.mount.tsx
@@ -531,6 +531,11 @@ function UrlView({
       icon: DebianIcon,
     },
     {
+      name: "Debian 12 Bookworm (old-stable)",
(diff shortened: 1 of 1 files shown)


KayTeeFive:
I'm prefer to keep Debian 12 also, because it's well know version, that will be supported at least 3 years.
And in reality, a lot of private environments are compatible with that well known version and switching to latest stable release will take a time.

PR #742, opened 2025-08-19
Improve Custom jiggler settings and add timezone support

### LINKED ISSUE #741: Findings testing jiggler in dev branch
### Disclaimer

### Application version

0.4.6_dev-9f573200b1f8a58e4347e5b5ab357fd5aaad91bb

### System version

0.2.5

### Device model

JetKVM

### Extension model

None

### Remote device Hardware

_No response_

### Remote device OS

_No response_

### Bug description

I was interested testing #316 PR when I saw it had merged as the latest release version is moving the pointer to the corner quite often  even when actively used, which 316 addresses.

Here's some things that I would like to address as potential issues for the next release version.

* "Business hours" profile will likely cause issues logged on GitHub for "not working" simply due to timezones as the device operates in UTC, which most probably won't even think about.
* Secondly the "Business hours" profile is set to 9-17, at least here it's usually 8-17 for a workday.
* Setting the jiggler to "custom" and entering custom values -> Save says that it saves. But reloading the page and going back to settings it's as if the settings been reset for the jiggler. This is likely it defaulting to the values of the "custom" profile rather than the saved values for it

The changes in 316 are sought after, but I do think it will generate some noise.
My suggestion are:
* Fix so that custom reads back the saved values, rather than the default of custom
* Remove the "Business hours" profile
* Give more cron examples instead of 
* Clarify that time for cron is in UTC timezone

adamshiervani:
- Add timezone field to JigglerConfig with comprehensive IANA timezone list
- Fix custom settings not loading current values
- Remove business hours preset, add as examples in custom settings
- Improve error handling for invalid cron expressions


ym (approved):

Jonher937 (approved):
I've tested this PR and from what I can tell it addreses all of the points I raised in the issue.
LGTM!

Issue #325, opened 2025-04-03
Random repeated keystrokes & sluggish mouse movement


thierryrama:
I just received my jetKvm unit today and while playing with it, I experienced random mouse move slugginess and repeated keys. 

1. Like while typing, I would get a long streaks of iiiiiiii or carriage return or any key that happened to be typed at a point it decided to repeat it.
2. Moving the mouse suddenly becomes jittery and would delayed.

The issue resolved by itself but can last for 30 second to 1 minute for the mouse and I cannot tell for the keyboard since once you notice the repeated characters then I stopped typing.

Has anybody else experienced a similar issue. I am using a Mac with Safari 16.4 to control a Windows 11 laptop.

quadcom:
Here too.

Running latest releases A0.3.8 and S0.2.3.

Working with Linux CLI (Unraid, HOAS) becomes difficult, especially if trying to type in a log in password.

hushang999:
I'm having this exact same issue, super frustrating.

STR4NG3RdotSH:
Here too. I'm suspecting this is because I'm accessing over wifi. Will know for sure once I switch my client machine over to ethernet.

DrWheelicus:
I have been having the same issue with repeating keystrokes, but I have had the issue from the start (January). Even Ethernet doesn't solve it

STR4NG3RdotSH:
Issue hasn't returned for me. Its been 2 days or so. Ethernet was the fix, so the issue is very likely brought on by network quality issues. If its happening on ethernet, I'd first look at your network quality. Just my opinion, since we're both likely running the same codebase (I have latest update available thru web gui), meaning the only differing factor between you and me (since you are on ethernet too) is our network gear, varied other network devices on the network, etc.

stylophile:
My devices have been hardwired Ethernet since day 1, and I’ve having the issue since day 1. Windows 11
Client in Edge connecting to a MacOS host.

DrWheelicus:
> Issue hasn't returned for me. Its been 2 days or so. Ethernet was the fix, so the issue is very likely brought on by network quality issues. If its happening on ethernet, I'd first look at your network quality. Just my opinion, since we're both likely running the same codebase (I have latest update available thru web gui), meaning the only differing factor between you and me (since you are on ethernet too) is our network gear, varied other network devices on the network, etc.
> 
[...]

PR #744, opened 2025-08-21
Fix for issue 325. Ensure that rpc messages get processed sequentially

alexballas:
Ensure that RPC messages get processed sequentially and avoid phantom and repeated key presses

### DIFF
--- a/webrtc.go
+++ b/webrtc.go
@@ -23,6 +23,7 @@ type Session struct {
 	HidChannel               *webrtc.DataChannel
 	DiskChannel              *webrtc.DataChannel
 	shouldUmountVirtualMedia bool
+	rpcQueue                 chan webrtc.DataChannelMessage
 }
 
 type SessionConfig struct {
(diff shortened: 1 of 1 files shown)


IDisposable (approved):
:shipit: Fixes #325

IDisposable:
Do we want to setup the same cool logic for the back-channel (the push of keyboard LED state) from the device to the client/browser? These messages would originate here https://github.com/jetkvm/kvm/blob/dev/usb.go#L30

alexballas:
> Do we want to setup the same cool logic for the back-channel (the push of keyboard LED state) from the device to the client/browser? These messages would originate here https://github.com/jetkvm/kvm/blob/dev/usb.go#L30

The logic that reads the status messages from the keyboard seems to be fairly synchronous as per https://github.com/jetkvm/kvm/blob/dev/internal/usbgadget/hid_keyboard.go#L167 
It reads the keyboards events in a loop.

PR #745, opened 2025-08-22
Actually fix the useJsonRpc "any" problem

IDisposable:
PR #743 didn't have all the files included in the commit.
Mea culpa, many apologies.


IDisposable:
If we want just revert #743 for now... it's just failing the UI Lint

PR #748, opened 2025-08-25
Enhance connection stats sidebar

adamshiervani:
This PR refactors the Connection Stats sidebar to use a new reusable Metric component and adds clearer diagnostics for video performance.

**Changes**
- New Metric component for consistent chart rendering
- Grouped metrics into Connection and Video sections
- Added two key metrics:
- Network Stability (RTP Jitter) → how steady inbound packets arrive
- Playback Delay (Jitter Buffer Avg. Delay) → extra wait added by the browser before showing frames

These help distinguish between network jitter (transport) and playback delay (buffering), making video lag issues measurable.

**Screenshot**
image


ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,167 @@
...
+export function createChartArray<T, K extends keyof T>(
+  stream: Map<number, T>,
+  metric: K,
+): { date: number; stat: T[K] | null }[] {
+  const stat = Array.from(stream).map(([key, stats]) => {
+    return { date: key, stat: stats[metric] };
+  });
+
+  // Sort the dates to ensure they are in chronological order
+  const sortedStat = stat.map(x => x.date).sort((a, b) => a - b);
```
IDisposable: This seems expensive to do... is there any way we can just make it a requirement that the stat stream is already sorted (or more importantly constructed in the right order as the stats are accumulated)

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,167 @@
...
+  // Determine the starting point for the chart data
+  const firstChartDate = earliestStat ? Math.min(earliestStat, now - 120) : now - 120;
+
+  // Generate the chart array for the range between 'firstChartDate' and 'now'
+  return Array.from({ length: now - firstChartDate }, (_, i) => {
+    const currentDate = firstChartDate + i;
+    return {
+      date: currentDate,
+      // Find the statistic for 'currentDate', or use the last known statistic if none exists for that date
+      stat: stat.find(x => x.date === currentDate)?.stat ?? null,
```
IDisposable: This also seems extra costly, especially because we know that the sortedStat is in ascending order, we can just skip forward from a (initialized-to-firstChartDate) cursor... in fact, this is a classic merge/left-outer-join sort where the left is the dates and the right is the sortedStats... and the `stat` in the output is undefined if we don't have stats for that date...right?

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,167 @@
...
+  heightClassName = "h-[127px]",
+  referenceValue,
+  badge,
+  badgeTheme,
+}: MetricProps<T, K>) {
+  const ready = gate ? gate.size > 0 : stream ? stream.size > 0 : true;
+  const supportedFinal =
+    supported ??
+    (stream && metric
+      ? Array.from(stream).some(([, s]) => s[metric] !== undefined)
```
IDisposable: We can just iterate the `stream` directly (and avoid converting to an Array) for _some_ behaviour with this

```typescript
function someIterable<T>(iterable: Iterable<T>, predicate: (item: T) => boolean): boolean {
  for (const item of iterable) {
    if (predicate(item)) {
      return true; // Found an item that satisfies the predicate
    }
  }
  return false; // No item satisfied the predicate
}
```

used here like: 
```typescript
stream.someIterable(([, s]) => s[metric] !== undefined)
```

IDisposable (commented):
Cool refactoring!

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+/* eslint-disable-next-line */
+export function createChartArray<T, K extends keyof T>(
+  metrics: Map<number, T>,
+  metricName: K,
+) {
+  const result: { date: number; metric: number | null }[] = [];
+  const iter = metrics.entries();
+  let next = iter.next() as IteratorResult<[number, T]>;
+  const now = Math.floor(Date.now() / 1000);
```
IDisposable: maybe call this `nowSeconds` so it's clearer why the funging.

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,170 @@
...
+  unit: string;
+  heightClassName?: string;
+  referenceValue?: number;
+  badge?: ComponentProps<typeof MetricHeader>["badge"];
+  badgeTheme?: ComponentProps<typeof MetricHeader>["badgeTheme"];
+}
+
+/* eslint-disable-next-line */
+export function createChartArray<T, K extends keyof T>(
+  metrics: Map<number, T>,
```
IDisposable, line 39: Can we document  the expectation that the `metrics` are supplied in-order from oldest to newest?

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,170 @@
...
+  // We want 120 data points, in the chart.
+  const firstDate = Math.min(next.value?.[0] ?? now, now - 120);
+
+  for (let t = firstDate; t < now; t++) {
+    while (!next.done && next.value[0] < t) next = iter.next();
+    const has = !next.done && next.value[0] === t;
+
+    let metric = null;
+    if (has) metric = next.value[1][metricName] as number;
+    result.push({ date: t, metric });
```
IDisposable, line 56: I think I understand this, but might want some more comments.

1. First line is zooming the iterator from current, skipping points until we reach one that is at or after the current date (which is actually a _second_, right?)
2. Second line sets the flag saying if we're not done and the current point is EXACTLY at the specified second.
3. Then we yield up either `null` or (if at exact current second and we still have more points) yield up the next data point.
4. The we push that next point or a `null`

So this seems a bit off to me, I think you wanted to zoom forward to the "current second" in `t`... skipping points that are too early. I think that point should be pushed, why are we discarding it? 

- Was the intent to check if we've already pushed a point at `t`, which shouldn't be possible since we just incremented `t`
- Was the intent to skip the first point in the second and use the next one? If so, what if that next point is NOT in the next second, but further down?

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,170 @@
...
+  domain = [0, 600],
+  unit = "",
+  heightClassName = "h-[127px]",
+  badge,
+  badgeTheme,
+}: MetricProps<T, K>) {
+  const ready = gate ? gate.size > 0 : stream ? stream.size > 0 : true;
+  const supportedFinal =
+    supported ??
+    (stream && metric ? someIterable(stream, ([, s]) => s[metric] !== undefined) : true);
```
IDisposable, line 136: Unlikely to matter, but wanted to mention that this _has data_ test doesn't test the projection of the `map` function, so it's "possible" that it could project every metric value to `undefined` (not likely both because we might not expect the mapping to result in an `undefined` and also that every point projects there)... but just a note.

I do not think we should map them here, but at least make sure we know it's not going to happen normally.

ui/src/components/Metric.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  // Either we let the consumer provide their own chartArray, or we create one from the stream and metric.
+  const raw = data ?? ((stream && metric && createChartArray(stream, metric)) || []);
+
+  // If the consumer provides a map function, we apply it to the raw data.
+  const dataFinal: ChartPoint[] = map ? raw.map(map) : raw;
+  const recent = dataFinal
+    .slice(-(raw.length - 1))
```
IDisposable: Assuming the intent is to find the next "recent" point but skip  the first one... thoughts

- Should this be `dataFinal.length` not `raw.length`? I assume the map method can strip/skip points as well as translate/convert?
- `slice(1)` would do the same thing _IFF_ we have at least 2 points (if we have exactly 1 point, current code will return an array with only dataFinal[0], while `slice(1)` would return an empty array. Not sure what we want `recent` to be here
- Was the intent to zoom backwards from the end instead?
- Should the `filter` also discard `undefined`?

ui/src/components/sidebar/connectionStats.tsx
```diff
@@ -66,15 +29,13 @@ export default function ConnectionStatsSidebar() {
   useInterval(function collectWebRTCStats() {
     (async () => {
       if (!mediaStream) return;
-      const videoTrack = mediaStream.getVideoTracks()[0];
```
IDisposable, line 63: Is this guard being removed intentional? Seems odd that we shouldn't be ensuring we have a `videoTrack` like before?

ui/src/components/sidebar/connectionStats.tsx
```diff
@@ -98,144 +59,133 @@ export default function ConnectionStatsSidebar() {
...
+  const jitterBufferEmittedCount = createChartArray(
+    inboundVideoRtpStats,
+    "jitterBufferEmittedCount",
+  );
+
+  const jitterBufferAvgDelayData = jitterBufferDelay.map((d, idx) => {
+    if (idx === 0) return { date: d.date, metric: null };
+    const prevDelay = jitterBufferDelay[idx - 1]?.metric as number | null | undefined;
+    const currDelay = d.metric as number | null | undefined;
+    const prevEmitted =
```
IDisposable: `prevCountEmitted` and `currCountEmitted` might be nicer names?

ui/src/components/sidebar/connectionStats.tsx
```diff
@@ -98,144 +59,133 @@ export default function ConnectionStatsSidebar() {
...
+                  unit=" packets"
+                />
+
+                {/* Frames Per Second */}
+                <Metric
+                  title="Frames per second"
+                  description="Number of inbound video frames displayed per second."
+                  stream={inboundVideoRtpStats}
+                  metric="framesPerSecond"
+                  domain={[0, 80]}
```
IDisposable, line 184: This is unchanged, but curious why 0-80 instead of the more typical 0-100 or 0-60 in the other graphs?

IDisposable (commented):
Looks really solid, just a few clarifications for my brain to ensure I reviewed it for what it's intended to do :)

ui/src/components/Metric.tsx
IDisposable: Oh, dear... I think I read this entirely wrong. We're trying to skip to the next _second_ and grab the first point that has a metric value AND is in that second. If we don't get a non-null value in that second (e.g. before the next one), then we don't push a point at all and if we zoom past the current second on input points, we want to "hold them up" for the outer loop to catch up.  If that's what we want, then this is a merge left join of "seconds" against points in that second...

Should we skip subsequent points within the second (e.g. same second has _n_ points) or accumulate them in some why (average/sum/etc)?
Should we insert a row with no metric for those missing a match (e.g. a left OUTER join)?

This sort of generator-based processing function is something I use a bit (mine is slightly more generic-named)... nice because it doesn't allocate any arrays, just the state needed for the generator/yield behavior (which is tiny and well-optimized by the JavaScript runtimes) and has pluggable accumulator options.

```typescript
type Metric = [number, number]; // [second, value]
type ResultPoint = [number, number | null]; // [second, value | null]
type Accumulator = (current: number | undefined, next: number) => number;

// Generator function to process metrics, yielding points for specified seconds, requires both metrics and timeWindow to be preordered ascending 
function* processMetrics(
  metrics: Iterable<Metric>,
  timeWindow: Iterable<number>,
  outer: boolean = false,
  accumulate: Accumulator = (current, next) => (current ?? 0) + next,
  divideAccumulationBy: number | undefined = undefined
): Generator<ResultPoint> {
  // Get iterator for metrics and time window
  const metricIterator = metrics[Symbol.iterator]();
  const timeIterator = timeWindow[Symbol.iterator]();
  let currentMetric = metricIterator.next();
  let currentTime = timeIterator.next();

  // Process each second in the time window
  while (!currentTime.done) {
    const second = currentTime.value;

    // Skip metrics before the current second
    while (!currentMetric.done && currentMetric.value[0] < second) {
      currentMetric = metricIterator.next();
    }

    // Accumulate values for the current second
    let accumulated: number | undefined;
    let count = 0;
    while (!currentMetric.done && currentMetric.value[0] === second) {
      accumulated = accumulate(accumulated, currentMetric.value[1]); // possibly skip undefined and null
      count++;
      currentMetric = metricIterator.next();
    }

    // Yield result: value if present, or null if outer is true
    if (accumulated !== undefined && count > 0) {
      // Compute final value with intentional division
      let finalValue = accumulated;
      if (divideAccumulationBy === undefined) {
        finalValue = accumulated / count;
      } else if (divideAccumulationBy !== undefined) {
        finalValue = accumulated / divideAccumulationBy;
      }
      yield [second, finalValue];
    } else if (outer) {
      yield [second, null];
    }

    currentTime = timeIterator.next();
  }
}

/*
 * Example Accumulator Functions:
 *
 * Note: For accumulators that require division by count (e.g., average), use the sum accumulator
 * with divideAccumulationBy = undefined. For others, use divideAccumulationBy = 1 (no effective division).
 *
 * 1. Running Average (use sum accumulator with divideAccumulationBy = undefined):
 *    Averages values by summing and dividing by count.
 *    Accumulator: (current, next) => (current ?? 0) + next
 *    Example: processMetrics(..., (current, next) => (current ?? 0) + next, undefined, outer)
 *    For [20, 30]: (20 + 30) / 2 = 25
 *
 * 2. Minimum (divideAccumulationBy = 1):
 *    Takes the smallest value.
 *    const minValue: Accumulator = (current, next) => Math.min(current ?? next, next);
 *
 * 3. Maximum (divideAccumulationBy = 1):
 *    Takes the largest value.
 *    const maxValue: Accumulator = (current, next) => Math.max(current ?? next, next);
 *
 * 4. First (divideAccumulationBy = 1):
 *    Takes the first value.
 *    const firstValue: Accumulator = (current, next) => current ?? next;
 *
 * 5. Last (divideAccumulationBy = 1):
 *    Takes the last value.
 *    const lastValue: Accumulator = (current, next) => next;
 *
 * 6. Cumulative Sum (divideAccumulationBy = 1):
 *    Sums all values without dividing.
 *    const cumulativeSum: Accumulator = (current, next) => (current ?? 0) + next;
 *
 * 7. Exponential Smoothing (divideAccumulationBy = 1):
 *    Applies exponential moving average with smoothing factor alpha.
 *    const exponentialSmoothing: Accumulator = (current, next) => {
 *      const alpha = 0.5;
 *      return current === undefined ? next : alpha * next + (1 - alpha) * current;
 *    };
 */
```

ui/src/components/sidebar/connectionStats.tsx
adamshiervani: I blame the weather. 🌧️  Re-added

ui/src/components/Metric.tsx
adamshiervani: Stupid implementation. Changed it all completely.

ui/src/components/sidebar/connectionStats.tsx
adamshiervani: We're mostly going to be around 60 fps, and just a bit of headroom looked the best.

IDisposable (approved):
:shipit:

PR #750, opened 2025-08-27
Clean up Virtual Keyboard styling

IDisposable:
Make the Virtual Keyboard header a bit taller
Add keyboard layout name to header
Make the detached keyboard renders smaller key text so you can read them.
Updated the settings text for keyboard layout.


adamshiervani:
Closing in favor of https://github.com/jetkvm/kvm/pull/761

PR #732, opened 2025-08-13
feat: static ip & dhcp

ym:

PR #751, opened 2025-08-27
Allow disabling IPv6

IDisposable:
Simply ignores any IPv6 addresses in the lease and doesn't offer them to the RPC
Fixes https://github.com/orgs/jetkvm/projects/7/views/1?pane=issue&itemId=122761546&issue=jetkvm%7Ckvm%7C685

### DIFF
--- a/internal/network/netif.go
+++ b/internal/network/netif.go
@@ -239,6 +239,10 @@ func (s *NetworkInterfaceState) update() (DhcpTargetState, error) {
 			ipv4Addresses = append(ipv4Addresses, addr.IP)
 			ipv4AddressesString = append(ipv4AddressesString, addr.IPNet.String())
 		} else if addr.IP.To16() != nil {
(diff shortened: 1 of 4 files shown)


IDisposable:
Rebased and resolved conflicts

adamshiervani:
Doesn't work for me; I can still access the UI over IPv6 after disabling it in the settings.

Let's just wait until https://github.com/jetkvm/kvm/pull/732 gets merged, and then add this. A lot will change in #732 in regard to networking, and it will just be easier to base it off that work.

IDisposable:
Yep, would be best. I wasn't really sure of the scope of this change and saw the other one in flight.

IDisposable:
@adamshiervani I cleaned this code up a lot, perhaps it's good enough to ship now?

IDisposable:
(I can't reopen, and the PR is not updating :( ) started a new one at #803

PR #752, opened 2025-08-27
chore/Deprecate browser mount

IDisposable:
No longer supported.


IDisposable:
This may be too little or too much, the connection betwixt _browser_ and _webrtc_ is inferred, so I might have gotten too much.

adamshiervani:
Could you remove the Golang parts too?

IDisposable:
Oh, yeah, that would be nice. Done.

ui/src/components/popovers/MountPopover.tsx
```diff
@@ -14,38 +11,17 @@ import { useLocation } from "react-router-dom";
...
 import { formatters } from "@/utils";
-import { RemoteVirtualMediaState, useMountMediaStore, useRTCStore } from "@/hooks/stores";
+import { RemoteVirtualMediaState, useMountMediaStore } from "@/hooks/stores";
 import { SettingsPageHeader } from "@components/SettingsPageheader";
 import { JsonRpcResponse, useJsonRpc } from "@/hooks/useJsonRpc";
 import { useDeviceUiNavigation } from "@/hooks/useAppNavigation";
 import notifications from "@/notifications";
 
 const MountPopopover = forwardRef<HTMLDivElement, object>((_props, ref) => {
-  const { diskDataChannelStats } = useRTCStore();
```
adamshiervani, line 24: I don't think this is used anywhere else. Remove from the store too.

adamshiervani (changes requested):
Tested, and all remaining mount options work as expected! Just remove that disk channel stats ref from the store, and I'll merge.

ui/src/components/popovers/MountPopover.tsx
IDisposable: Removed diskChannel and localFile too...

DemiMarie:
I saw this PR and I’m curious why this change was made, especially given that the browser mount is mentioned at <https://jetkvm.com/docs/peripheral-devices/mount-drive>.

PR #755, opened 2025-08-29
feat: hid rpc channel

ym:
This PR moved HID events to a dedicated queue and switched from JSONRPC to a custom binary format to improve performance under poor network conditions.

It's also worth noting that the `/dev/hidg*` endpoints can sometimes hang indefinitely, so a timeout was introduced to prevent the queue from being blocked forever in such cases.

### DIFF
--- a/hidrpc.go
+++ b/hidrpc.go
@@ -2,6 +2,7 @@ package kvm
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/jetkvm/kvm/internal/hidrpc"
 	"github.com/jetkvm/kvm/internal/usbgadget"
@@ -19,9 +20,14 @@ func onHidMessage(data []byte, session *Session) {
 	)
 
 	if err := hidrpc.Unmarshal(data, &message); err != nil {
-		logger.Warn().Err(err).Msg("failed to unmarshal HID RPC message")
+		logger.Warn().Err(err).Bytes("data", data).Msg("failed to unmarshal HID RPC message")
 		return
 	}
+
+	scopedLogger := hidRpcLogger.With().Str("payload", message.String()).Logger()
+
+	scopedLogger.Debug().Msg("received HID RPC message from the queue")
+	startTime := time.Now()
 
 	switch message.Type() {
 	case hidrpc.TypeHandshake:
@@ -62,6 +68,9 @@ func onHidMessage(data []byte, session *Session) {
 	if rpcErr != nil {
 		logger.Warn().Err(rpcErr).Msg("failed to handle HID RPC message")
 	}
+
+	duration := time.Since(startTime)
+	scopedLogger.Debug().Dur("duration", duration).Msg("handled HID RPC message")
 }
 
 func handleHidRpcKeyboardInput(message hidrpc.Message) (*usbgadget.KeysDownState, error) {
--- a/hidrpc.go
+++ b/hidrpc.go
@@ -8,7 +8,7 @@ import (
 	"github.com/jetkvm/kvm/internal/usbgadget"
 )
 
-func handleHidRpcMessage(message hidrpc.Message, session *Session) {
+func handleHidRPCMessage(message hidrpc.Message, session *Session) {
 	var rpcErr error
 
 	switch message.Type() {
@@ -22,11 +22,11 @@ func handleHidRpcMessage(message hidrpc.Message, session *Session) {
 			logger.Warn().Err(err).Msg("failed to send handshake message")
 			return
 		}
-		session.hidRpcAvailable = true
+		session.hidRPCAvailable = true
 	case hidrpc.TypeKeypressReport, hidrpc.TypeKeyboardReport:
-		keysDownState, err := handleHidRpcKeyboardInput(message)
+		keysDownState, err := handleHidRPCKeyboardInput(message)
 		if keysDownState != nil {
-			reportHidRpcKeysDownState(*keysDownState, session)
+			session.reportHidRPCKeysDownState(*keysDownState)
 		}
 		rpcErr = err
 	case hidrpc.TypePointerReport:
@@ -53,7 +53,7 @@ func handleHidRpcMessage(message hidrpc.Message, session *Session) {
 }
 
 func onHidMessage(data []byte, session *Session) {
-	scopedLogger := hidRpcLogger.With().Bytes("data", data).Logger()
+	scopedLogger := hidRPCLogger.With().Bytes("data", data).Logger()
 	scopedLogger.Debug().Msg("HID RPC message received")
 
 	if len(data) < 1 {
@@ -74,7 +74,7 @@ func onHidMessage(data []byte, session *Session) {
 
 	r := make(chan interface{})
 	go func() {
-		handleHidRpcMessage(message, session)
+		handleHidRPCMessage(message, session)
 		r <- nil
 	}()
 	select {
@@ -85,7 +85,7 @@ func onHidMessage(data []byte, session *Session) {
 	}
 }
 
-func handleHidRpcKeyboardInput(message hidrpc.Message) (*usbgadget.KeysDownState, error) {
+func handleHidRPCKeyboardInput(message hidrpc.Message) (*usbgadget.KeysDownState, error) {
 	switch message.Type() {
 	case hidrpc.TypeKeypressReport:
 		keypressReport, err := message.KeypressReport()
@@ -108,7 +108,7 @@ func handleHidRpcKeyboardInput(message hidrpc.Message) (*usbgadget.KeysDownState
 	return nil, fmt.Errorf("unknown HID RPC message type: %d", message.Type())
 }
 
-func reportHidRpc(params any, session *Session) {
+func reportHidRPC(params any, session *Session) {
 	var (
 		message []byte
 		err     error
@@ -118,6 +118,8 @@ func reportHidRpc(params any, session *Session) {
 		message, err = hidrpc.NewKeyboardLedMessage(params).Marshal()
 	case usbgadget.KeysDownState:
 		message, err = hidrpc.NewKeydownStateMessage(params).Marshal()
+	default:
+		err = fmt.Errorf("unknown HID RPC message type: %T", params)
 	}
 
 	if err != nil {
@@ -135,16 +137,16 @@ func reportHidRpc(params any, session *Session) {
 	}
 }
 
-func reportHidRpcKeyboardLedState(state usbgadget.KeyboardState, session *Session) {
-	if !session.hidRpcAvailable {
-		writeJSONRPCEvent("keyboardLedState", state, currentSession)
+func (s *Session) reportHidRPCKeyboardLedState(state usbgadget.KeyboardState) {
+	if !s.hidRPCAvailable {
+		writeJSONRPCEvent("keyboardLedState", state, s)
 	}
-	reportHidRpc(state, session)
+	reportHidRPC(state, s)
 }
 
-func reportHidRpcKeysDownState(state usbgadget.KeysDownState, session *Session) {
-	if !session.hidRpcAvailable {
-		writeJSONRPCEvent("keysDownState", state, currentSession)
+func (s *Session) reportHidRPCKeysDownState(state usbgadget.KeysDownState) {
+	if !s.hidRPCAvailable {
+		writeJSONRPCEvent("keysDownState", state, s)
 	}
-	reportHidRpc(state, session)
+	reportHidRPC(state, s)
 }
--- a/hidrpc.go
+++ b/hidrpc.go
@@ -109,6 +109,16 @@ func handleHidRPCKeyboardInput(message hidrpc.Message) (*usbgadget.KeysDownState
 }
 
 func reportHidRPC(params any, session *Session) {
+	if session == nil {
+		logger.Warn().Msg("session is nil, skipping reportHidRPC")
+		return
+	}
+
+	if !session.hidRPCAvailable || session.HidChannel == nil {
+		logger.Warn().Msg("HID RPC is not available, skipping reportHidRPC")
+		return
+	}
+
 	var (
 		message []byte
 		err     error
--- a/hidrpc.go
+++ b/hidrpc.go
@@ -8,26 +8,8 @@ import (
 	"github.com/jetkvm/kvm/internal/usbgadget"
 )
 
-func onHidMessage(data []byte, session *Session) {
-	if len(data) < 1 {
-		logger.Warn().Int("length", len(data)).Msg("received empty data in HID RPC message handler")
-		return
-	}
-
-	var (
-		message hidrpc.Message
-		rpcErr  error
-	)
-
-	if err := hidrpc.Unmarshal(data, &message); err != nil {
-		logger.Warn().Err(err).Bytes("data", data).Msg("failed to unmarshal HID RPC message")
-		return
-	}
-
-	scopedLogger := hidRpcLogger.With().Str("payload", message.String()).Logger()
-
-	scopedLogger.Debug().Msg("received HID RPC message from the queue")
-	startTime := time.Now()
+func handleHidRpcMessage(message hidrpc.Message, session *Session) {
+	var rpcErr error
 
 	switch message.Type() {
 	case hidrpc.TypeHandshake:
@@ -68,9 +50,39 @@ func onHidMessage(data []byte, session *Session) {
 	if rpcErr != nil {
 		logger.Warn().Err(rpcErr).Msg("failed to handle HID RPC message")
 	}
+}
 
-	duration := time.Since(startTime)
-	scopedLogger.Debug().Dur("duration", duration).Msg("handled HID RPC message")
+func onHidMessage(data []byte, session *Session) {
+	scopedLogger := hidRpcLogger.With().Bytes("data", data).Logger()
+	scopedLogger.Debug().Msg("HID RPC message received")
+
+	if len(data) < 1 {
+		scopedLogger.Warn().Int("length", len(data)).Msg("received empty data in HID RPC message handler")
+		return
+	}
+
+	var message hidrpc.Message
+
+	if err := hidrpc.Unmarshal(data, &message); err != nil {
+		scopedLogger.Warn().Err(err).Msg("failed to unmarshal HID RPC message")
+		return
+	}
+
+	scopedLogger = scopedLogger.With().Str("descr", message.String()).Logger()
+
+	t := time.Now()
+
+	r := make(chan interface{})
+	go func() {
+		handleHidRpcMessage(message, session)
+		r <- nil
+	}()
+	select {
+	case <-time.After(1 * time.Second):
+		scopedLogger.Warn().Msg("HID RPC message timed out")
+	case <-r:
+		scopedLogger.Debug().Dur("duration", time.Since(t)).Msg("HID RPC message handled")
+	}
 }
 
 func handleHidRpcKeyboardInput(message hidrpc.Message) (*usbgadget.KeysDownState, error) {
--- a/hidrpc.go
+++ b/hidrpc.go
@@ -1,0 +1,129 @@
+package kvm
+
+import (
+	"fmt"
+
+	"github.com/jetkvm/kvm/internal/hidrpc"
+	"github.com/jetkvm/kvm/internal/usbgadget"
+)
+
+func onHidMessage(data []byte, session *Session) {
+	if len(data) < 1 {
+		logger.Warn().Int("length", len(data)).Msg("received empty data in HID RPC message handler")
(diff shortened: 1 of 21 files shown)


hidrpc.go
```diff
@@ -0,0 +1,150 @@
...
+	var (
+		message []byte
+		err     error
+	)
+	switch params := params.(type) {
+	case usbgadget.KeyboardState:
+		message, err = hidrpc.NewKeyboardLedMessage(params).Marshal()
+	case usbgadget.KeysDownState:
+		message, err = hidrpc.NewKeydownStateMessage(params).Marshal()
+	}
```
IDisposable, line 133: maybe add `default: return` here because if it's not one of those messages that return any response?

internal/hidrpc/hidrpc.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/jetkvm/kvm/internal/usbgadget"
+)
+
+// HID RPC is a variable-length packet format that is used to exchange keyboard and mouse events between the client and the server.
+// The packet format is as follows:
+// 1 byte: Event Type
+
+// MessageType is the type of the HID RPC message
+type MessageType uint8
```
IDisposable: _technically_ it's a `byte` :)

internal/usbgadget/hid_keyboard.go
```diff
@@ -86,6 +86,12 @@ type KeyboardState struct {
 	Compose    bool `json:"compose"`
 	Kana       bool `json:"kana"`
```
IDisposable, line 87: I wished I had killed off these bools already :(

internal/usbgadget/hid_keyboard.go
```diff
@@ -151,7 +157,9 @@ func (u *UsbGadget) updateKeyDownState(state KeysDownState) {
 	u.keysDownState = state
```
IDisposable: I was wondering if we should really u.keyboardStateLock.Unlock() before the trace (would require storing off the old value), but **at minimum** releasing the lock before we call the u.onKeysDownChange because we have no idea how long that callback will run, and likely if it needed to acquire a lock (not likely) it should do what it needs self-sufficiently. Something like 

```golang
func (u *UsbGadget) updateKeyDownState(state KeysDownState) {
    var currentState KeysDownState

    // lock scope
    {
	    u.keyboardStateLock.Lock()
	    defer u.keyboardStateLock.Unlock()
        currentState := u.keysDownState
	
    	if currentState.Modifier == state.Modifier &&
	    	bytes.Equal(currentState.Keys, state.Keys) {
    		return // No change in key down state
    	}
	    u.keysDownState = state
    }

	u.log.Trace().Interface("old", currentState).Interface("new", state).Msg("keysDownState updated")
	
    if u.onKeysDownChange != nil {
		u.log.Trace().Interface("state", state).Msg("calling onKeysDownChange")
		(*u.onKeysDownChange)(state)
		u.log.Trace().Interface("state", state).Msg("onKeysDownChange called")
	}
}
```

internal/usbgadget/utils.go
```diff
@@ -107,6 +110,23 @@ func compareFileContent(oldContent []byte, newContent []byte, looserMatch bool)
...
+	n, err = file.Write(data)
+	if err == nil {
+		return
+	}
+
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		err = nil
+	}
+
+	return
```
IDisposable, line 135: ... so this will return a `nil` for the `err`, effectively, when the deadline expires which means the callers will silently not know that an timeout occurred. That means that there will be nothing logged (might want to at least log something a like 124) AND ALSO we won't trigger the "close the hid file" attempt at recovery (which might be okay, not sure of the intent)

ui/src/hooks/stores.ts
```diff
@@ -105,6 +105,12 @@ export interface RTCState {
   setRpcDataChannel: (channel: RTCDataChannel) => void;
   rpcDataChannel: RTCDataChannel | null;
 
+  rpcHidProtocolVersion: number | null;
+  setRpcHidProtocolVersion: (version: number) => void;
+
+  setRpcHidChannel: (channel: RTCDataChannel) => void;
+  rpcHidChannel: RTCDataChannel | null;
```
IDisposable: _nit_ swap these two lines to be consistent with the pattern?

ui/src/hooks/useHidRpc.ts
```diff
@@ -0,0 +1,251 @@
...
+    throw new Error(`Invalid HID RPC message length: ${data.length}`);
+  }
+
+  const payload = data.slice(1);
+
+  switch (data[0]) {
+    case HID_RPC_MESSAGE_TYPES.Handshake:
+      return {
+        type: HID_RPC_MESSAGE_TYPES.Handshake,
+      };
```
IDisposable: Would we be better served to actually return a protocol version in the handshake for future use?

ui/src/hooks/useHidRpc.ts
```diff
@@ -0,0 +1,251 @@
...
+      console.debug("Received HID RPC message", e.data);
+
+      const message = unmarshalHidRpcMessage(new Uint8Array(e.data));
+      if (!message) {
+        console.warn("Received invalid HID RPC message", e.data);
+        return;
+      }
+
+      if (message.type === HID_RPC_MESSAGE_TYPES.Handshake) {
+        setRpcHidProtocolVersion(1);
```
IDisposable: Can we make the handshake response actually return a version?

ui/src/hooks/useKeyboard.ts
```diff
@@ -19,17 +20,36 @@ export default function useKeyboard() {
...
+  });
 
   // sendKeyboardEvent is used to send the full keyboard state to the device for macro handling
   // and resetting keyboard state. It sends the keys currently pressed and the modifier state.
   // The device will respond with the keysDownState if it supports the keyPressReport API
   // or just accept the state if it does not support (returning no result)
   const sendKeyboardEvent = useCallback(
     async (state: KeysDownState) => {
-      if (rpcDataChannel?.readyState !== "open") return;
+      if (rpcDataChannel?.readyState !== "open" && !rpcHidReady) return;
```
IDisposable, line 49: `&&` or `||`? Seems like we wanna skip it if _either_ is not ready? Or is the idea that if the old-school (`send` method) is not ready **nor** is the new-wave distinct RPC channel ready...

webrtc.go
```diff
@@ -105,17 +108,51 @@ func newSession(config SessionConfig) (*Session, error) {
 		scopedLogger.Warn().Err(err).Msg("Failed to create PeerConnection")
 		return nil, err
 	}
+
 	session := &Session{peerConnection: peerConnection}
 	session.rpcQueue = make(chan webrtc.DataChannelMessage, 256)
+	session.hidQueue = make(chan webrtc.DataChannelMessage, 1024)
```
IDisposable: That's a lot of backlog... might be a bit excessive?

webrtc.go
```diff
@@ -105,17 +108,51 @@ func newSession(config SessionConfig) (*Session, error) {
...
+	go func() {
+		for msg := range session.hidQueue {
+			onHidMessage(msg.Data, session)
 		}
 	}()
 
 	peerConnection.OnDataChannel(func(d *webrtc.DataChannel) {
+		defer func() {
+			if r := recover(); r != nil {
+				scopedLogger.Warn().Interface("error", r).Msg("Recovered from panic in DataChannel handler")
```
IDisposable:
```golang
scopedLogger.Error().Err(r),Msg("Recovered from panic in DataChannel handler")
```

 so it's an error-level log, and has the error details formatted as an Error

IDisposable (approved):
Wow, that was fast :)  suggestions/questions are not blockers

webrtc.go
ym: Yes, given that we're introducing a very short timeout for hid write, setting it to `1024` here seems unnecessary.

ui/src/hooks/useKeyboard.ts
ym: It is skipped only if both RPC methods are unavailable, so use `&&` here.

webrtc.go
IDisposable: I think we still want to use `.Err(r)` instead of `.Interface("Error", r)` here.

ym: I felt the same until I found that the return type of `recover()` is `interface{}`, rather than `Error`. Are you surprised too? :-)

IDisposable: Not really surprised, since as far as I can tell `Interface {}` is just `any` 😄

The difference is I was going for is `Err(` is (to the logger) known to be an error object, so the logging doesn't need reflection to log out the details as `Interface(` does.

Either is fine, as hopefully we never hit this code anyway 😁

internal/usbgadget/utils.go
ym: Yes, I set it to nil intentionally, as it's likely to generate lots of logs when it times out. However, we could also use `logWithSuppression` instead.

ym: Yes, I set it to nil intentionally, as it's likely to generate lots of logs when it times out. However, we could also use `logWithSuppression` instead.

internal/usbgadget/hid_keyboard.go
ym: This was kept for backwards compatibility.

internal/usbgadget/hid_keyboard.go
ym: added an extra trace log before the lock

ui/src/components/WebRTCVideo.tsx
```diff
@@ -243,11 +257,21 @@ export default function WebRTCVideo() {
   const sendAbsMouseMovement = useCallback(
     (x: number, y: number, buttons: number) => {
       if (settings.mouseMode !== "absolute") return;
-      send("absMouseReport", { x, y, buttons });
+      if (rpcHidReady) {
+        reportAbsMouseEvent(x, y, buttons);
+      } else {
+        send("absMouseReport", { x, y, buttons });
```
adamshiervani: Could you comment why we're doing this for the future

ui/src/components/WebRTCVideo.tsx
```diff
@@ -222,10 +224,22 @@ export default function WebRTCVideo() {
       if (settings.mouseMode !== "relative") return;
       // if we ignore the event, double-click will not work
       // if (x === 0 && y === 0 && buttons === 0) return;
-      send("relMouseReport", { dx: calcDelta(x), dy: calcDelta(y), buttons });
+      const dx = calcDelta(x);
+      const dy = calcDelta(y);
+      if (rpcHidReady) {
+        reportRelMouseEvent(dx, dy, buttons);
+      } else {
+        send("relMouseReport", { dx, dy, buttons });
```
adamshiervani: Could you comment why we're doing this for the future

ui/src/hooks/useKeyboard.ts
```diff
@@ -1,13 +1,15 @@
 import { useCallback } from "react";
```
adamshiervani, line 1: The reason this hook contains so much fallback (e.g. falling back on `keyboardReport`) logic is that PR #725 assumed the old code was already deployed on devices. In cloud mode, this mismatch could cause a mismatch: new JS code expecting RPC methods that the old Go code didn’t yet have, as it wasn't updated.

This PR, however, isn’t bound by that constraint, since #725 never reached production. That means when we call `keypressReport`, we can simply use the new hidRPC data channel. There’s no need to support both versions of keypressReport (one via hidRPC and one via RPC) -- only the hidRPC variant is required.

internal/usbgadget/consts.go
```diff
@@ -1,3 +1,7 @@
 package usbgadget
 
+import "time"
+
 const dwc3Path = "/sys/bus/platform/drivers/dwc3"
+
+const hidWriteTimeout = 10 * time.Millisecond
```
adamshiervani, line 7: Just a note:

Having a timeout on USB commands is good, but it doesn’t really improve reliability under poor network conditions.

In my limited testing with macOS’s Network Link Conditioner, the main issue isn’t the device failing to execute commands on time. Instead, it’s the delayed arrival of messages to release the key. As a result, the remote OS often interprets it as a stuck keypress — leading to `aaaaaaaaaaaaaa` far too often on bad networks.

adamshiervani (changes requested):

internal/usbgadget/consts.go
ym: The timeout added here is specifically to address a USB gadget issue:
In some cases, writing to the endpoint can block indefinitely if the target device isn't responding, for example, hibernation.

The issue was exposed after this PR:

https://github.com/jetkvm/kvm/commit/8e27cd6b60a6229f0a370bd2e652b3a7a6329bfa

ui/src/components/WebRTCVideo.tsx
ym: This is kept for backward compatibility with the cloud version. We always deploy the latest UI to [app.jetkvm.com](https://app.jetkvm.com/), but a user's device might not be updated yet to support HidChannel.

ui/src/hooks/useKeyboard.ts
ym: We've already pushed this to the dev channel, and I believe there are still quite a few users using the dev release. So it might be safer to keep it for now, but yes, I’m also okay with removing it if needed.

adamshiervani: Let's not be backwards compatible for those; if you're on the dev channel and got the latest `0.4.7-dev` on the device, you extremely likely have Auto Update turned on. It will effectively be for those who disabled "Auto Update" a few days ago. 

Additionally, regardless of `keyboardReport` on RPC or hidRPC, it will fall back to `keyboardReport` if the method doesn't exist.

ui/src/components/WebRTCVideo.tsx
adamshiervani: Oh, sorry for the lack of specificity. I get that; it's just I would like to have it in the code. By just looking at the code, it's not apparent why you use relMouseReport if the `rpcHidReady` is false

ym: Sorry, I read too quickly and missed some context 😅 - adding comment now

ui/src/hooks/useKeyboard.ts
```diff
@@ -135,9 +139,15 @@ export default function useKeyboard() {
...
 
+      if (rpcHidReady) {
+        console.debug("Sending keypress event via HidRPC");
+        reportKeypressEvent(key, press);
+        return;
+      }
+
       if (keyPressReportApiAvailable) {
         // if the keyPress api is available, we can just send the key press event
         sendKeypressEvent(key, press);
```
adamshiervani: There is a bit of duplicate logic:

- `keyPressReportApiAvailable` isn’t useful, since it’s effectively the same as rpcHidReady.
- `sendKeypressEvent()` duplicates the preceding if statement in this function. Just replace `keyPressReportApiAvailable` with `rpcHidReady` and remove the preceding if statement.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -222,10 +224,23 @@ export default function WebRTCVideo() {
       if (settings.mouseMode !== "relative") return;
       // if we ignore the event, double-click will not work
       // if (x === 0 && y === 0 && buttons === 0) return;
-      send("relMouseReport", { dx: calcDelta(x), dy: calcDelta(y), buttons });
+      const dx = calcDelta(x);
+      const dy = calcDelta(y);
+      if (rpcHidReady) {
```
adamshiervani: I’d suggest not using `useHidRPC()` directly for mouse and keyboard events. Instead, let hidRPC serve purely as a transport layer, with dedicated hooks handling input logic.

We use `useKeyboard()` for keyboard input and `useRPC()` for mouse reports. Mixing approaches like this adds confusion and clutters an already complex HID JS API surface.

A clearer approach would be to introduce both `useMouse()` and `useKeyboard()` hooks. That way, consumption is consistent and obvious, while each hook can encapsulate its own complexity. 

We added `useKeyboard()` out of necessity because its fallback logic was big. That isn’t the case for the mouse, where the fallback is rather simple -- but, I would say the added clarity and consistency still make a `useMouse()` hook worthwhile.

ym: Good point, moving mouse logic to `useMouse`

internal/usbgadget/hid_keyboard.go
IDisposable: I meant when I was doing the stack cleanup, sorry

internal/usbgadget/hid_keyboard.go
IDisposable: My concern (poorly communicated, sorry) is that we're holding the `u.keyboardStateLock.Lock()` for the duration of the method _including_ the call out to `(*u.onKeysDownChange)(state)` which means we're locked for longer than we might intend if the `onKeysDownChange` takes a while to execute... since it's actually used to send data _back_ to the client/browser... that might be a longish time.

My code example above used an extra _lock scope_ to ensure that we release the lock before we continue on to lines 162-166.

ui/src/hooks/hidRpc.ts
```diff
@@ -0,0 +1,303 @@
...
+    constructor(messageType: HidRpcMessageType) {
+        this.messageType = messageType;
+    }
+
+    marshal(): Uint8Array {
+        throw new Error("Not implemented");
+    }
+
+    // @ts-expect-error: this is a base class, so we don't need to implement it
+    public static unmarshal(data: Uint8Array): RpcMessage | undefined {
```
IDisposable: To make lint happy use `_data` here (only)

webrtc.go
```diff
@@ -105,17 +129,64 @@ func newSession(config SessionConfig) (*Session, error) {
...
+			}
+		}()
+
 		scopedLogger.Info().Str("label", d.Label()).Uint16("id", *d.ID()).Msg("New DataChannel")
+
 		switch d.Label() {
+		case "hidrpc":
+			session.HidChannel = d
+			d.OnMessage(func(msg webrtc.DataChannelMessage) {
+				l := scopedLogger.With().Str("data", string(msg.Data)).Int("length", len(msg.Data)).Logger()
```
IDisposable: The call to `string(msg.Data)` will execute here and allocate/copy the data to a string even if we subsequently do not call any `.Warn()` or `.Trace()` (and at trace logging level). If we use .Bytes("data", msg.Data) like we do in [_jsonrpc_](https://github.com/jetkvm/kvm/blob/dev/jsonrpc.go#L414)

ui/src/hooks/useMouse.ts
```diff
@@ -0,0 +1,178 @@
...
+      setMouseMove,
+      mouseMode,
+      rpcHidReady,
+    ],
+  );
+
+  const getRelMouseMoveHandler = useCallback(
+    ({ isPointerLockActive, isPointerLockPossible }: RelMouseMoveHandlerProps) => (e: MouseEvent) => {
+      if (mouseMode !== "relative") return;
+      if (isPointerLockActive === false && isPointerLockPossible) return;
```
IDisposable: I know you didn't actually change this (just moved here from WebRTVideo.tsx), but this seems odd... do we only support relative mouse mode if we can lock the pointer, AND have locked the pointer?

If that's the case, then what happens if we haven't locked? We ignore all mouse events?

ui/src/hooks/stores.ts
```diff
@@ -449,8 +461,8 @@ export interface HidState {
   keysDownState: KeysDownState;
   setKeysDownState: (state: KeysDownState) => void;
 
-  keyPressReportApiAvailable: boolean;
-  setkeyPressReportApiAvailable: (available: boolean) => void;
+  // keyPressReportApiAvailable is no longer needed, we'll simply use hidChannel available to
```
IDisposable: Could probably just delete this commented-out code.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -211,129 +211,32 @@ export default function WebRTCVideo() {
       }
     };
 
-    document.addEventListener("fullscreenchange ", handleFullscreenChange);
+    document.addEventListener("fullscreenchange", handleFullscreenChange);
```
IDisposable, line 214: WOW, nice catch... this could explain some of the oddities of focus when going in and out of fullscreen repeatedly

IDisposable (approved):
:shipit: (we can discuss the relative mouse locked mode elsewhere)

ui/src/hooks/useMouse.ts
ym: Yes, it's definitely odd ... Let me just remove it

IDisposable:
:shipit:

PR #761, opened 2025-09-02
Clean up Virtual Keyboard styling

adamshiervani:
A copy of https://github.com/jetkvm/kvm/pull/750 as I don't write access to do the last touches.


IDisposable (approved):
Looks good!

PR #776, opened 2025-09-03
Upgrade UI packages

IDisposable:
| Package                                     | From     | To         |
|-----------------------------------|----------|----------|
| dayjs                                           | 1.1.13   | 1.11.18 |
| react-simple-keyboard               | 3.8.115 | 3.8.119 |
| @types/react                              | 19.1.11 | 19.1.12 |
| @types/react-dom                     | 19.1.8   | 19.1.9   |
|  @types/semver                         | 7.7.0      | 7.7.1    |
| @types/validator                        | 13.15.2 | 13.15.3 |
| @typescript-eslint/eslint-plugin | 8.41.0   | 8.42.0   |
| @typescript-eslint/parser           | 8.41.0   | 8.42.0   |

Also fixed lint warning about the missing autocomplete on the password field

PR #777, opened 2025-09-03
Bumps recharts to 3.1.2

IDisposable:
This is a major version upgrade, but the charts we have seem to work fine.
Builds on/includes #776 

Note that this version now includes its own Animation sub-package, which me might consider moving toward?


IDisposable:
The full set of updates is in #778 (except Zustand)

PR #779, opened 2025-09-04
feat(ui) Fix the CapsLock and Shift key for VirtualKeyboard

IDisposable:
In the previous 0.4.x releases the Virtual Keyboard layout (key caps shown) would take into account the Caps Lock key (showing the shifted layout) and reverse the layout back to default when they user pressed the shift key. 

This is the standard behavior for PCs running Linux or Windows (but doesn't always work that way on MacOS). Once we start tracking the actual installed keyboard on the remote/host device we can do a better job, but this PR restores the previous behavior

### DIFF
--- a/ui/src/components/VirtualKeyboard.tsx
+++ b/ui/src/components/VirtualKeyboard.tsx
@@ -47,7 +47,7 @@ function KeyboardWrapper() {
   }, [keysDownState]);
 
   const isCapsLockActive = useMemo(() => {
-    return keyboardLedState?.caps_lock || false;
+    return keyboardLedState.caps_lock;
   }, [keyboardLedState]);
 
   const mainLayoutName = useMemo(() => {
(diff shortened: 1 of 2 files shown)


ui/src/components/VirtualKeyboard.tsx
```diff
@@ -46,9 +46,15 @@ function KeyboardWrapper() {
     return decodeModifiers(keysDownState.modifier);
   }, [keysDownState]);
 
+  const isCapsLockActive = useMemo(() => {
+    return keyboardLedState?.caps_lock || false;
```
adamshiervani: Let's set default values in the Zustand store instead. I'd rather do this in the store than in all consumers.

adamshiervani (commented):
Tested and works!

ui/src/components/VirtualKeyboard.tsx
IDisposable: Sure, this is just a coalesce if we don't have LED state yet (never should actually fire) :)

IDisposable:
Rebased

ui/src/components/VirtualKeyboard.tsx
IDisposable: Done

PR #780, opened 2025-09-04
chore: upgrade deps

ym:
I've already checked the changelog and release notes, and run the tests. It should be safe to upgrade.


adamshiervani (approved):
Tested and works

Issue #648, opened 2025-06-27
Paste text prefix character randomly dropped


kmahyyg:
I tried to paste text using Web "Paste text" button to Windows MS Teams running on victim / MSEdge in RDP Client running on victim.

The text originally is:
```
C:\Users\xxxx\a\really\long\long\long\path1
C:\Users\xxxx\a\really\long\long\long\path2
C:\Users\xxxx\a\really\long\long\long\path3
C:\Users\xxxx\a\really\long\long\long\path4
```

After I click paste, on remote machine, it became:
```
C:\Users\xxxx\a\really\long\long\long\path1
rs\xxxx\a\really\long\long\long\path2
x\a\really\long\long\long\path3
ly\long\long\long\path4
```

JetKVM Version: App 0.4.4 System 0.2.4
Controller side: MacOS 15 Chrome Latest (keyboard layout: en-us-Colemak) with M1 Chip, no antivirus, 3840x2160 resolution
Victim side (controlled by JetKVM): Windows 11 22H2, Crowdstrike Antivirus, 1920x1080 resolution.

Typing and Keyboard shortcut in general is working correctly.

BTW, there's also some mouse drift, I had to change to absolute mouse movement and struggling clicking it.

IDisposable:
That sounds like the key-released events aren't being sent reliably. Might need to have a "Paste Speed" setting to determine how long between keystrokes when pasting.

That said, the refactored keyboard code in PR #497 does NOT have this problem (as that's my daily driver). Maybe we can get that merged in eventually

ym:
The latest version (0.4.9) should have this issue fixed. Could you please try it and let us know if it works for you?

kmahyyg:
dears, 

have you already created a release for 0.4.9? I've checked my device, it still shows v0.4.8 is latest.

thank you!

PR #789, opened 2025-09-09
feat: send all paste keystrokes to backend

ym:


jsonrpc.go
```diff
@@ -1049,88 +1050,194 @@ func rpcSetLocalLoopbackOnly(enabled bool) error {
...
+		// Check for cancellation before each step
+		select {
+		case <-ctx.Done():
+			logger.Debug().Msg("Keyboard report multi context cancelled")
+			return last, ctx.Err()
+		default:
+		}
+
+		var modifier byte
+		if m, ok := step["modifier"].(float64); ok {
```
IDisposable: `float64` seems odd, why would it transmit like that since it's actually known to be a `uint8`/`byte`?

I think we would want to try these types in exactly the opposite order.

jsonrpc.go
```diff
@@ -1049,88 +1050,194 @@ func rpcSetLocalLoopbackOnly(enabled bool) error {
...
+			modifier = byte(mi)
+		} else if mb, ok := step["modifier"].(uint8); ok {
+			modifier = mb
+		}
+
+		var keys []byte
+		if arr, ok := step["keys"].([]any); ok {
+			keys = make([]byte, 0, len(arr))
+			for _, v := range arr {
+				if f, ok := v.(float64); ok {
```
IDisposable: ... and here?

ui/src/components/popovers/PasteModal.tsx
```diff
@@ -46,21 +35,26 @@ export default function PasteModal() {
...
+    cancelExecuteMacro();
     setDisableVideoFocusTrap(false);
     setInvalidChars([]);
-  }, [setDisableVideoFocusTrap, setPasteModeEnabled]);
+  }, [setDisableVideoFocusTrap, cancelExecuteMacro]);
 
   const onConfirmPaste = useCallback(async () => {
-    setPasteModeEnabled(false);
-    setDisableVideoFocusTrap(false);
+    // setPasteModeEnabled(false);
```
IDisposable: Do these need to be restored or commented-out code deleted? I'm a bit concerned about the focus trap logic as a whole....

ui/src/hooks/useKeyboard.ts
```diff
@@ -48,69 +54,82 @@ export default function useKeyboard() {
...
-      } else {
-        // This is a delay-only step, just wait for the delay amount
-        await new Promise(resolve => setTimeout(resolve, step.delay || 50));
-      }
-
-      // Add a small pause between steps if not the last step
-      if (index < steps.length - 1) {
-        await new Promise(resolve => setTimeout(resolve, 10));
+        macro.push({ keys: keyValues, modifier: modifierMask });
+        keysDownState.keys.length = hidKeyBufferSize;
```
IDisposable: This block is identical to lines 85-87... perhaps a helper?

ui/src/routes/devices.$id.tsx
```diff
@@ -598,6 +598,12 @@ export default function KvmIdRoute() {
       setUsbState(usbState);
     }
 
+    if (resp.method === "keyboardReportMultiState") {
+      const reportMultiState = resp.params as unknown as boolean;
+      console.debug("Setting keyboard report multi state", reportMultiState);
```
IDisposable: Is this to so the device go code can stop the pasting?

IDisposable (commented):
Looks pretty clean, just a few questions.

ui/src/routes/devices.$id.tsx
ym: No, this part is just to notify the client that the pasting process has finished or been cancelled. `cancelKeyboardReportMulti` is used to cancel an ongoing paste process.

adamshiervani (commented):
I just spent some time collecting my thoughts, and as far as I see it, there’s really only one keyboard issue left.

**Missing Keys**  
Earlier I thought the remote host was dropping keystrokes. I was using Windows 11 Notepad for testing, assuming it would be the most “neutral” environment. After digging deeper, I realized Notepad actually has **Spellcheck** and **Autocorrect**, which can swallow keystrokes ([Reddit thread](https://www.reddit.com/r/WindowsHelp/comments/1iolltk/windows_notepad_missing_keystrokes/)).  

Once I disabled those features, input worked as expected. I’ve since switched to a dedicated keyboard testing tool on the remote host...

That said, we’ve had reports of missing keys outside of Notepad as well ([#648](https://github.com/jetkvm/kvm/issues/648)). But given that events are processed sequentially - both via `rpc` and `hidRPC` - I consider this a non-issue until we have a clear and reliable STR.

**Repeating Keys**  
The most common [keyboard-related issues](https://github.com/jetkvm/kvm/issues?q=label:%22component:%20hid/keyboard%22) are about key repeats. This is definitely the biggest remaining keyboard problem.  

The root cause is that the **key release event is delayed**, leading the device to interpret it as a long press. The challenge is distinguishing between a genuine long press and an accidental one caused by latency.  

To reproduce the issue, simply introduce network instability with a tool like [Network Link Conditioner](https://nshipster.com/network-link-conditioner/) - unintended long presses appear quickly in normal use.  

Image

My best idea to solve this is a *deadman switch*: sending keep-alive signals for real long presses, so the system can safely assume that missing keep-alives mean “release the key.” This would require extensive testing but seems like the most robust approach.  

**This PR**  
I’m not keen on adding a new HID-related method to the `rpc` datachannel, since we’re moving away from it for HID messages.

Also, `keyboardReportMulti` currently ignores the `delay` parameter from macros, so it isn’t backwards-compatible with existing macros.  

**Moving Forward**  
I suggest we narrow the scope of this PR and focus squarely on solving the key repeat problem.  

1. **Slimming down this PR**  
The Paste Modal in `dev` currently bypasses `useKeyboard()` and calls `send('keyboardReport', [...])` directly.

This PR migrates that to the macro format and calls `executeMacros()`. That’s the right direction, since all keyboard calls should go through `useKeyboard()`, which in turn tries to use the new hidRPC data channel. Let’s keep this part and leave out the rest for now. 

2. **Unintended long-presses**  

We need a clear way to differentiate real long presses from ones caused by latency. My proposal is to send periodic keep-alive signals (during a legitimate long press??), so the device knows the key is still being held. If the “up” event doesn’t arrive within a safe threshold and no keep-alive is present, the key should be released automatically.  

@ym since you implemented the `hidRPC`, could you take the lead on this? I’m happy to support, but my strongest contribution here would be testing and iterating on the keep-alive approach.

alexballas:
@adamshiervani the issue with the repeating keys is not about key releases getting delayed but the opposite. Key releases registering before the key down. This leads to the continued key presses which only stop if you trigger another key press and expect the key release to be at the end of the sequence. Network stability has nothing to do with the issue.

Going back to my initial code suggestion, I dont know if the full rework of the queue was mandatory or if it made an actual difference. The bit that indeed made a difference however was the 10msec timeout when writing to the device.

adamshiervani:
Using the latest `dev` branch, my logs show that the core problem is key releases arriving "late" in bad network conditions.

I can reproduce this reliably by simulating poor network conditions with [Network Link Conditioner](https://nshipster.com/network-link-conditioner/).

On a remote Ubuntu host, executing a paste from the Paste Modal while monitoring with `libinput debug-events`, I consistently see sequences like this:

<details>

<summary>Logs from libinput</summary>

```
ubuntu@ubuntu:~$ sudo libinput debug-events
-event2   DEVICE_ADDED            Power Button                      seat0 default group1  cap:k
-event3   DEVICE_ADDED            Video Bus                         seat0 default group2  cap:k
-event1   DEVICE_ADDED            Power Button                      seat0 default group3  cap:k
-event0   DEVICE_ADDED            Sleep Button                      seat0 default group4  cap:k
-event4   DEVICE_ADDED            JetKVM USB Emulation Device       seat0 default group5  cap:k
-event5   DEVICE_ADDED            JetKVM USB Emulation Device       seat0 default group5  cap:p left scroll-nat calib
-event6   DEVICE_ADDED            JetKVM USB Emulation Device       seat0 default group5  cap:p left scroll-nat scroll-button
-event7   DEVICE_ADDED            JetKVM USB Emulation Device       seat0 default group6  cap:k
-event8   DEVICE_ADDED            JetKVM USB Emulation Device       seat0 default group6  cap:p left scroll-nat calib
-event9   DEVICE_ADDED            JetKVM USB Emulation Device       seat0 default group6  cap:p left scroll-nat scroll-button
-event8   POINTER_MOTION_ABSOLUTE +0.000s	 26.02/ 68.41
 event8   POINTER_MOTION_ABSOLUTE +0.016s	 25.29/ 71.13
 event8   POINTER_MOTION_ABSOLUTE +0.033s	 24.71/ 75.41
 event8   POINTER_MOTION_ABSOLUTE +0.049s	 24.27/ 83.18
 event8   POINTER_MOTION_ABSOLUTE +0.067s	 23.91/ 89.40
 event8   POINTER_MOTION_ABSOLUTE +1.459s	  0.00/  0.00
 event8   POINTER_MOTION_ABSOLUTE +1.921s	 40.89/ 54.94
 event8   POINTER_MOTION_ABSOLUTE +3.078s	 39.14/ 12.83
 event8   POINTER_MOTION_ABSOLUTE +6.128s	  3.79/ 19.96
 event8   POINTER_MOTION_ABSOLUTE +6.176s	 42.35/  3.11
-event7   KEYBOARD_KEY            +20.321s	*** (-1) pressed
a event7   KEYBOARD_KEY            +20.388s	*** (-1) released
 event7   KEYBOARD_KEY            +20.450s	*** (-1) pressed
b event7   KEYBOARD_KEY            +20.514s	*** (-1) released
 event7   KEYBOARD_KEY            +21.587s	*** (-1) pressed
c event7   KEYBOARD_KEY            +21.656s	*** (-1) released
 event7   KEYBOARD_KEY            +21.726s	*** (-1) pressed
d event7   KEYBOARD_KEY            +21.792s	*** (-1) released
 event7   KEYBOARD_KEY            +21.855s	*** (-1) pressed
e event7   KEYBOARD_KEY            +21.921s	*** (-1) released
 event7   KEYBOARD_KEY            +21.987s	*** (-1) pressed
f event7   KEYBOARD_KEY            +22.053s	*** (-1) released
 event7   KEYBOARD_KEY            +22.120s	*** (-1) pressed
g event7   KEYBOARD_KEY            +22.188s	*** (-1) released
 event7   KEYBOARD_KEY            +22.253s	*** (-1) pressed
hhhhhhhhhhhhhhhhhh event7   KEYBOARD_KEY            +23.255s	*** (-1) released
 event7   KEYBOARD_KEY            +23.324s	*** (-1) pressed
i event7   KEYBOARD_KEY            +23.394s	*** (-1) released
 event7   KEYBOARD_KEY            +23.462s	*** (-1) pressed
j event7   KEYBOARD_KEY            +23.528s	*** (-1) released
 event7   KEYBOARD_KEY            +24.224s	*** (-1) pressed
k event7   KEYBOARD_KEY            +24.310s	*** (-1) released
 event7   KEYBOARD_KEY            +24.380s	*** (-1) pressed
l event7   KEYBOARD_KEY            +24.450s	*** (-1) released
 event7   KEYBOARD_KEY            +24.517s	*** (-1) pressed
m event7   KEYBOARD_KEY            +24.990s	*** (-1) released
 event7   KEYBOARD_KEY            +25.522s	*** (-1) pressed
n event7   KEYBOARD_KEY            +26.004s	*** (-1) released
 event7   KEYBOARD_KEY            +26.082s	*** (-1) pressed
o event7   KEYBOARD_KEY            +26.155s	*** (-1) released
 event7   KEYBOARD_KEY            +26.224s	*** (-1) pressed
p event7   KEYBOARD_KEY            +26.292s	*** (-1) released
 event7   KEYBOARD_KEY            +26.359s	*** (-1) pressed
q event7   KEYBOARD_KEY            +26.425s	*** (-1) released
 event7   KEYBOARD_KEY            +26.492s	*** (-1) pressed
r event7   KEYBOARD_KEY            +26.561s	*** (-1) released
 event7   KEYBOARD_KEY            +26.628s	*** (-1) pressed
s event7   KEYBOARD_KEY            +26.698s	*** (-1) released
 event7   KEYBOARD_KEY            +26.767s	*** (-1) pressed
t event7   KEYBOARD_KEY            +26.834s	*** (-1) released
 event7   KEYBOARD_KEY            +26.901s	*** (-1) pressed
uuuuuuuuuuuuuuuuuuuu event7   KEYBOARD_KEY            +27.981s	*** (-1) released
u event7   KEYBOARD_KEY            +28.053s	*** (-1) pressed
v event7   KEYBOARD_KEY            +28.121s	*** (-1) released
 event7   KEYBOARD_KEY            +28.192s	*** (-1) pressed
w event7   KEYBOARD_KEY            +28.257s	*** (-1) released
 event7   KEYBOARD_KEY            +28.325s	*** (-1) pressed
x event7   KEYBOARD_KEY            +28.390s	*** (-1) released
 event7   KEYBOARD_KEY            +28.458s	*** (-1) pressed
yyyyyyyyyyyyyyyyyyyyyyyyyyy event7   KEYBOARD_KEY            +29.757s	*** (-1) released
y event7   KEYBOARD_KEY            +29.835s	*** (-1) pressed
z event7   KEYBOARD_KEY            +29.904s	*** (-1) released
 event7   KEYBOARD_KEY            +30.380s	*** (-1) pressed
```

</details>

As shown in the logs, the repeated characters (`h`, `u`, `y`, etc.) always coincide with cases where there is a gap of >1 second between a press and its matching release.

The code path from the keyboard event to the remote host is complex, and i'm sure there still are issues at different stages. To move forward with a "release before press" debugging session, we need a reproducible scenario and logs that can capture the specific case where a release shows up before its corresponding press.

alexballas:
> Using the latest `dev` branch, my logs show that the core problem is key releases arriving "late" in bad network conditions.
> 
> I can reproduce this reliably by simulating poor network conditions with [Network Link Conditioner](https://nshipster.com/network-link-conditioner/).
> 
> On a remote Ubuntu host, executing a paste from the Paste Modal while monitoring with `libinput debug-events`, I consistently see sequences like this:
> 
> Logs from libinput
> As shown in the logs, the repeated characters (`h`, `u`, `y`, etc.) always coincide with cases where there is a gap of >1 second between a press and its matching release.
> 
> The code path from the keyboard event to the remote host is complex, and i'm sure there still are issues at different stages. To move forward with a "release before press" debugging session, we need a reproducible scenario and logs that can capture the specific case where a release shows up before its corresponding press.

In my test I just mashed pressed "asd".
I got extra presses, plus a repeated key press. In the following logs this can be seen by either checking the sequence ID  (414 comes before 413) or checking the actual events
We first get (key up)
```
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[0,0,0,0,0,0]}}"
```
and then (key down)

```
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[7,0,0,0,0,0]}}"
```
It should be the other way around. I see that we're again using goroutines. So this regressed.
Can you please do a `git checkout d952480c2aa26c7ea6fb0d4db8e88759a0d1a4e3` and test again?
This last commit should give you a working build just after the fix I proposed.


Full logs for reference:
```
2025-09-10T16:48:48Z TRC jetkvm jsonrpc RPC handler returned id=408 method=keyboardReport params={"keys":[],"modifier":0} result={"keys":[0,0,0,0,0,0],"modifier":0}
2025-09-10T16:48:48Z DBG jetkvm websocket ping frame received connectionID=97d0680c9aa85851-ATH length=0 payload= source=api.jetkvm.com sourceType=cloud
2025-09-10T16:48:48Z DBG jetkvm native sending ctrl action action=lv_obj_set_state params={"obj":"ui_Home_Footer_Usb_Status_Label","state":"LV_STATE_DEFAULT"}
2025-09-10T16:48:48Z TRC jetkvm native ctrl sock msg addr=@ data={"seq":55} type=ctrl
2025-09-10T16:48:48Z DBG jetkvm native sending ctrl action action=lv_obj_set_state params={"obj":"ui_Home_Footer_Hdmi_Status_Label","state":"LV_STATE_DEFAULT"}
2025-09-10T16:48:48Z TRC jetkvm native ctrl sock msg addr=@ data={"seq":56} type=ctrl
2025-09-10T16:48:48Z DBG jetkvm native sending ctrl action action=lv_obj_clear_flag params={"flag":"LV_OBJ_FLAG_HIDDEN","obj":"ui_Home_Header_Cloud_Status_Icon"}
2025-09-10T16:48:49Z TRC jetkvm websocket sending ping frame connectionID=97d0680c9aa85851-ATH source=api.jetkvm.com sourceType=cloud
2025-09-10T16:48:49Z TRC jetkvm websocket received pong frame connectionID=97d0680c9aa85851-ATH duration=150.839224ms source=api.jetkvm.com sourceType=cloud
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Received RPC request id=409 method=keyboardReport params={"keys":[7],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Calling RPC handler id=409 method=keyboardReport params={"keys":[7],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[7,0,0,0,0,0]}}"
2025-09-10T16:48:50Z WRN jetkvm jetkvm HID RPC is not available, skipping reportHidRPC
2025-09-10T16:48:50Z TRC jetkvm jsonrpc RPC handler returned id=409 method=keyboardReport params={"keys":[7],"modifier":0} result={"keys":[7,0,0,0,0,0],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Received RPC request id=411 method=keyboardReport params={"keys":[7,4,22],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Calling RPC handler id=411 method=keyboardReport params={"keys":[7,4,22],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[7,4,22,0,0,0]}}"
2025-09-10T16:48:50Z WRN jetkvm jetkvm HID RPC is not available, skipping reportHidRPC
2025-09-10T16:48:50Z TRC jetkvm jsonrpc RPC handler returned id=411 method=keyboardReport params={"keys":[7,4,22],"modifier":0} result={"keys":[7,4,22,0,0,0],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Received RPC request id=410 method=keyboardReport params={"keys":[7,4],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Calling RPC handler id=410 method=keyboardReport params={"keys":[7,4],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[7,4,0,0,0,0]}}"
2025-09-10T16:48:50Z WRN jetkvm jetkvm HID RPC is not available, skipping reportHidRPC
2025-09-10T16:48:50Z TRC jetkvm jsonrpc RPC handler returned id=410 method=keyboardReport params={"keys":[7,4],"modifier":0} result={"keys":[7,4,0,0,0,0],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Received RPC request id=412 method=keyboardReport params={"keys":[7,4],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Calling RPC handler id=412 method=keyboardReport params={"keys":[7,4],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc RPC handler returned id=412 method=keyboardReport params={"keys":[7,4],"modifier":0} result={"keys":[7,4,0,0,0,0],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Received RPC request id=414 method=keyboardReport params={"keys":[],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Calling RPC handler id=414 method=keyboardReport params={"keys":[],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[0,0,0,0,0,0]}}"
2025-09-10T16:48:50Z WRN jetkvm jetkvm HID RPC is not available, skipping reportHidRPC
2025-09-10T16:48:50Z TRC jetkvm jsonrpc RPC handler returned id=414 method=keyboardReport params={"keys":[],"modifier":0} result={"keys":[0,0,0,0,0,0],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Received RPC request id=413 method=keyboardReport params={"keys":[7],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc Calling RPC handler id=413 method=keyboardReport params={"keys":[7],"modifier":0}
2025-09-10T16:48:50Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[7,0,0,0,0,0]}}"
2025-09-10T16:48:50Z WRN jetkvm jetkvm HID RPC is not available, skipping reportHidRPC
2025-09-10T16:48:50Z TRC jetkvm jsonrpc RPC handler returned id=413 method=keyboardReport params={"keys":[7],"modifier":0} result={"keys":[7,0,0,0,0,0],"modifier":0}
```

IDisposable:
I really like the change to using the macro for past, and passing the entire sequence as one RPC message, because that limits the out-of-order issues :)

ui/src/hooks/hidRpc.ts
```diff
@@ -186,6 +211,64 @@ export class KeyboardReportMessage extends RpcMessage {
     }
 }
 
+export interface KeyboardMacro extends KeysDownState {
```
IDisposable: Maybe better name `KeyboardMacroStep` ?

ui/src/hooks/useKeyboard.ts
```diff
@@ -48,69 +55,78 @@ export default function useKeyboard() {
...
+  }), []);
+
   // resetKeyboardState is used to reset the keyboard state to no keys pressed and no modifiers.
   // This is useful for macros and when the browser loses focus to ensure that the keyboard state
   // is clean.
-  const resetKeyboardState = useCallback(
-    async () => {
-      // Reset the keys buffer to zeros and the modifier state to zero
-      keysDownState.keys.length = hidKeyBufferSize;
-      keysDownState.keys.fill(0);
```
IDisposable, line 80: This change does **not** reset the running `keyDownState` keys buffer... which is important when we've had a keyboard overflow, to reset the keystate to empty. See https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R222-R234

ym: The variable name was slightly misleading. The function was only intended to send a keyboard event (modifiers: 0; keys: 0) to the controlled device.

IDisposable: Sorry about the name not being clearer... but the code in this PR [lines 86-90](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R86-R90) will **_not_** clear the `keysDownState.keys` "current record" of the currently-down keys (as the old code did). This means that if the user holds down 6 keys (the max size in `hidKeyBufferSize`, as specified in the HID spec), then on the 7th key, a real USB keyboard (and the old code), will _fill_ the `keyDownState.keys` array with `hidErrorRollOver` (0x01) on [line 231](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R231) which when sent to the host indicates that the 6-key-rollover has been exceeded. The host typically logically resets all down-keys and does nothing until it gets another HID message that _isn't_ filled with 0x01(s). 

Since this PR would **eliminate** the clearing of the `keys` buffer (e.g. setting all the 0x01 to 0x00) then we would be stuck device-side thinking the roll-over was still on-going.

We _must_ clear the `keysDownState.keys` when we encountered that case after sending the roll-over overflow to the host. [Line 120](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R120) is fine in the PR, we're not processing actual keystrokes here, but the simulation code at [lines 161-163](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R161-R163) _must_ clear the `keys` buffer either right there after we sent the HID report now, or as the side-effect of the old code.

IDisposable (changes requested):
Other than the rollover issue, this looks GOOD TO GO.

IDisposable (changes requested):
Other than the rollover issue, this looks GOOD TO GO.

ui/src/hooks/useKeyboard.ts
ym: Sorry if I've misunderstood... Let me know if this works.

https://github.com/jetkvm/kvm/blob/d6de9668bd1e2da6e2910da09ddb3eadf64948eb/jsonrpc.go#L1129

adamshiervani:
> In my test I just mashed pressed "asd". I got extra presses, plus a repeated key press. In the following logs this can be seen by either checking the sequence ID (414 comes before 413) or checking the actual events We first get (key up)
> 
> It should be the other way around. I see that we're again using goroutines. So this regressed. 

@alexballas, I think your issue here comes from mismatched versions - the UI from an older commit combined with the backend from the latest dev. That combination produces broken behavior, but it's not something that will happen in production. Let me explain.

Two things in your logs stand out: unordered events and HidRPC not being available. 

Unordered events aside, the only realistic explanation for hidRPC being unavailable is a negotiation failure. This can only occur if something is seriously wrong or the client code is too old to know about hidRPC. 

To determine which one, I did a checkout on a pre-hidRPC commit and built the UI against a newer Go backend. 

```
# Checkout pre-hidRPC commit
git checkout d952480c2aa26c7ea6fb0d4db8e88759a0d1a4e3

# Build both UI & Go
./dev_deploy.sh -r 192.168.1.202

# Checkout latest dev
git checkout dev

# Use "old" UI on new Go
./dev_deploy.sh -r 192.168.1.202 --skip-ui-build
```

Once we do that, you see the same thing as the logs you posted. ***Unordered RPC events*** & ***HID RPC not available***.

```
2025-09-10T22:54:13Z TRC jetkvm jsonrpc RPC handler returned id=6 method=keyboardReport params={"keys":[],"modifier":8} result={"keys":[0,0,0,0,0,0],"modifier":8}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc Received RPC request id=7 method=keyboardReport params={"keys":[],"modifier":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc Calling RPC handler id=7 method=keyboardReport params={"keys":[],"modifier":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc sending JSONRPC event data="{\"jsonrpc\":\"2.0\",\"method\":\"keysDownState\",\"params\":{\"modifier\":0,\"keys\":[0,0,0,0,0,0]}}"
2025-09-10T22:54:13Z WRN jetkvm jetkvm HID RPC is not available, skipping reportHidRPC
2025-09-10T22:54:13Z TRC jetkvm jsonrpc RPC handler returned id=7 method=keyboardReport params={"keys":[],"modifier":0} result={"keys":[0,0,0,0,0,0],"modifier":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc Received RPC request id=9 method=absMouseReport params={"buttons":0,"x":0,"y":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc Calling RPC handler id=9 method=absMouseReport params={"buttons":0,"x":0,"y":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc RPC handler returned id=9 method=absMouseReport params={"buttons":0,"x":0,"y":0} result=null
2025-09-10T22:54:13Z TRC jetkvm jsonrpc Received RPC request id=8 method=keyboardReport params={"keys":[],"modifier":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc Calling RPC handler id=8 method=keyboardReport params={"keys":[],"modifier":0}
2025-09-10T22:54:13Z TRC jetkvm jsonrpc RPC handler returned id=8 method=keyboardReport params={"keys":[],"modifier":0} result={"keys":[0,0,0,0,0,0],"modifier":0}
```

An alternative explanation is that you ran all the commands properly but didn't refresh the UI code between backend deploys. This also results in the same type of logs.

Regardless of the cause for the mismatch, this scenario only happens with **old UI + new backend**, which is very unlikely to occur in production:
- **New UI + old backend** -> Happens with older devices using cloud access. The UI code falls back on the "old" `keyboardReport` over JSON-RPC messages. Not ideal, nothing we can do, but also no regression.
- **New UI + new backend** -> Normal case. Uses sequential `hidRPC`, no fallback needed.
- **Old UI + new backend** -> Very unlikely, unless someone has an open cloud connection while the device upgrades. Still not a regression, and once a refresh is made, hidRPC will be used again.

What about the unordered events? When @ym added hidRPC, we moved the responsibility for sequentiality to that hidRPC data channel, and your fix was effectively reverted.

So in a sense, this is expected; JSON-RPC events are unordered because we don't need them to be ordered anymore, as we moved all the messages that need sequentiality to hidRPC. That didn't work for you as you didn't negotiate, which was most likely caused by a version mismatch.

Actually, now that I think of it, we might not even need to keep `keyboardReport` or `keypressReport` on the JSON-RPC data channel.

When hidRPC doesn't exist on the Go backend, the UI code correctly uses those methods on jsonRPC. Which makes total sense for users using the latest cloud UI but who have old devices. BUT do they need to exist on updated Go backends? Keeping them around on the go backend on new versions doesn't make much sense - and if we would've removed those, this situation wouldn't even have been able to happen.

cc @ym and @IDisposable, see a reason to keep those methods around?

ui/src/hooks/useKeyboard.ts
```diff
@@ -48,69 +60,75 @@ export default function useKeyboard() {
...
-    ],
+    [rpcDataChannel?.readyState, rpcHidReady, send, sendKeyboardEventHidRpc],
   );
 
+  const MACRO_RESET_KEYBOARD_STATE = useMemo(() => ({
+    keys: new Array(hidKeyBufferSize).fill(0),
+    modifier: 0,
+    delay: 0,
+  }), []);
+
```
adamshiervani: Don't need this in react, `hidKeyBufferSize` is an exported constant.

ui/src/components/popovers/PasteModal.tsx
```diff
@@ -171,17 +176,39 @@ export default function PasteModal() {
...
+                    label="Delay between keys"
+                    placeholder="Delay between keys"
+                    min={50}
+                    max={65534}
+                    value={delayValue}
+                    onChange={e => {
+                      setDelayValue(parseInt(e.target.value, 10));
+                    }}
+                  />
+                  {delayValue < 50 || delayValue > 65534 && (
```
adamshiervani, line 196: I would say, remove this for now and just use 50 ms. Later, we can push a cadence option in the settings with something like `Default`, `Custom` and `Human`, where Human has word segmentation and jitter in the cadence.

alexballas:
@adamshiervani  The issue only happens when using the Cloud (app.jetkvm). Connecting directly via IP, whether locally or over 5G/VPN, works fine. I’ll just wait until the final version is deployed on app.jetkvm.

adamshiervani (changes requested):

ui/src/components/popovers/PasteModal.tsx
ym: Yes, but it's also useful for troubleshooting. Manually changing the mode doesn't sound like a good option to me.

What about hiding it by default when `Troubleshooting Mode` is disabled?

adamshiervani:
@alexballas Oh, didn't even think of that one! Yeah, that's another case in the **Old UI + new backend** combination - same thing applies there too.

If you have time, feel free to build the latest version, make sure to use the device UI and see if you encounter any issues. Currently, the only known issue we have on the `dev` branch is the key release delays in bad networks. @ym will soon start work on that.

ui/src/components/popovers/PasteModal.tsx
adamshiervani: Yeah, let's do that!

IDisposable:
> When hidRPC doesn't exist on the Go backend, the UI code correctly uses those methods on jsonRPC. Which makes total sense for users using the latest cloud UI but who have old devices. BUT do they need to exist on updated Go backends? Keeping them around on the go backend on new versions doesn't make much sense - and if we would've removed those, this situation wouldn't even have been able to happen.

YES, the old device-side code in go for the old RPC-based calls **CAN** be deleted _unless_ we want (eventually?) the user to be able the _select_ which method the browser/client uses... but I can't say I'm a fan of that...

ui/src/hooks/useKeyboard.ts
IDisposable: Nope, that has _nothing_ to do with the browser/client-side state management by `simulateDeviceSideKeyHandlingForLegacyDevices`... this is for _legacy_ devices, where we handle the n-key-rollover exceeding 6 down keys... if we _on the browser/client_ don't clear this buffer and we're talking to an old device that doesn't have the new device-side code, we will get into a rollover error and never recover. 

This ONLY affects old-device + new-ui (e.g. app.jetkvm.com)

Trust this, we need to, in typescript, clear the `keyDownState.keys` array after we've reported the rollover... that's what [old line 80](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8L80) is for when called from [here](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R161-R163)

cloud.go
```diff
@@ -475,6 +475,10 @@ func handleSessionRequest(
 
 	cloudLogger.Info().Interface("session", session).Msg("new session accepted")
 	cloudLogger.Trace().Interface("session", session).Msg("new session accepted")
+
+	// Cancel any ongoing keyboard report multi when session changes
```
IDisposable: s/report multi/macro would be clearer

internal/hidrpc/message.go
```diff
@@ -84,6 +90,51 @@ func (m *Message) KeyboardReport() (KeyboardReport, error) {
 	}, nil
 }
 
+// Macro ..
+type KeyboardMacro struct {
+	Modifier byte   // 1 byte
+	Keys     []byte // 6 bytes, to make things easier, the keys length is fixed to 6
```
IDisposable: Can we please use the constant in _hid_keyboard.go_ `	hidKeyBufferSize  = 6` for these, since it really is the reason we have a 6 byte array.

internal/hidrpc/message.go
```diff
@@ -84,6 +90,51 @@ func (m *Message) KeyboardReport() (KeyboardReport, error) {
...
+	Delay    uint16 // 2 bytes
+}
+type KeyboardMacroReport struct {
+	IsPaste bool
+	Length  uint32
```
IDisposable: I think a `uint16` would be more than sufficient to handle reasonable macros (65535 steps seems pretty huge)

IDisposable: ... maybe the variable would be better named `StepCount` as it's not the message/report length...

internal/hidrpc/message.go
```diff
@@ -84,6 +90,51 @@ func (m *Message) KeyboardReport() (KeyboardReport, error) {
...
+	macro := make([]KeyboardMacro, 0, int(length))
+	for i := 0; i < int(length); i++ {
+		offset := 5 + i*9
+
+		macro = append(macro, KeyboardMacro{
+			Modifier: m.d[offset],
+			Keys:     m.d[offset+1 : offset+7],
+			Delay:    binary.BigEndian.Uint16(m.d[offset+7 : offset+9]),
+		})
+	}
```
IDisposable, line 133:
```golang
      offset += 1 + hidKeyBufferSize + 2
   }
```

internal/hidrpc/message.go
```diff
@@ -84,6 +90,51 @@ func (m *Message) KeyboardReport() (KeyboardReport, error) {
...
+
+	// check total length
+	expectedLength := int(length)*9 + 5
+	if len(m.d) != expectedLength {
+		return KeyboardMacroReport{}, fmt.Errorf("invalid length: %d, expected: %d", len(m.d), expectedLength)
+	}
+
+	macro := make([]KeyboardMacro, 0, int(length))
+	for i := 0; i < int(length); i++ {
+		offset := 5 + i*9
```
IDisposable: Can we avoid the repeated multiplication and just keep a running offset?

```golang
offset := 5
for i :=0; i < int(length); i++) {
```

jsonrpc.go
```diff
@@ -1049,6 +1052,107 @@ func rpcSetLocalLoopbackOnly(enabled bool) error {
...
+	}
+
+	return result, err
+}
+
+func rpcCancelKeyboardMacro() {
+	cancelKeyboardMacro()
+}
+
+var keyboardClearStateKeys = make([]byte, 6)
```
IDisposable: Should be referencing `hidKeyBufferSize` here.

jsonrpc.go
```diff
@@ -1049,6 +1052,107 @@ func rpcSetLocalLoopbackOnly(enabled bool) error {
...
+		case <-ctx.Done():
+			// make sure keyboard state is reset
+			_, err := rpcKeyboardReport(0, keyboardClearStateKeys)
+			if err != nil {
+				logger.Warn().Err(err).Msg("failed to reset keyboard state")
+			}
+			gadget.UpdateKeysDown(0, keyboardClearStateKeys)
+
+			logger.Debug().Int("step", i).Msg("Keyboard macro cancelled during sleep")
+			return last, ctx.Err()
```
IDisposable, line 1148: ... so when a macro is cancelled, from this method we'll return the keyDownState at the last step we executed, not the cleared state that we forced at line 1142. That value is caught in [`rpcExecuteKeyboardMacro`](https://github.com/jetkvm/kvm/pull/789/files#diff-4ce98339156c9db92c77843a1c1b6256deb9c8219a44d37a1e6d2cc5e525a5a6R1094) and [returned from that function](https://github.com/jetkvm/kvm/pull/789/files#diff-4ce98339156c9db92c77843a1c1b6256deb9c8219a44d37a1e6d2cc5e525a5a6R1103) where it is ignored at [line 40 of hidrpc.go](https://github.com/jetkvm/kvm/pull/789/files#diff-1c11b96be37effc6353457fe769a28feebc850d520bd2e609a1d9bb9ad5817daR40)... should we actually do something with that value, or just not cascade unused state back up?

ui/src/components/popovers/PasteModal.tsx
```diff
@@ -122,14 +127,18 @@ export default function PasteModal() {
...
                   animationDuration: "0.7s",
                   animationDelay: "0.1s",
                 }}
               >
                 <div>
-                  <div className="w-full" onKeyUp={e => e.stopPropagation()} onKeyDown={e => e.stopPropagation()}>
+                  <div
+                    className="w-full"
+                    onKeyUp={e => e.stopPropagation()}
+                    onKeyDown={e => e.stopPropagation()}
```
IDisposable: We really want to add these to ensure that the keypress events never travel down to the children
```
                    onKeyDownCapture={e => e.stopPropagation()}
                    onKeyUpCapture={e => e.stopPropagation()}
```

ui/src/components/popovers/PasteModal.tsx
IDisposable: Are we going to hide this before merging (with the troubleshooting mode gate)?

ui/src/components/popovers/PasteModal.tsx
```diff
@@ -200,6 +231,7 @@ export default function PasteModal() {
             size="SM"
             theme="primary"
             text="Confirm Paste"
+            disabled={isPasteModeEnabled}
```
IDisposable: Is this flag really `isCurrentlyPasting` ?

ui/src/hooks/hidRpc.ts
```diff
@@ -7,8 +7,11 @@ export const HID_RPC_MESSAGE_TYPES = {
     WheelReport: 0x04,
     KeypressReport: 0x05,
     MouseReport: 0x06,
+    KeyboardMacroReport: 0x07,
+    CancelKeyboardMacroReport: 0x08,
     KeyboardLedState: 0x32,
     KeysDownState: 0x33,
+    KeyboardMacroStateReport: 0x34,
```
IDisposable: To follow the pattern, this really aught to be `KeyboardMacroState` as the direction is device-to-client/browser (e.g. the ...`Report` messages are client-to-device and the ...`State` are are device-to-client.

ui/src/hooks/hidRpc.ts
```diff
@@ -32,6 +35,30 @@ const fromInt32toUint8 = (n: number) => {
...
 };
 
+const fromUint16toUint8 = (n: number) => {
+    if (n > 65535 || n < 0) {
+        throw new Error(`Number ${n} is not within the uint16 range`);
+    }
+
+    return new Uint8Array([
+        (n >> 8) & 0xFF,
+        (n >> 0) & 0xFF,
```
IDisposable: `(n >> 0)` is really just `n`, would it be clearer to not have the noop shift?

ui/src/hooks/hidRpc.ts
```diff
@@ -32,6 +35,30 @@ const fromInt32toUint8 = (n: number) => {
...
+const fromUint32toUint8 = (n: number) => {
+    if (n > 4294967295 || n < 0) {
+        throw new Error(`Number ${n} is not within the uint32 range`);
+    }
+
+    return new Uint8Array([
+        (n >> 24) & 0xFF,
+        (n >> 16) & 0xFF,
+        (n >> 8) & 0xFF,
+        (n >> 0) & 0xFF,
```
IDisposable: Same question as on line 45.

ui/src/hooks/hidRpc.ts
```diff
@@ -186,6 +213,99 @@ export class KeyboardReportMessage extends RpcMessage {
...
+export interface KeyboardMacroStep extends KeysDownState {
+    delay: number;
+}
+
+export class KeyboardMacroReportMessage extends RpcMessage {
+    isPaste: boolean;
+    length: number;
+    steps: KeyboardMacroStep[];
+
+    KEYS_LENGTH = 6;
```
IDisposable: This really should be `hidKeyBufferSize` from stores.ts (or we should relocated that constant) since that one dictates this one.

ui/src/hooks/hidRpc.ts
```diff
@@ -186,6 +213,99 @@ export class KeyboardReportMessage extends RpcMessage {
     }
 }
 
+export interface KeyboardMacroStep extends KeysDownState {
+    delay: number;
+}
+
+export class KeyboardMacroReportMessage extends RpcMessage {
+    isPaste: boolean;
+    length: number;
```
IDisposable: I think `stepCount` is a clearer name for this property

ui/src/hooks/hidRpc.ts
```diff
@@ -186,6 +213,99 @@ export class KeyboardReportMessage extends RpcMessage {
...
+        // validate if length is correct
+        if (this.length !== this.steps.length) {
+            throw new Error(`Length ${this.length} is not equal to the number of steps ${this.steps.length}`);
+        }
+
+        const data = new Uint8Array(this.length * 9 + 6);
+        data.set(new Uint8Array([
+            this.messageType,
+            this.isPaste ? 1 : 0,
+            ...fromUint32toUint8(this.length),
```
IDisposable: ... as suggested in the device-side go code, I doubt we really need a 32bit integer here. If we switch to a 16 bit integer, we save 2 bytes that will (essentially always) be zero.

ui/src/hooks/useHidRpc.ts
```diff
@@ -68,6 +71,21 @@ export function useHidRpc(onHidRpcMessage?: (payload: RpcMessage) => void) {
     [sendMessage],
   );
 
+  const reportKeyboardMacroEvent = useCallback(
+    (macro: KeyboardMacroStep[]) => {
+      const d = new KeyboardMacroReportMessage(false, macro.length, macro);
```
IDisposable: _nit_ all the other cases just inlined the `new ...Message` call, might as well follow the pattern?

ui/src/hooks/useKeyboard.ts
```diff
@@ -1,40 +1,52 @@
...
-  } = useHidRpc((message) => {
+  } = useHidRpc(message => {
     switch (message.constructor) {
       case KeysDownStateMessage:
         setKeysDownState((message as KeysDownStateMessage).keysDownState);
         break;
       case KeyboardLedStateMessage:
         setKeyboardLedState((message as KeyboardLedStateMessage).keyboardLedState);
         break;
+      case KeyboardMacroStateReportMessage:
```
IDisposable: ...as mentioned in the device-side go code, this should really be `KeyboardMacroStateMessage` to match the pattern

ui/src/hooks/useKeyboard.ts
IDisposable: ☝️  this still needs to be addressed. We need to clear the `keysDownState.keys` array when we've sent the reset message.

IDisposable (changes requested):
Other comments are up to you, but we _absolutely_ have to handle the keyDownState.keys clearing when we've reported the keys with a rollover error set at [line 231 of _useKeyboard.ts_](https://github.com/jetkvm/kvm/pull/789/files#diff-65766e6d595f36d9b19d614d311cb4030d166e046056ff47090673593ff727e8R231)

internal/hidrpc/message.go
ym: The maximum value for a Uint16 is `65535`. Considering that we need two or more steps for each key, the total number of usable characters would be around `25000`. I would prefer to keep a large value rather than save two bytes here. @adamshiervani any idea?

ui/src/components/popovers/PasteModal.tsx
ym: Yes, already hidden now.

internal/hidrpc/message.go
adamshiervani: Yeah, let's keep the uint32. One of the YouTube reviewers pasted in the entire manuscript of the Bee movie...

adamshiervani (approved):

adamshiervani (approved):

PR #790, opened 2025-09-10
OCR

ym:
Pure front-end implementation of OCR using tesseract.js

image

### DIFF
--- a/ui/src/components/ActionBar.tsx
+++ b/ui/src/components/ActionBar.tsx
@@ -1,4 +1,4 @@
-import { MdOutlineContentPasteGo } from "react-icons/md";
+import { MdOutlineContentPasteGo, MdOutlineDocumentScanner } from "react-icons/md";
 import { LuCable, LuHardDrive, LuMaximize, LuSettings, LuSignal } from "react-icons/lu";
 import { FaKeyboard } from "react-icons/fa6";
 import { Popover, PopoverButton, PopoverPanel } from "@headlessui/react";
@@ -19,11 +19,14 @@ import WakeOnLanModal from "@/components/popovers/WakeOnLan/Index";
 import MountPopopover from "@/components/popovers/MountPopover";
 import ExtensionPopover from "@/components/popovers/ExtensionPopover";
 import { useDeviceUiNavigation } from "@/hooks/useAppNavigation";
+import OCRModal from "./popovers/OCRModal";
 
 export default function Actionbar({
   requestFullscreen,
+  videoElmRef,
 }: {
   requestFullscreen: () => Promise<void>;
+  videoElmRef?: React.RefObject<HTMLVideoElement | null>;
 }) {
   const { navigateTo } = useDeviceUiNavigation();
   const { isVirtualKeyboardEnabled, setVirtualKeyboardEnabled } = useHidStore();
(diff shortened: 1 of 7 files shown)


ui/src/components/ActionBar.tsx
```diff
@@ -99,6 +103,36 @@ export default function Actionbar({
...
           </Popover>
+          <Popover>
+            <PopoverButton as={Fragment}>
+              <Button
+                size="XS"
+                theme="light"
+                text="OCR"
+                LeadingIcon={MdOutlineDocumentScanner}
+                onClick={() => {
+                  setDisableVideoFocusTrap(true);
```
IDisposable, line 114: Not related to this change specifically, but I wonder if we should make this trap logic be fired by the open/close of the popover itself (e.g. related to the popover being onscreen/visible, vs. hoping we caught the click).

ui/src/components/popovers/OCRModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      status: "Capturing image",
+      progress: 0,
+      jobId: "",
+      userJobId: "",
+      workerId: "",
+    });
+
+    // create a canvas from the video element then capture the image from the canvas
+    const video = videoElmRef?.current;
+    const canvas = document.createElement("canvas");
```
IDisposable, line 51: Do we need to null-check the `video` element and bail out here?

ui/src/components/popovers/OCRModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+    canvas.width = video.videoWidth;
+    canvas.height = video.videoHeight;
+
+    const ctx = canvas.getContext("2d");
+    ctx?.drawImage(video, 0, 0, canvas.width, canvas.height);
```
IDisposable, line 56: Would we want to "pause" the video before and "unpause" the video after this `drawImage` call to ensure we get a full frame instead of a (possibly partial) live view?

ui/src/components/popovers/OCRModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    const text = await ocrImage(["eng"], canvas, { logger: setOcrStatus, errorHandler: handleOcrError });
```
IDisposable, line 58: Future: do we want the user to select the "scanned for" language somehow?

ui/src/hooks/useOCR.ts
```diff
@@ -0,0 +1,25 @@
...
+// tesseract.js is h
+async function ocrImage(
+  language: string | string[],
+  image: ImageLike,
+  options?: Partial<WorkerOptions>,
+) {
+  const tesseract = await import('tesseract.js')
+  const createWorker = tesseract.createWorker || tesseract.default.createWorker
+
+  const worker = await createWorker(language, undefined, options)
```
IDisposable, line 15: I've used tesseract a bunch on dom-to-image-more's testing... I found that spinning up the worker is the slowest part as it has to load the OCR models. Perhaps we can have a lazily terminated one that stays around for a bit in case they want to do repeated OCRs?

ui/vite.config.ts
```diff
@@ -31,20 +31,30 @@ export default defineConfig(({ mode, command }) => {
...
       pure: ["console.debug"],
     },
-    build: { outDir: isCloud ? "dist" : "../static" },
+    build: {
+      outDir: isCloud ? "dist" : "../static",
+      rollupOptions: {
+        external: ["tesseract.js"],
+        output: {
+          paths: {
+            "tesseract.js": "https://cdn.jsdelivr.net/npm/tesseract.js@6.0.1/dist/tesseract.esm.min.js",
```
IDisposable, line 40: Not loving the version hardcoding here... pretty buried. I wonder if _actually adding_ the tesseract NPM module to package.json and the exracting the version number or CDN link for use here would be more manageable?

ui/src/components/popovers/OCRModal.tsx
ym: Yes, I also want it to be auto-detected based on the browser locale. Let's make the PR a draft, as there are too many improvements needed.

IDisposable: It actually looks pretty close.

PR #794, opened 2025-09-11
feat: validate ssh public key before saving

ym:


adamshiervani (approved):

IDisposable (approved):
:shipit:

adamshiervani:
@ym The test fails for some reason. Any clue?

IDisposable:
That test log is not very useful ;) ... and I guess I need to run `./dev_deploy.sh -r 192.168.1.249 --run-go-tests-only --skip-ui-build` when looking at things with a test (at least)

Looks like _ssh_test.go_ needs an update to the expected types.

```golang
func TestValidSSHKeyTypes(t *testing.T) {
	expectedTypes := []string{"ssh-rsa", "ssh-ed25519", "ecdsa-sha2-nistp256", "ecdsa-sha2-nistp384", "ecdsa-sha2-nistp521"}
```

ym:
Yea ... forgot to update the test after changing `ValidSSHKeyTypes`. It should work now.

IDisposable (approved):

PR #795, opened 2025-09-11
Don't fetch KeybardAndMouse Icon on every re-render

adamshiervani:
This doesn't solve the underlying re-rendering issue, but at least we don't fetch an icon from the device on every keypress.

@IDisposable when you've got some time over, could you spend some time looking into what's causing these re-renders?


IDisposable (approved):
I'll figure out the rerender, but this is good to go.

PR #796, opened 2025-09-11
feat: release keyPress automatically

ym:


internal/usbgadget/hid_keyboard.go
```diff
@@ -173,6 +178,64 @@ func (u *UsbGadget) SetOnKeysDownChange(f func(state KeysDownState)) {
...
+	if u.kbdAutoReleaseTimer != nil {
+		u.kbdAutoReleaseTimer.Stop()
+	}
+}
+
+func (u *UsbGadget) DelayAutoRelease() {
+	u.kbdAutoReleaseLock.Lock()
+	defer u.kbdAutoReleaseLock.Unlock()
+
+	u.log.Trace().Msg("delaying auto-release")
```
IDisposable: _nit_ we should probably emit the trace _before_ trying to grab the lock... would be easier to see deadlocks if they ever happen.

ui/src/hooks/useKeyboard.ts
```diff
@@ -111,12 +104,61 @@ export default function useKeyboard() {
     }
   };
 
+  const KEEPALIVE_INTERVAL = 75; // 200ms interval
```
IDisposable: How is `75` related to `200ms`?

IDisposable (approved):

internal/usbgadget/hid_keyboard.go
```diff
@@ -22,6 +22,11 @@ var keyboardConfig = gadgetConfigItem{
 	reportDesc: keyboardReportDesc,
 }
 
+// macOS default: 15 * 15 = 225ms https://discussions.apple.com/thread/1316947?sortBy=rank
+// Linux default: 250ms https://man.archlinux.org/man/kbdrate.8.en
+// Windows default: 1ms `HKEY_CURRENT_USER\Control Panel\Accessibility\Keyboard Response\AutoRepeatDelay`
```
adamshiervani: nit: 1s on windows

internal/usbgadget/hid_keyboard.go
```diff
@@ -398,5 +464,26 @@ func (u *UsbGadget) KeypressReport(key byte, press bool) (KeysDownState, error)
 		u.log.Warn().Uint8("modifier", modifier).Uints8("keys", keys).Msg("Could not write keypress report to hidg0")
 	}
 
+	if press {
+		{
+			u.log.Trace().Msg("acquiring kbdAutoReleaseLock to update last key")
+			u.kbdAutoReleaseLock.Lock()
+			u.kbdAutoReleaseLastKey = key
```
adamshiervani: Not needed anymore as you reset the entire keyboard state on autorelease

internal/usbgadget/usbgadget.go
```diff
@@ -68,6 +68,10 @@ type UsbGadget struct {
 	keyboardState byte          // keyboard latched state (NumLock, CapsLock, ScrollLock, Compose, Kana)
 	keysDownState KeysDownState // keyboard dynamic state (modifier keys and pressed keys)
 
+	kbdAutoReleaseLock    sync.Mutex
+	kbdAutoReleaseTimer   *time.Timer
+	kbdAutoReleaseLastKey byte
```
adamshiervani: Same

ui/src/hooks/useKeyboard.ts
```diff
@@ -111,12 +104,61 @@ export default function useKeyboard() {
...
+      keepAliveTimerRef.current = null;
+    }
+  }, []);
+
+  const scheduleKeepAlive = useCallback(() => {
+    // Clear existing timer if it exists
+    if (keepAliveTimerRef.current) {
+      clearInterval(keepAliveTimerRef.current);
+    }
+
```
adamshiervani, line 130: Call `sendKeypressKeepAliveHidRpc()` immediately and don't wait for the first `setInterval`.

webrtc.go
```diff
@@ -157,40 +205,12 @@ func newSession(config SessionConfig) (*Session, error) {
...
-				queue := session.hidQueue[queueIndex]
-				if queue != nil {
-					queue <- msg
-				} else {
-					l.Warn().Int("queueIndex", queueIndex).Msg("received data in HID RPC message handler, but queue is nil")
-					return
-				}
-			})
+			d.OnMessage(getOnHidMessageHandler(session, scopedLogger, "hidrpc"))
+		// we won't send anything over the unreliable channels
```
adamshiervani, line 250: Could you clarify this a bit more?

adamshiervani (approved):

internal/usbgadget/hid_keyboard.go
```diff
@@ -215,9 +219,12 @@ func (u *UsbGadget) DelayAutoRelease() {
 	u.log.Trace().Msg("auto-release timer reset")
 }
 
-func (u *UsbGadget) performAutoRelease(key byte) {
+func (u *UsbGadget) performAutoRelease() {
+	u.log.Trace().Msg("performing autoRelease")
 	u.kbdAutoReleaseLock.Lock()
-	defer u.kbdAutoReleaseLock.Unlock()
+	defer unlockWithLog(&u.kbdAutoReleaseLock, u.log, "autoRelease unlocked")
```
IDisposable: s/unlocked/released

internal/usbgadget/hid_keyboard.go
```diff
@@ -226,7 +233,8 @@ func (u *UsbGadget) performAutoRelease(key byte) {
 	}
 
 	// we just reset the keyboard state to 0 no matter what
-	_, err := u.keypressReport(0, false, false)
+	u.log.Trace().Uint8("key", key).Msg("auto-releasing keyboard key")
+	_, err := u.keypressReport(key, false, false)
```
IDisposable: 👍

internal/usbgadget/hid_keyboard.go
```diff
@@ -445,28 +458,37 @@ func (u *UsbGadget) keypressReport(key byte, press bool, autoRelease bool) (Keys
...
 			} else {
 				// If we are releasing a key, and we didn't find it in a slot, who cares?
-				u.log.Warn().Uint8("key", key).Msg("key not found in buffer, nothing to release")
+				l.Warn().Msg("key not found in buffer, nothing to release")
 			}
 		}
 	}
 
+	if l.GetLevel() <= zerolog.DebugLevel {
+		l = l.With().Uint8("modifier", modifier).Uints8("keys", keys).Logger()
```
IDisposable: Should this be `.Bytes("keys", keys)` so we get the more readable logging?

internal/usbgadget/hid_keyboard.go
IDisposable: > Not needed anymore as you reset the entire keyboard state on autorelease

Seem it is now _correctly_ is [now using this value](https://github.com/jetkvm/kvm/pull/796/files/cbf1c7fba68dc9c4a0a083537444bc85d3a28020..12210fac96ef7b9ddc4d0bdf4c7e476e7455505a#diff-ca2c65f26d876c5c490651ad24cf7b6cbc9a996b1d969d2d9658c7d439351a8cR226-R237)

internal/usbgadget/hid_keyboard.go
```diff
@@ -481,6 +503,31 @@ func (u *UsbGadget) keypressReport(key byte, press bool, autoRelease bool) (Keys
...
+type keypressReportResult struct {
+	KeysDownState KeysDownState
+	Error         error
+}
+
+func (u *UsbGadget) keypressReport(key byte, press bool, autoRelease bool) (KeysDownState, error) {
+	u.keyboardLock.Lock()
+	defer u.keyboardLock.Unlock()
+
+	r := make(chan keypressReportResult)
```
IDisposable: Seems like a lot of setup and tear-down to execute every single keypress... can we preconstruct the channel and go routine during setup and just use it here?

internal/usbgadget/utils.go
```diff
@@ -164,3 +171,8 @@ func (u *UsbGadget) resetLogSuppressionCounter(counterName string) {
 		u.logSuppressionCounter[counterName] = 0
 	}
 }
+
+func unlockWithLog(lock *sync.Mutex, logger *zerolog.Logger, msg string, args ...any) {
+	logger.Trace().Msgf(msg, args...)
+	lock.Unlock()
```
IDisposable, line 177: If this throws during the trace (unlikely?) then we would not release the lock. I suggest the normal pattern
```golang
	defer lock.Unlock()
    logger.Trace().Msgf(msg, args...)
```

ui/src/hooks/useHidRpc.ts
```diff
@@ -167,10 +167,24 @@ export function useHidRpc(onHidRpcMessage?: (payload: RpcMessage) => void) {
       onHidRpcMessage?.(message);
     };
 
+    const openHandler = () => {
+      console.warn("HID RPC channel opened");
```
IDisposable: Warn? Seems like this is more Trace/Info not an error state...

ui/src/hooks/useHidRpc.ts
```diff
@@ -167,10 +167,24 @@ export function useHidRpc(onHidRpcMessage?: (payload: RpcMessage) => void) {
...
+      sendHandshake();
+    };
+
+    const closeHandler = () => {
+      console.warn("HID RPC channel closed");
```
IDisposable: Warn? Seems like this is more Trace/Info not an error state...

adamshiervani:
I've now fixed all the issues that I found while testing. There were a bunch of places where we got stuck, mainly caused by bad network or overloaded hidWrites.

I've just pushed all the fixes, and mark this to **Ready for Review**. In addition to a normal review, I suggest checking the code out and following the hidRPC code path - good to understand, but also try to find issues.

My testing was done by creating different combinations of `handleKeyPress(key, press)` in JS: 
```js
// Prints abc...xyz 3 times
for (let i = 0; i < 3; i++) {
  for (let i = 0; i < 26; i++) {
    handleKeyPress(keys[`Key${String.fromCharCode(65 + i)}`], true);
    await new Promise(resolve => setTimeout(resolve, 10));
    handleKeyPress(keys[`Key${String.fromCharCode(65 + i)}`], false);
    await new Promise(resolve => setTimeout(resolve, 10));
  }
}
```

Other than that, just smashing your keyboard on a bad network can be a good way to ensure the auto-release is doing what it should.


My testing was on Ubuntu 22.04, and I used [Network Link Conditioner](https://nshipster.com/network-link-conditioner/) to simulate both packet loss and delay.

@IDisposable @ym

internal/usbgadget/hid_keyboard.go
```diff
@@ -22,6 +26,11 @@ var keyboardConfig = gadgetConfigItem{
 	reportDesc: keyboardReportDesc,
 }
 
+// macOS default: 15 * 15 = 225ms https://discussions.apple.com/thread/1316947?sortBy=rank
+// Linux default: 250ms https://man.archlinux.org/man/kbdrate.8.en
+// Windows default: 1s `HKEY_CURRENT_USER\Control Panel\Accessibility\Keyboard Response\AutoRepeatDelay`
+const autoReleaseKeyboardInterval = time.Millisecond * 100
```
ym: Should we change it back around `225ms` ?

hidrpc.go
```diff
@@ -85,27 +91,88 @@ func onHidMessage(data []byte, session *Session) {
...
+				// This is normal jitter (e.g., Wi-Fi contention).
+				// We still accept the tick, but *reduce the extension*
+				// so that the total hold time stays aligned with REAL client side intent.
+				timerExtension -= lateness
+			} else {
+				// --- Large lateness (beyond jitterBudget) ---
+				// This is likely a retransmit stall or ordering delay.
+				// We reject the tick entirely and DO NOT extend,
+				// so the auto-release still fires on time.
+				validTick = false
```
IDisposable, line 150: this can just be `return nil` since we're going to bail out at line 143 anyway (and thus that block and the flag can be deleted)

internal/usbgadget/hid_keyboard.go
```diff
@@ -145,32 +149,105 @@ func (u *UsbGadget) GetKeysDownState() KeysDownState {
...
+}
+
+func (u *UsbGadget) DelayAutoReleaseWithDuration(resetDuration time.Duration) {
+	u.kbdAutoReleaseLock.Lock()
+	defer unlockWithLog(&u.kbdAutoReleaseLock, u.log, "autoRelease delayed")
 
-		u.log.Trace().Interface("old", u.keysDownState).Interface("new", state).Msg("keysDownState updated")
-		u.keysDownState = state
+	if u.kbdAutoReleaseTimers == nil {
+		return
```
IDisposable: I can't see any path where this would be nil (the entire collection, not the individual timer). I don't think this check is needed.

internal/usbgadget/hid_keyboard.go
```diff
@@ -145,32 +149,105 @@ func (u *UsbGadget) GetKeysDownState() KeysDownState {
...
+		if timer != nil {
+			timer.Reset(resetDuration)
+		}
 	}
 }
 
-func (u *UsbGadget) SetOnKeysDownChange(f func(state KeysDownState)) {
-	u.onKeysDownChange = &f
+func (u *UsbGadget) performAutoRelease(key byte) {
+	u.kbdAutoReleaseLock.Lock()
```
IDisposable, line 209: This block could be better structured as a typical defer block if we surround with a scope block, thus:

```golang
// lock scope
{ 
    u.kbdAutoReleaseLock.Lock()
	defer u.kbdAutoReleaseLock.Unlock()

	if u.kbdAutoReleaseTimers[key] == nil {
		u.log.Warn().Uint8("key", key).Msg("autoRelease timer not found")
		return
	}

	u.kbdAutoReleaseTimers[key].Stop()
	u.kbdAutoReleaseTimers[key] = nil
	delete(u.kbdAutoReleaseTimers, key
} // end lock scope
```

internal/usbgadget/hid_keyboard.go
```diff
@@ -145,32 +149,105 @@ func (u *UsbGadget) GetKeysDownState() KeysDownState {
...
+		return
+	}
+
+	u.kbdAutoReleaseTimers[key].Stop()
+	u.kbdAutoReleaseTimers[key] = nil
+	delete(u.kbdAutoReleaseTimers, key)
+	u.kbdAutoReleaseLock.Unlock()
+
+	// Skip if already released
+	state := u.GetKeysDownState()
```
IDisposable, line 223: Should this have been done inside the lock scope?

internal/usbgadget/usbgadget.go
```diff
@@ -149,3 +156,35 @@ func newUsbGadget(name string, configMap map[string]gadgetConfigItem, enabledDev
...
+	// Cancel keyboard state context
+	if u.keyboardStateCancel != nil {
+		u.keyboardStateCancel()
+	}
+
+	// Stop auto-release timer
+	u.kbdAutoReleaseLock.Lock()
+	for _, timer := range u.kbdAutoReleaseTimers {
+		timer.Stop()
+	}
```
IDisposable, line 171: In other paths, we explicitly `delete` the timers as we go... doing this way will not leak anything?

ui/src/components/WebRTCVideo.tsx
```diff
@@ -237,6 +237,7 @@ export default function WebRTCVideo() {
   const keyDownHandler = useCallback(
     (e: KeyboardEvent) => {
       e.preventDefault();
+      if (e.repeat) return;
```
IDisposable, line 240: This means that we'll not reset the auto-key-release timers on the device while we _KNOW_ that the key is still being held down. That seems incorrect, since the only way we will be getting these `e.repeat == true` client-side events is if the key is still down. 

Wouldn't it be better to allow this to still be communicated in some way to the device so the auto-release doesn't fire?

ui/src/hooks/useHidRpc.ts
```diff
@@ -6,45 +6,101 @@ import {
...
+
   const rpcHidReady = useMemo(() => {
     return rpcHidChannel?.readyState === "open" && rpcHidProtocolVersion !== null;
   }, [rpcHidChannel, rpcHidProtocolVersion]);
 
+  const rpcHidUnreliableReady = useMemo(() => {
+    return (
+      rpcHidUnreliableChannel?.readyState === "open" && rpcHidProtocolVersion !== null
+    );
+  }, [rpcHidUnreliableChannel, rpcHidProtocolVersion]);
```
IDisposable: Would this be a better dependency as `[rpcHidUnreliableChannel?.ready, rpcHidProtocolVersion]` (which is what VSCode would have inserted)?

ui/src/hooks/useHidRpc.ts
```diff
@@ -6,45 +6,101 @@ import {
...
+
+  const rpcHidUnreliableNonOrderedReady = useMemo(() => {
+    return (
+      rpcHidUnreliableNonOrderedChannel?.readyState === "open" &&
+      rpcHidProtocolVersion !== null
+    );
+  }, [rpcHidUnreliableNonOrderedChannel, rpcHidProtocolVersion]);
```
IDisposable: same question as line 42

ui/src/hooks/useHidRpc.ts
```diff
@@ -68,29 +126,36 @@ export function useHidRpc(onHidRpcMessage?: (payload: RpcMessage) => void) {
...
-      // so we can't support this
-      // TODO: use capabilities to determine rather than version number
-      console.error("Server is using a newer HID RPC version than the client", message);
-      return;
-    }
+      if (message.version > HID_RPC_VERSION) {
+        // we assume that the UI is always using the latest version of the HID RPC protocol
+        // so we can't support this
+        // TODO: use capabilities to determine rather than version number
+        console.error("Server is using a newer HID RPC version than the client", message);
```
IDisposable, line 175: We could trigger a forced reload of the client-side page here with `window.location.reload();` but if that results in the client still being behind the server... something is wonky.

ui/src/hooks/useKeyboard.ts
```diff
@@ -111,12 +114,60 @@ export default function useKeyboard() {
...
+    if (keepAliveTimerRef.current) {
+      clearInterval(keepAliveTimerRef.current);
+      keepAliveTimerRef.current = null;
+    }
+  }, []);
+
+  const scheduleKeepAlive = useCallback(() => {
+    // Clear existing timer if it exists
+    if (keepAliveTimerRef.current) {
+      clearInterval(keepAliveTimerRef.current);
```
IDisposable: should we also `delete keepAliveTimerRef.current;` to ensure there's not a reference cycle?

ui/src/hooks/useKeyboard.ts
```diff
@@ -111,12 +114,60 @@ export default function useKeyboard() {
...
+    }
+
+    keepAliveTimerRef.current = setInterval(() => {
+      sendKeypressKeepAliveHidRpc();
+    }, KEEPALIVE_INTERVAL);
+  }, [sendKeypressKeepAliveHidRpc]);
+
+  // resetKeyboardState is used to reset the keyboard state to no keys pressed and no modifiers.
+  // This is useful for macros and when the browser loses focus to ensure that the keyboard state
```
IDisposable: `for macros, in case of client-side rollover,  and when the browser loses focus`

webrtc.go
```diff
@@ -78,16 +97,85 @@ func (s *Session) initQueues() {
...
+		onHidMessage(msg, s)
+	}
+}
+
+const keysDownStateQueueSize = 256
+
+func (s *Session) initKeysDownStateQueue() {
+	// serialise outbound key state reports so unreliable links can't stall input handling
+	s.keysDownStateQueue = make(chan usbgadget.KeysDownState, keysDownStateQueueSize)
+	go s.handleKeysDownStateQueue()
```
IDisposable, line 118: Do we need a cancel flag to exit the queue's go-routine loop?

webrtc.go
```diff
@@ -78,16 +97,85 @@ func (s *Session) initQueues() {
...
 }
 
 func (s *Session) handleQueues(index int) {
 	for msg := range s.hidQueue[index] {
-		onHidMessage(msg.Data, s)
+		onHidMessage(msg, s)
+	}
+}
+
+const keysDownStateQueueSize = 256
```
IDisposable: This seems _really deep_ :)

webrtc.go
```diff
@@ -157,40 +246,12 @@ func newSession(config SessionConfig) (*Session, error) {
...
-
-				queue := session.hidQueue[queueIndex]
-				if queue != nil {
-					queue <- msg
-				} else {
-					l.Warn().Int("queueIndex", queueIndex).Msg("received data in HID RPC message handler, but queue is nil")
-					return
-				}
-			})
+			d.OnMessage(getOnHidMessageHandler(session, scopedLogger, "hidrpc"))
```
IDisposable, line 249: Since the label is part of the RPC message, would it be better to use shorter strings, especially for the newly introduced ones (since we don't have legacy support concerns), perhaps "hidr", "hiduo", and "hidun" or even "hr", "ho", and "hn" since the label is really meaningless outside this code.

IDisposable (changes requested):
There's a couple suggestions, and a couple questions... feel free to disregard.

webrtc.go
adamshiervani: 😂

hidrpc.go
adamshiervani: I totally understand your comment, but I'd like to keep it extremely verbose, as this part can be difficult to grasp, and for me that's just a tad easier to follow.

internal/usbgadget/usbgadget.go
adamshiervani: Checking for null. Thanks!

internal/usbgadget/hid_keyboard.go
adamshiervani: Neither can I

ui/src/hooks/useKeyboard.ts
adamshiervani: Calling `cancelKeepAlive()` instead.

internal/usbgadget/hid_keyboard.go
adamshiervani: Its lock is inside the call.

webrtc.go
adamshiervani: It's also used for logs/debugging, and in general I really prefer senseful verbosity over terseness.

ui/src/components/WebRTCVideo.tsx
adamshiervani: We don’t need repeats - those only add complexity - just much easier to reason with fewer events. The backend already maintains the key state using KeypressKeepAliveReport.

For devices that support the keypress hidRPC, each key-down starts a setInterval that sends a KeypressKeepAliveReport every 50 ms. The corresponding key-up clears that interval and stops sending keep alives.

Legacy devices (no hidRPC) run in stateless mode: each JSON-RPC KeyboardReport fully specifies which keys are down, and the device holds that state until a new report arrives - no repeats or keep-alives required.

ui/src/hooks/useHidRpc.ts
adamshiervani: Let's just keep it as is and think it through properly when we tackle the new/old UI and new/old backend task.

internal/usbgadget/hid_keyboard.go
adamshiervani: It was something along those lines in a previous commit, but it was causing some issues. That’s when I realized defer isn’t block-scoped, only function-scoped.

webrtc.go
adamshiervani: No need up here, BUT I realized that we weren't closing the channel on connection close, which I added.

adamshiervani (approved):

internal/usbgadget/hid_keyboard.go
IDisposable: Oh, I was misinformed, so sorry... I thought it was block scoped in go. Nice to learn

PR #799, opened 2025-09-11
update netboot.xyz-multiarch.iso to 2.0.88

ym:


scripts/update_netboot_xyz.sh
```diff
@@ -0,0 +1,77 @@
...
+msg_info "Downloading ISO file ..."
+curl -L -o "$TMP_ISO" "$ISO_URL"
+
+# Verify SHA256 checksum
+msg_info "Verifying SHA256 checksum ..."
+ACTUAL_CHECKSUM=$(shasum -a 256 "$TMP_ISO" | awk '{print $1}')
+
+if [ "$EXPECTED_CHECKSUM" = "$ACTUAL_CHECKSUM" ]; then
+    msg_ok "Verified SHA256 checksum."
+    mv "$TMP_ISO" "resource/netboot.xyz-multiarch.iso"
```
IDisposable: might want `mv -f ` to be more explicit about overwriting the existing iso (and also makes sure we don't get tripped up by the iso file being flagged read-only)

IDisposable (approved):
Looks great... just one _nit suggestion_ that you're welcome to ignore :)

PR #801, opened 2025-09-12
fix: goroutine leak issue of cloudBlink

### LINKED ISSUE #798: Question about the codebase. Should we be returning here?
https://github.com/jetkvm/kvm/blame/ea068414dccf8d52400164dd1ea8c10d305756ad/display.go#L166

Should we be returning here? It feels that if we run `startCloudBlink` and `cloudBlinkStopped` is false we're just going to spawn more goroutines.

@ym

ym:
fixes https://github.com/jetkvm/kvm/issues/798


IDisposable (approved):
Looks good, :shipit:

PR #803, opened 2025-09-12
feat: Adds IPv6 disabling feature

IDisposable:
Adds the ability to disable IPv6 support.

- Updates mDNS and web server to not listen on disabled addresses.
- Updates network interface state to allow disabling IPv6.

### DIFF
--- a/internal/network/netif.go
+++ b/internal/network/netif.go
@@ -239,6 +239,10 @@ func (s *NetworkInterfaceState) update() (DhcpTargetState, error) {
 			ipv4Addresses = append(ipv4Addresses, addr.IP)
 			ipv4AddressesString = append(ipv4AddressesString, addr.IPNet.String())
 		} else if addr.IP.To16() != nil {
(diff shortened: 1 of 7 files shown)


internal/network/netif.go
```diff
@@ -287,35 +291,37 @@ func (s *NetworkInterfaceState) update() (DhcpTargetState, error) {
...
-				scopedLogger.Info().
-					Str("old_ipv6", s.ipv6Addr.String()).
-					Msg("IPv6 address changed")
-			} else {
-				scopedLogger.Info().Msg("IPv6 address found")
+		s.ipv6Addresses = ipv6Addresses
+
+		if len(ipv6Addresses) > 0 {
+			// compare the addresses to see if there's a change
+			if s.ipv6Addr == nil || s.ipv6Addr.String() != ipv6Addresses[0].Address.String() {
```
IDisposable, line 313: This doesn't really check _all_ the addresses, but the old code didn't either.

ym (approved):

PR #815, opened 2025-09-20
feat(macros): add text/wait steps and added export and import

SilkePilon:
- Lower minimum step delay to 10ms for finer-grained macro timing
- Add optional "text" field on macro steps and client-side expansion into per-character key events (handles shift, AltRight, dead/accent keys, trailing space)
- Add optional "wait" field on macro steps to represent explicit wait-only pauses (treated as no-op delays)
- Wire text expansion into both remote and client-side macro execution
- Ensure typed keys send press then explicit release delay
- Update Go/TypeScript types and JSON-RPC parsing for new step fields
- added the option to export and import macros

Demo:
https://www.youtube.com/watch?v=cD0neh8rioA

### DIFF
--- a/config.go
+++ b/config.go
@@ -23,7 +23,7 @@ const (
 	MaxMacrosPerDevice = 25
 	MaxStepsPerMacro   = 10
 	MaxKeysPerStep     = 10
-	MinStepDelay       = 50
+	MinStepDelay       = 10
 	MaxStepDelay       = 2000
 )
 
@@ -31,6 +31,10 @@ type KeyboardMacroStep struct {
 	Keys      []string `json:"keys"`
 	Modifiers []string `json:"modifiers"`
 	Delay     int      `json:"delay"`
+	// Optional: when set, this step types the given text using the configured keyboard layout.
+	// The delay value is treated as the per-character delay.
+	Text      string   `json:"text,omitempty"`
+	Wait      bool     `json:"wait,omitempty"`
 }
 
 func (s *KeyboardMacroStep) Validate() error {
--- a/ui/src/hooks/useKeyboard.ts
+++ b/ui/src/hooks/useKeyboard.ts
@@ -16,6 +16,7 @@ import {
 import { useHidRpc } from "@/hooks/useHidRpc";
 import { JsonRpcResponse, useJsonRpc } from "@/hooks/useJsonRpc";
 import { hidKeyToModifierMask, keys, modifiers } from "@/keyboardMappings";
+import useKeyboardLayout from "@/hooks/useKeyboardLayout";
 
 const MACRO_RESET_KEYBOARD_STATE = {
   keys: new Array(hidKeyBufferSize).fill(0),
@@ -27,6 +28,8 @@ export interface MacroStep {
   keys: string[] | null;
   modifiers: string[] | null;
   delay: number;
+  text?: string | undefined;
+  wait?: boolean | undefined;
 }
 
 export type MacroSteps = MacroStep[];
@@ -34,6 +37,7 @@ export type MacroSteps = MacroStep[];
 const sleep = (ms: number): Promise<void> => new Promise(resolve => setTimeout(resolve, ms));
 
 export default function useKeyboard() {
+  const { selectedKeyboard } = useKeyboardLayout();
   const { send } = useJsonRpc();
   const { rpcDataChannel } = useRTCStore();
   const { keysDownState, setKeysDownState, setKeyboardLedState, setPasteModeEnabled } =
@@ -284,9 +288,38 @@ export default function useKeyboard() {
   // After the delay, the keys and modifiers are released and the next step is executed.
   // If a step has no keys or modifiers, it is treated as a delay-only step.
   // A small pause is added between steps to ensure that the device can process the events.
+  const expandTextSteps = useCallback((steps: MacroSteps): MacroSteps => {
+    const expanded: MacroSteps = [];
+    for (const step of steps) {
+      if (step.text && step.text.length > 0 && selectedKeyboard) {
+        for (const char of step.text) {
+          const keyprops = selectedKeyboard.chars[char];
+          if (!keyprops) continue;
+          const { key, shift, altRight, deadKey, accentKey } = keyprops;
+          if (!key) continue;
+          if (accentKey) {
+            const accentModifiers: string[] = [];
+            if (accentKey.shift) accentModifiers.push("ShiftLeft");
+            if (accentKey.altRight) accentModifiers.push("AltRight");
+            expanded.push({ keys: [String(accentKey.key)], modifiers: accentModifiers, delay: step.delay });
+          }
+          const mods: string[] = [];
+          if (shift) mods.push("ShiftLeft");
+          if (altRight) mods.push("AltRight");
+          expanded.push({ keys: [String(key)], modifiers: mods, delay: step.delay });
+          if (deadKey) expanded.push({ keys: ["Space"], modifiers: null, delay: step.delay });
+        }
+      } else {
+        expanded.push(step);
+      }
+    }
+    return expanded;
+  }, [selectedKeyboard]);
+
   const executeMacroRemote = useCallback(async (
-    steps: MacroSteps,
+    stepsIn: MacroSteps,
   ) => {
+    const steps = expandTextSteps(stepsIn);
     const macro: KeyboardMacroStep[] = [];
 
     for (const [_, step] of steps.entries()) {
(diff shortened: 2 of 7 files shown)


config.go
```diff
@@ -23,14 +23,18 @@ const (
...
-	MinStepDelay       = 50
+	MinStepDelay       = 10
 	MaxStepDelay       = 2000
 )
 
 type KeyboardMacroStep struct {
 	Keys      []string `json:"keys"`
 	Modifiers []string `json:"modifiers"`
 	Delay     int      `json:"delay"`
+	// Optional: when set, this step types the given text using the configured keyboard layout.
```
IDisposable, line 34: Clarification, is it going to "press the named key" or is it "press the correct key for the mentioned character, based on selected keyboard layout"?

config.go
```diff
@@ -23,14 +23,18 @@ const (
...
+	// The delay value is treated as the per-character delay.
+	Text      string   `json:"text,omitempty"`
+	Wait      bool     `json:"wait,omitempty"`
```
IDisposable, line 37: Can't we already do that with no keys and delay?

config.go
SilkePilon: It’s layout-aware: we expand each character in the text into the proper key + modifier (and accent/dead-key where needed) events for the selected layout before sending.

config.go
SilkePilon: The "Wait" makes a pause explicit, always clears state first, and gives us a future hook; an empty keys+delay is ambiguous and less robust.

IDisposable: I'll defer to @adamshiervani or @ym on this... I don't think empty-keys+delay is any less robust... but if there's value in wait's implementation, the by all means. Otherwise this seems to be pretty solid

ui/src/hooks/useKeyboard.ts
```diff
@@ -284,9 +288,38 @@ export default function useKeyboard() {
...
   // If a step has no keys or modifiers, it is treated as a delay-only step.
   // A small pause is added between steps to ensure that the device can process the events.
+  const expandTextSteps = useCallback((steps: MacroSteps): MacroSteps => {
+    const expanded: MacroSteps = [];
+    for (const step of steps) {
+      if (step.text && step.text.length > 0 && selectedKeyboard) {
+        for (const char of step.text) {
+          const keyprops = selectedKeyboard.chars[char];
+          if (!keyprops) continue;
+          const { key, shift, altRight, deadKey, accentKey } = keyprops;
```
IDisposable, line 298: This code is very similar to what is done in _VirtualKeyboard.tsx_. Is there any way we could refactor both or put a comment in here to note that they should be kept in sync.

ui/src/hooks/useKeyboard.ts
```diff
@@ -297,16 +330,22 @@ export default function useKeyboard() {
 
         .reduce((acc, val) => acc + val, 0);
 
-      // If the step has keys and/or modifiers, press them and hold for the delay
-      if (keyValues.length > 0 || modifierMask > 0) {
+      if (step.wait) {
+        // pure wait: send a no-op clear state with desired delay
+        macro.push({ ...MACRO_RESET_KEYBOARD_STATE, delay: step.delay || 100 });
```
IDisposable, line 335: Can we use DEFAULT_DELAY here (and line 338 and line 341) for the delay fallback? (it's 50ms, which really shouldn't be a problem)

ui/src/routes/devices.$id.settings.macros.tsx
```diff
@@ -22,20 +23,40 @@ import { ConfirmDialog } from "@/components/ConfirmDialog";
 import LoadingSpinner from "@/components/LoadingSpinner";
 import useKeyboardLayout from "@/hooks/useKeyboardLayout";
 
-const normalizeSortOrders = (macros: KeySequence[]): KeySequence[] => {
-  return macros.map((macro, index) => ({
-    ...macro,
-    sortOrder: index + 1,
-  }));
+const normalizeSortOrders = (macros: KeySequence[]): KeySequence[] => macros.map((m, i) => ({ ...m, sortOrder: i + 1 }));
```
IDisposable, line 26: Can we leave the original `normalizeSortOrders` here, this change is less readable and seems identical in behavior

ui/src/hooks/useKeyboard.ts
```diff
@@ -318,11 +357,14 @@ export default function useKeyboard() {
         .map(mod => modifiers[mod])
         .reduce((acc, val) => acc + val, 0);
 
-      // If the step has keys and/or modifiers, press them and hold for the delay
-      if (keyValues.length > 0 || modifierMask > 0) {
+      if (step.wait) {
+        promises.push(() => sleep(step.delay || 100));
```
IDisposable, line 361: Can we use DEFAULT_DELAY here (and line 365 and line 367) for the delay fallback? (it's 50ms, which really shouldn't be a problem)

ui/src/hooks/useKeyboard.ts
```diff
@@ -284,9 +288,38 @@ export default function useKeyboard() {
   // After the delay, the keys and modifiers are released and the next step is executed.
   // If a step has no keys or modifiers, it is treated as a delay-only step.
   // A small pause is added between steps to ensure that the device can process the events.
+  const expandTextSteps = useCallback((steps: MacroSteps): MacroSteps => {
```
IDisposable, line 291: This could be coded as a generator function so-as to not need to reallocate/copy the entire array.
 
 ```typescript
// Generator function to process steps, expanding text type ones into keystrokes, yielding steps
function* expandTextSteps(steps: Iterable<MacroStep>, selectedKeyboard: KeyboardLayout): Generator<MacroStep> {
   const stepIterator = steps[Symbol.iterator]();
   let currentStep = stepIterator.next(); 

   while (!currentStep.done) {
      const step = currentStep.value;

      if (step.text && step.text.length > 0) {
         for (const char of step.text) {
            const keyprops = selectedKeyboard.chars[char];
            if (!keyprops) continue;
            const { key, shift, altRight, deadKey, accentKey } = keyprops;

            if (!key) continue;

            if (accentKey) {
               const accentModifiers: string[] = [];
               if (accentKey.shift) accentModifiers.push("ShiftLeft");
               if (accentKey.altRight) accentModifiers.push("AltRight");
               accentStep: MacroStep = { keys: [String(accentKey.key)], modifiers: accentModifiers, delay: step.delay };
               yield accentStep;
          }

          const mods: string[] = [];
          if (shift) mods.push("ShiftLeft");
          if (altRight) mods.push("AltRight");

          characterStep: MacroStep = { keys: [String(key)], modifiers: mods, delay: step.delay };
          yield characterStep;

          if (deadKey) {
             deadStep: MacroStep = { keys: ["Space"], modifiers: null, delay: step.delay };
             yield deadStep;
          }
      } else {
         yield step;
      }

      currentStep = stepIterator.next();
}
```

_the above was composed in GitHub review editor and easily could be wrong, but it's close :)_

File: ui/src/routes/devices.$id.settings.macros.tsx [hunk 8]
```diff
@@ -312,18 +314,57 @@ export default function SettingsMacrosRoute() {
...
+                  const parsed = JSON.parse(raw);
+                  const candidates = Array.isArray(parsed) ? parsed : [parsed];
+                  for (const c of candidates) {
+                    if (working.length >= MAX_TOTAL_MACROS) { skipped += (candidates.length - candidates.indexOf(c)); break; }
+                    if (!c || typeof c !== "object") { errors++; continue; }
+                    const sanitized = sanitizeImportedMacro(c, working.length + 1);
+                    working.push(sanitized);
+                    imported.push(sanitized.name);
+                  }
+                } catch { errors++; }
```
IDisposable, line 347: Might be nice to capture the error and what the raw line was for an error message modal?

ui/src/routes/devices.$id.settings.macros.tsx
```diff
@@ -312,18 +314,57 @@ export default function SettingsMacrosRoute() {
...
+            size="SM"
+            theme="primary"
+            text={isMaxMacrosReached ? `Max Reached` : "Add New Macro"}
+            onClick={() => navigate("add")}
+            disabled={isMaxMacrosReached}
+            aria-label="Add new macro"
+          />
+          <div className="ml-2 flex items-center gap-2">
+            <input ref={fileInputRef} type="file" accept="application/json" multiple className="hidden" onChange={async e => {
+              const fl = e.target.files;
```
IDisposable, line 328: Can we extract this code out to a function `onFileUpload(files)`?

IDisposable (changes requested):

PR #822, opened 2025-09-24
Add pointer even when HDMI isn't feeding video

tomwopat:
My use case for my JetKVM is to control another computer on my desktop that has its own display, so I have no need for the video display. However, I do want to be able to enable Pointer Lock and Keyboard Lock.

This PR adds two UI features to provide this:

* It allows enabling Pointer Lock by clicking anywhere where the video display would be
* It adds a Fullscreen button in the HDMI signal message. There is not currently a way with any browser to enable Keyboard Lock without entering fullscreen.

I accept any input on the UI, as I'm not a frontend dev. I had also implemented Pointer Lock and Fullscreen buttons in the Actionbar, but it's already a bit crowded. It might also be worth hiding this functionality behind an options toggle, but it mirrors similar behavior on PiKVM.

### DIFF
--- a/ui/src/components/WebRTCVideo.tsx
+++ b/ui/src/components/WebRTCVideo.tsx
@@ -542,7 +542,7 @@ export default function WebRTCVideo() {
                           style={{ animationDuration: "500ms" }}
                           className="animate-slideUpFade pointer-events-none absolute inset-0 flex items-center justify-center"
                         >
-                          <div className="relative h-full w-full rounded-md">
+                          <div className="relative h-full w-full rounded-md" onClick={requestPointerLock}>
                             <LoadingVideoOverlay show={isVideoLoading} />
                             <HDMIErrorOverlay show={hdmiError} hdmiState={hdmiState} />
                             <NoAutoplayPermissionsOverlay
--- a/ui/src/components/WebRTCVideo.tsx
+++ b/ui/src/components/WebRTCVideo.tsx
@@ -405,7 +405,7 @@ export default function WebRTCVideo() {
       });
 
       if (isRelativeMouseMode) {
-        videoElmRefValue.addEventListener("click",
+        containerRef.current?.addEventListener("click",
           () => {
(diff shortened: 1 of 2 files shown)


IDisposable:
Interesting idea! Note that we already have request full screen in the ActionBar but it (and Macros) are disabled when we're not fully connected... perhaps we should revise that to allow clicking the button even if HDMI is not running?

https://github.com/jetkvm/kvm/pull/822/files#diff-503edceb49a3a14452eb4663ab40040a5ff04a2a89f13e48be02539e2c939f08R488

tomwopat:
> Interesting idea! Note that we already have request full screen in the ActionBar but it (and Macros) are disabled when we're not fully connected... perhaps we should revise that to allow clicking the button even if HDMI is not running?
> 
> https://github.com/jetkvm/kvm/pull/822/files#diff-503edceb49a3a14452eb4663ab40040a5ff04a2a89f13e48be02539e2c939f08R488

Totally missed that. Seems simpler and cleaner from a UI perspective to do that, will take a look at modifying that.

tomwopat:
So I really am blind, the Fullscreen button is visible even when HDMI is not running. So just reverted the second commit.

IDisposable (approved):

ui/src/components/WebRTCVideo.tsx
```diff
@@ -542,7 +542,7 @@ export default function WebRTCVideo() {
                           style={{ animationDuration: "500ms" }}
                           className="animate-slideUpFade pointer-events-none absolute inset-0 flex items-center justify-center"
                         >
-                          <div className="relative h-full w-full rounded-md">
+                          <div className="relative h-full w-full rounded-md" onClick={requestPointerLock}>
```
adamshiervani: Now call `requestPointerLock()` in two places. The easiest change would simply be to replace `videoElmRefValue` with `containerRef` in row 408.

adamshiervani (changes requested):
We can add this in the next version, but the pointer and keyboard lock UX is really in need of some care.

IDisposable (approved):

PR #833, opened 2025-09-25
chore: remove msgpack from gin dependency

ym:
It should be able to reduce the binary size by ~7 MB.

https://github.com/jetkvm/kvm/actions/runs/18001570015
https://github.com/ym/jetkvm-kvm/actions/runs/18023015779
 
```
➜  Downloads du -sh jetkvm-app\ \(13\)/bin/jetkvm_app
 37M	jetkvm-app (13)/bin/jetkvm_app
➜  Downloads du -sh jetkvm-app\ \(14\)/bin/jetkvm_app
 30M	jetkvm-app (14)/bin/jetkvm_app
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ VERSION := 0.4.8
 PROMETHEUS_TAG := github.com/prometheus/common/version
 KVM_PKG_NAME := github.com/jetkvm/kvm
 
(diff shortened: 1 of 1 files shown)


IDisposable (approved):
... and here I was thinking we should switch away from JSON messaging ;)

PR #838, opened 2025-09-27
Improves OTA update reporting and process

IDisposable:
Improves the OTA (Over-The-Air) update process by adding better reporting and handling of reboot requirements. 

- More detailed logging, state management for reboots

### DIFF
--- a/ui/src/routes/devices.$id.settings.general.update.tsx
+++ b/ui/src/routes/devices.$id.settings.general.update.tsx
@@ -21,6 +21,11 @@ export default function SettingsGeneralUpdateRoute() {
   const { setModalView, otaState } = useUpdateStore();
   const { send } = useJsonRpc();
 
+  const onClose = useCallback(() => {
+    navigate(".."); // back to the devices.$id.settings page
+    window.location.reload(); // force a full reload to ensure the current device/cloud UI version is loaded
+  }, [navigate]);
+
   const onConfirmUpdate = useCallback(() => {
     send("tryUpdate", {});
     setModalView("updating");
@@ -36,9 +41,9 @@ export default function SettingsGeneralUpdateRoute() {
     } else {
       setModalView("loading");
     }
-  }, [otaState.updating, otaState.error, setModalView, updateSuccess]);
+  }, [otaState.error, otaState.updating, setModalView, updateSuccess]);
 
-  return <Dialog onClose={() => navigate("..")} onConfirmUpdate={onConfirmUpdate} />;
+  return <Dialog onClose={onClose} onConfirmUpdate={onConfirmUpdate} />;
 }
 
 export function Dialog({
--- a/ui/src/routes/devices.$id.tsx
+++ b/ui/src/routes/devices.$id.tsx
@@ -245,7 +245,7 @@ export default function KvmIdRoute() {
 
       shouldReconnect(event: WebSocketEventMap['close']) {
         console.debug("[Websocket] shouldReconnect", event);
-        return !isLegacySignalingEnabled.current && !connectionFailed;
+        return !isLegacySignalingEnabled.current;
       },
 
       onClose(event: WebSocketEventMap['close']) {
--- a/ui/src/routes/devices.$id.tsx
+++ b/ui/src/routes/devices.$id.tsx
@@ -78,10 +78,8 @@ const deviceLoader = async () => {
 
 const cloudLoader = async (params: Params<string>): Promise<CloudLoaderResp> => {
   const user = await checkAuth();
-
   const iceResp = await api.POST(`${CLOUD_API}/webrtc/ice_config`);
   const iceConfig = await iceResp.json();
-
   const deviceResp = await api.GET(`${CLOUD_API}/devices/${params.id}`);
 
   if (!deviceResp.ok) {
--- a/ui/src/routes/devices.$id.tsx
+++ b/ui/src/routes/devices.$id.tsx
@@ -1,7 +1,6 @@
 import { lazy, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   Outlet,
-  redirect,
   useLoaderData,
   useLocation,
   useNavigate,
@@ -16,7 +15,7 @@ import { motion, AnimatePresence } from "framer-motion";
 import useWebSocket from "react-use-websocket";
 
 import { cx } from "@/cva.config";
-import { CLOUD_API, DEVICE_API } from "@/ui.config";
+import { CLOUD_API } from "@/ui.config";
 import api from "@/api";
 import { checkAuth, isInCloud, isOnDevice } from "@/main";
 import {
@@ -51,8 +50,13 @@ import {
   RebootingOverlay,
 } from "@components/VideoOverlay";
 import { FeatureFlagProvider } from "@providers/FeatureFlagProvider";
-import { DeviceStatus } from "@routes/welcome-local";
 import { m } from "@localizations/messages.js";
+
+export type AuthMode = "password" | "noPassword" | null;
+
+interface LocalLoaderResp {
+  authMode: AuthMode;
+}
 
 interface CloudLoaderResp {
   deviceName: string;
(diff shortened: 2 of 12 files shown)


IDisposable:
@adamshiervani Not actually sure how to go about testing this...

ym (approved):

adamshiervani:
In the Makefile, you just set a low number and run `./dev_deploy.sh -i`. If you check for updates with that, you should be able to get through the OTA process. 

I'll try that out now.

adamshiervani (changes requested):
The OTA mechanism technically works. I was able to upgrade a device by lowering the version number in the `Makefile`. The code cleanup and the more granular state pushes to the client are great, but I have a few usability and robustness concerns.

The updated UI is confusing (I suspect `flex-col` is missing 😄). Despite the copy, it’s not immediately clear whether the reboot is automatic or requires the user to click “Reboot,” and it’s not obvious what happens if a reboot is already pending. All valid UX questions from the user’s perspective.

<details>
<summary>OTA Component</summary>
SCR-20250930-mgco
</details>

With the current implementation, if the reboot + IP assignment takes longer than the WebSocket reconnect policy allows, the “Reconnect” and “Reboot” buttons can lead the client to stale, lazy-loaded settings routes that break the app.

<details>
<summary>Stale settings pages</summary>
image
</details>

This PR aside, the only actual problem I noticed with OTA is the client-side state becomes desynchronized after a reboot: the client doesn’t perform a full refresh when the update completes. This is likely not caused by insufficient OTA-status events, but rather, the WebSocket reconnect threshold is being exceeded during the reboot/IP-assignment cycle (for example, DHCP can take a while). You can actually reproduce this by pulling the Ethernet cable during reboot -- the client hits the reconnect timeout and stops trying.

I would simply suggest extending the WebSocket reconnect threshold during updates to infinity. This way, the only thing that I can think of that could break the client-side part of the OTA mechanism is if DHCP assigned a different IP to the JetKVM, but that's fine for now.

IDisposable:
@adamshiervani clean-up complete and added exponential backoff and 2000 retries (so effectively 5.5 hours)

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -278,12 +286,51 @@ function UpdatingDeviceState({
...
+                      The JetKVM is preparing to reboot. This may take a while.<br/>
+                      If it doesn{"'"}t automatically reboot after a few minutes, you
+                      can manually request a reboot by clicking here:
+                      <LinkButton
+                        size="XS"
+                        theme="light"
+                        text="Reboot the KVM"
+                        LeadingIcon={MdRestartAlt}
+                        textAlign="center"
+                        reloadDocument={true}
```
IDisposable: This forces a page reload as soon as the navigation is registered

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -278,12 +286,51 @@ function UpdatingDeviceState({
...
+                    reconnect once it is back online.<br/>
+                    If it doesn{"'"}t reconnect automatically, you can manually
+                    reconnect by clicking here: 
+                    <LinkButton
+                      size="XS"
+                      theme="light"
+                      text="Reconnect to KVM"
+                      LeadingIcon={MdConnectWithoutContact}
+                      textAlign="center"
+                      reloadDocument={true}
```
IDisposable: This forces a page reload as soon as the navigation is registered

IDisposable:
> DHCP assigned a different IP to the JetKVM, but that's fine for now.

In that case, we wouldn't be able to reconnect if they were using an IP, but if they were using a hostname or accessing through app.jetkvm.com it would eventually reconnect ;)

IDisposable:
@adamshiervani This should be ready for retest/review

IDisposable:
Rebased and ready for review.

ui/src/routes/devices.$id.tsx
```diff
@@ -230,26 +221,31 @@ export default function KvmIdRoute() {
...
       onReconnectStop: (numAttempts: number) => {
-        console.debug("Reconnect stopped", numAttempts);
+        console.debug("Reconnect stopped after ", numAttempts, "attempts");
         cleanupAndStopReconnecting();
       },
 
       shouldReconnect(event: WebSocketEventMap['close']) {
         console.debug("[Websocket] shouldReconnect", event);
-        return !isLegacySignalingEnabled.current;
+        return !isLegacySignalingEnabled.current && !connectionFailed;
```
adamshiervani: Let's not stop trying to reconnect the WebSocket connection, especially given now we have an exponential backoff with a 10s ceiling.

ui/src/routes/devices.$id.tsx
```diff
@@ -62,27 +66,14 @@ interface CloudLoaderResp {
...
-
-  const deviceRes = await api.GET(`${DEVICE_API}/device`);
-  if (deviceRes.status === 401) return redirect("/login-local");
-  if (deviceRes.ok) {
-    const device = (await deviceRes.json()) as LocalDevice;
-    return { authMode: device.authMode };
-  }
-
-  throw new Error("Error fetching device");
+  const device = await checkAuth();
```
adamshiervani, line 75: This won't work as the `checkAuth()` -> `checkDeviceAuth()` can return a redirect. As far as I can understand, you can just replace the returns with throws, and it should work.

adamshiervani (changes requested):

ui/src/routes/devices.$id.tsx
IDisposable: Cool, I was just restoring code that was present in my original branch. Will remove.

ui/src/routes/devices.$id.tsx
IDisposable: I admit I'm confused here... this change was merely reusing the identical code in main.tsx instead of repeating it. The code I replaced _also_ did redirects.  

Image

As I understand the code here, the `checkAuth` is called by `deviceLoader`, which is conditionally call by the `loader` `LoaderFunction` (if on device), which is called while bringing up the page... do we not want to redirect to the _/welcome_ or _/local-login_ pages here?

If not, what do you want `throw`n?

adamshiervani: The loader function effectively expects a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response). If the response has a `Location` header, it will redirect to that location, and when it doesn't, it will provide that `Response` data to the component via `useLoaderData()`

Instead of creating `Response` objects all the time, `react-router` provides helpers: when it's `data` you can just return the data like a normal function, and when it's a redirect, you can call `redirect`. 

In the previous implementation, the redirect(`Response` + Location) was simply returned to the loader function whenever the device wasn't set up. In this PR, we call the `checkAuth()`, which returns either a `Response` with location (redirect) or a basic object with the authMode. Whatever that value is, it's put inside the `const device = ...` and returned to the loader function as an object. 

Throwing is a quick way of bypassing this flow and is also what we do in the cloud loader. I'm sure there are other ways to differentiate between the redirect and data; it's just the easiest I know. 

(This contains a bunch of generalizations, but I hope my point is still conveyed. <3 )

IDisposable: Thanks for the detailed info. I was completely missing that my change resulted in the `redirect` returned getting wrapped in an object. I will swing through and fix any call-sites I broke :(

adamshiervani (approved):

tadic-luka:
@IDisposable  I'm not familiar with react/tsx that much but your changes introduced compilation errors:
```
src/routes/devices.$id.settings.advanced.tsx:317:19 - error TS1308: 'await' expressions are only allowed within async functions and at the top levels of modules.

317                   await sleep(2000);
                      ~~~~~

  src/routes/devices.$id.settings.advanced.tsx:314:26
    314                 onClick={() => {
                                 ~~~~~~~
    Did you mean to mark this function as 'async'?

src/routes/devices.$id.settings.advanced.tsx:317:25 - error TS2304: Cannot find name 'sleep'.

317                   await sleep(2000);
                            ~~~~~

src/routes/devices.$id.settings.general.reboot.tsx:15:5 - error TS1308: 'await' expressions are only allowed within async functions and at the top levels of modules.

15     await sleep(1000);
       ~~~~~

  src/routes/devices.$id.settings.general.reboot.tsx:12:31
    12   const onClose = useCallback(() => {
                                     ~~~~~~~
    Did you mean to mark this function as 'async'?

src/routes/devices.$id.settings.general.reboot.tsx:15:11 - error TS2304: Cannot find name 'sleep'.

15     await sleep(1000);
             ~~~~~

src/routes/devices.$id.settings.general.update.tsx:27:5 - error TS1308: 'await' expressions are only allowed within async functions and at the top levels of modules.

27     await sleep(1000);
       ~~~~~

  src/routes/devices.$id.settings.general.update.tsx:24:31
    24   const onClose = useCallback(() => {
                                     ~~~~~~~
    Did you mean to mark this function as 'async'?


Found 5 errors in 3 files.

Errors  Files
     2  src/routes/devices.$id.settings.advanced.tsx:317
     2  src/routes/devices.$id.settings.general.reboot.tsx:15
     1  src/routes/devices.$id.settings.general.update.tsx:27
make: *** [Makefile:95: frontend] Error 2
```

I would say this is a regression and not an issue with the version of the npm or other build tool used to compile this code.

IDisposable:
Yikes, sorry about that, the rebase broke that code (it was an old PR). Thanks for the reminder to recompile after doing rebases 😄

PR #840, opened 2025-09-28
Release/0.4.8_i18n

oupula:
add i18n support , add zh-CN language


oupula:
image
image

ui/public/locales/en.json
```diff
@@ -0,0 +1 @@
+{}
```
IDisposable: Is there a reason we don't put all the _undisputed_ English stuff here, and the UK/US differences as overrides in their respective files? Are we actually doing fallback?

ui/public/locales/zh-CN.json
```diff
@@ -0,0 +1,758 @@
+{
```
IDisposable, line 1: I _personally_ do not feel qualified to review the Chinese (I could possibly do a reasonable job with German)... how do we get another Chinese person to chip in?

ui/src/components/ActionBar.tsx
```diff
@@ -17,7 +18,7 @@ import { cx } from "@/cva.config";
 import PasteModal from "@/components/popovers/PasteModal";
 import WakeOnLanModal from "@/components/popovers/WakeOnLan/Index";
 import MountPopopover from "@/components/popovers/MountPopover";
-import ExtensionPopover from "@/components/popovers/ExtensionPopover";
+import {ExtensionPopover} from "@/components/popovers/ExtensionPopover";
```
IDisposable: _nit_ should be `import { ExtensionPopover } from ...` (missing blanks)

ui/src/components/ConfirmDialog.tsx
```diff
@@ -56,20 +58,20 @@ const variantConfig = {
...
+// @ts-ignore
 export function ConfirmDialog({
   open,
   onClose,
   title,
   description,
   variant = "info",
-  confirmText = "Confirm",
-  cancelText = "Cancel",
+  confirmText = useTranslation('Confirm').toString(),
```
IDisposable: This is really inconsistent across the use... sometimes you use `useTranslation`, sometimes you use `t`. We should use one style consistently.

ui/src/components/Header.tsx
```diff
@@ -27,7 +27,7 @@ interface NavbarProps {
   kvmName?: string;
 }
 
-export default function DashboardNavbar({
+export default function  DashboardNavbar({
```
IDisposable: _nit_ extra blank

ui/src/components/InfoBar.tsx
```diff
@@ -59,30 +61,30 @@ export default function InfoBar() {
...
-                <span className="text-xs font-semibold">Video Size: </span>
+                <span className="text-xs font-semibold">{t('Video_Size')}: </span>
                 <span className="text-xs">{videoClientSize}</span>
               </div>
             ) : null}
 
             {(debugMode && mouseMode == "absolute") ? (
-              <div className="flex w-[118px] items-center gap-x-1">
-                <span className="text-xs font-semibold">Pointer:</span>
+              <div className="flex w-[140px] items-center gap-x-1">
```
IDisposable, line 77: This points out a real issue. With our aggressive width controls in CSS, we're going to break things when the other languages are wider (which is true for most other languages vs. English)

ui/src/components/InfoBar.tsx
```diff
@@ -93,31 +95,31 @@ export default function InfoBar() {
 
             {debugMode && (
               <div className="flex w-[156px] items-center gap-x-1">
-                <span className="text-xs font-semibold">USB State:</span>
-                <span className="text-xs">{usbState}</span>
+                <span className="text-xs font-semibold">{t('USB_State')}:</span>
+                <span className="text-xs">{t(usbState.replace(' ','_').toString())}</span>
```
IDisposable: Hmmm, I see what's happening here, but perhaps we should change the _enumerated_ states (used here as a key) to always be single-word (or PascalCase)... we could easily then just have the English version "restore" the spaces.

ui/src/components/PeerConnectionStatusCard.tsx
```diff
@@ -27,7 +28,8 @@ export default function PeerConnectionStatusCard({
   state?: RTCPeerConnectionState | null;
   title?: string;
 }) {
-  if (!state) return null;
+  if (!state) return <></>;
```
IDisposable: Interesting decision. in what way is returning an empty fragment superior to returning null?

ui/src/components/PeerConnectionStatusCard.tsx
```diff
@@ -55,13 +57,12 @@ export default function PeerConnectionStatusCard({
     },
   };
   const props = StatusCardProps[state];
-  if (!props) return;
-
+  if (!props) return (<div></div>);
```
IDisposable: Would an empty fragment be more appropriate? e.g. `return (<></>);`

ui/src/components/UsbInfoSetting.tsx
```diff
@@ -32,33 +33,33 @@ export interface USBConfig {
...
 
   const [usbConfigProduct, setUsbConfigProduct] = useState("");
   const [deviceId, setDeviceId] = useState("");
+  const usbConfigs = [
+        {
+            label: t('JetKVM_Default'),
+            value: "USB Emulation Device",
+        },
+        {
+            label: t('Logitech_Universal_Adapter'),
```
IDisposable, line 51: I don't think these keyboard names should be translated. They're actual product names.

ui/src/routes/devices.$id.settings.general.reboot.tsx
```diff
@@ -46,19 +47,20 @@ function ConfirmationBox({
   onYes: () => void;
   onNo: () => void;
 }) {
+  const { t } = useTranslation();
   return (
     <div className="flex flex-col items-start justify-start space-y-4 text-left">
       <div className="text-left">
         <p className="text-base font-semibold text-black dark:text-white">
-          Reboot JetKVM
+            {t('重启JetKVM')}
```
IDisposable: Should be `{t('Reboot_JetKVM')}` (with suitable translations in the language files)

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -405,53 +408,53 @@ function UpdateAvailableState({
...
     <div className="flex flex-col items-start justify-start space-y-4 text-left">
       <div className="text-left">
         <p className="text-base font-semibold text-black dark:text-white">
-          Update available
+            {t('Update_available')}
         </p>
         <p className="mb-2 text-sm text-slate-600 dark:text-slate-300">
-          A new update is available to enhance system performance and improve
-          compatibility. We recommend updating to ensure everything runs smoothly.
+            {t('A_new_update_is_available_to_enhance_system_performance_and_improve_compatibility_We_recommend_updating_to_ensure_everything_runs_smoothly')}
```
IDisposable, line 419: We should probably shorten huge keys like this.

ui/src/routes/devices.$id.tsx
```diff
@@ -365,8 +367,10 @@ export default function KvmIdRoute() {
 
       console.log("Trying to get remote session description");
       setLoadingMessage(
-        `Getting remote session description...  ${signalingAttempts.current > 0 ? `(attempt ${signalingAttempts.current + 1})` : ""}`,
-      );
+          t('Getting_remote_session_description',{
+            attempt:signalingAttempts.current > 0 ? t('attempt_num',{num:signalingAttempts.current + 1}) : ''
```
IDisposable: _nit_ need a space after the `attempt:` and `num:`

ui/vite.config.ts
```diff
@@ -5,64 +5,65 @@ import tsconfigPaths from "vite-tsconfig-paths";
 import basicSsl from "@vitejs/plugin-basic-ssl";
 
 declare const process: {
```
IDisposable, line 7: Please revert this white-space/indentation-only change.

IDisposable (changes requested):

ui/vite.config.ts
oupula: emm .. why after commit with the white-space ??
image

ui/src/components/UsbInfoSetting.tsx
oupula: in china , Logitech call the "罗技" 
image

ui/public/locales/en.json
oupula: current only en-US and zh-CN , if browser detect without language file ,  i18n-next will auto failback to en-US 
why put empty language file ,  
my browser language is zh-CN , but i18next auto request zh.json & zh-CN.json language file ,
in english(US) the same , i18next request both file (en.json & en-US.json)
just without 404 error

ui/src/components/ConfirmDialog.tsx
oupula: this is my problem,will fix in next commit

ui/src/routes/devices.$id.settings.general.update.tsx
oupula: Yes, that's a good idea. but I can't decide which key replace with long sentences

IDisposable: Just if it's more that 5 or six words, maybe?

ui/vite.config.ts
IDisposable: The file ends with three blank lines

IDisposable (changes requested):
Thanks for starting this off! 

We've discussed internally and would rather use [InLang.com](https://inlang.com/) for our localization package as it gives a much better developer-experience and quite a bit smaller at runtime. It compiles messages into JS functions, which are fully typed so you cannot easily forget replacement parameters and such. 

[Architecture](https://inlang.com/m/gerre34r/library-inlang-paraglideJs/architecture)
[Comparison](https://inlang.com/m/gerre34r/library-inlang-paraglideJs/comparison)

We know it's tedious to do this initial pass, and really appreciate your efforts. Can you switch from i18next to inlang or would you like to resubmit after we pull it in?

ui/src/routes/devices.$id.settings.video.tsx
IDisposable, line 137: Please restore this code, it's fine if it's not localized.

ui/public/locales/en.json
IDisposable: My comment was more about the standard for example of putting all _English_ translations in _en.json_ and then where (for there are differences) the US and UK differences (and only the differences) in _en-US.json_ and _en-UK.json_

ui/src/routes/devices.$id.settings.video.tsx
oupula: the two const use in image red frame area
image

oupula:
> Thanks for starting this off!
> 
> We've discussed internally and would rather use [InLang.com](https://inlang.com/) for our localization package as it gives a much better developer-experience and quite a bit smaller at runtime. It compiles messages into JS functions, which are fully typed so you cannot easily forget replacement parameters and such.
> 
> [Architecture](https://inlang.com/m/gerre34r/library-inlang-paraglideJs/architecture) [Comparison](https://inlang.com/m/gerre34r/library-inlang-paraglideJs/comparison)
> 
> We know it's tedious to do this initial pass, and really appreciate your efforts. Can you switch from i18next to inlang or would you like to resubmit after we pull it in?

if you are do this work,I won't get involved.
but, please not use the machine translate
example : 
extension is translate to "扩展" , will not translate to "扩大" , single word  "extension" translate "扩大" is right! software extension function "extension" translate "扩展"

image

oupula:
I am interest this project (with go , c and react native)
pikvm project run with python , I'm not interesting
I running the jetkvm in rasberrypi cm4  now . 
modify some code(include jetkvm-native [ video encode , you are using the rk sdk ] && jetkvm-app source code)
image

ui/src/routes/devices.$id.settings.video.tsx
IDisposable: I was talking about the developer mode stuff, which seems to using the wrong controlling value. Should be `  const { developerMode } = useSettingsStore();`

IDisposable:
> > We know it's tedious to do this initial pass, and really appreciate your efforts. Can you switch from i18next to inlang or would you like to resubmit after we pull it in?
> 
> if you are do this work,I won't get involved. 

We welcome your input! I started a (draft) PR #864 with the paraglide translation stubbed out... will need to extract all the rest of the language-text once we're settled into a pattern.

> please not use the machine translate example : extension is translate to "扩展" , will not translate to "扩大" , single word "extension" translate "扩大" is right! software extension function "extension" translate "扩展"

Oh, I get it. We will likely do a machine translation for a baseline, but we _really_ want native speakers to contribute. That said, I will try to line up all your work in this PR with what we ship initially, and then welcome corrections.

adamshiervani:
Closing this PR now, but @oupula feel more than welcome to edit the translation files in #864 or in `dev` once that it's merged. As @IDisposable said, we really want native speakers to contribute.

PR #865, opened 2025-10-02
Update openSUSE image reference to Leap 16.0

aylen384:


ui/src/routes/devices.$id.mount.tsx
```diff
@@ -374,8 +374,8 @@ function UrlView({
       icon: FedoraIcon,
     },
     {
-      name: "openSUSE Leap 15.6",
-      url: "https://download.opensuse.org/distribution/leap/15.6/iso/openSUSE-Leap-15.6-NET-x86_64-Media.iso",
+      name: "openSUSE Leap 16.0",
+      url: "https://download.opensuse.org/distribution/leap/16.0/offline/Leap-16.0-online-installer-x86_64.install.iso",
```
IDisposable, line 378: Shouldn't that be `https://download.opensuse.org/distribution/leap/16.0/offline/Leap-16.0-offline-installer-x86_64.install.iso` (note the 16.0-offline vs 16.0-online)?

Also, I'm getting a 403 for all releases at https://get.opensuse.org/leap/16.0 so this release might not actually be valid?

aylen384: The release is official since yesterday. The image is the right one (I don't know why the folder has this unintuitive name, though) because one is the network install image and one is the desktop image.
Due to the file size and the (I suspect) less efficient random access to the virtual cdrom drive that are (I guess) mapped to http range requests, I would use the netboot image instead of the offline installation media (which is 4.7GB large).

The 403 only happens when accessing via web browser but not via curl/wget.
This seems to be an issue with the CDN that I just reported to the openSUSE admins.

IDisposable: > Due to the file size and the (I suspect) less efficient random access to the virtual cdrom drive that are (I guess) mapped to http range requests, I would use the netboot image instead of the offline installation media (which is 4.7GB large).

Got it... wasn't sure you intentionally didn't point at a fully-offline install, but given we're just proxying the image, I see where the Network image is more reasonable.

IDisposable (approved):

PR #875, opened 2025-10-04
Add podman support

IDisposable:
Reimplement #141 since we've changed everything since


IDisposable:
@josharrington reimplemented since we changed everything out from under you.

ym (approved):

PR #878, opened 2025-10-07
refactoring network stack (and static ip)

ym:
UI TODO:

- [x] 🐛 Missing placeholder for `Hostname` field.
- [x] 🐛 The tab key does not work for all fields.
- [x] 🐛 The `Save` button doesn't work (this might require fixing the Tailwind animation class).
- [x] 🐛  There's an additional separator if there's any unsaved changes. 
- [x] 🐛 The `MAC Address` field is read-only, but the style is the same. We should probably change it to `span` instead of 'input'.
- [x] 🐛 Pressing the Esc key inside the `Apply network settings` confirmation dialog also closes the whole settings tab instead of the confirmation dialog.
- [x] 🚀 (optional) Make `subnet mask` optional for IPv4 static address, hide it if `/` is detected.

image

Backend TODO:

- [x] mDNS test
- [x] DHCP will run twice during initial boot (init and onLinkUp)
- [x] Domain might not work
- [ ] Add failover DNS
- [ ] DHCP
	- [ ] Support PTP address
	- [x] Support MTU
	- [ ] Support Multiple routers
	- [x] Renew
	- [ ] Release
	- [ ] 🚀 Support option `Classless Static Route Option`
	- [ ] 🚀 Support option `6rd`
	- [ ] 🚀 Support option `IEEE 802.1Q VLAN ID`
- [x] IPv6 address not being picked up sometimes
- [x] gocron.RemoveByTags might stuck sometimes, replace it with something else
- [x] Allow to switch between JetKVM's DHCP client and udhcpc
	- [x] JSONRPC
	- [x] Touchscreen UI
	- [x] WebUI
- [x] Clean up locks in network manager code


.vscode/settings.json
```diff
@@ -3,5 +3,12 @@
     "cva",
     "cx"
   ],
-  "git.ignoreLimitWarning": true
+  "gopls": {
+    "build.buildFlags": [
+      "-tags",
+      "synctrace"
```
IDisposable, line 9: How costly is this at runtime? Do we want to perhaps only have it enabled optionally or for debug builds?

cmd/main.go
IDisposable, line 150: Insert a comment here?
`// the easy path of renaming the file failed, so now we have to copy the file contents instead`

cmd/main.go
```diff
@@ -160,6 +183,8 @@ func createErrorDump(logFile *os.File) {
 	}
 
 	fmt.Printf("error dump created: %s\n", filePath)
```
IDisposable: s/created/copied

cmd/main.go
IDisposable: Don't we want to symlink this renamed file line we now in the "copied" case? https://github.com/jetkvm/kvm/pull/878/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R187

internal/confparser/confparser.go
```diff
@@ -355,27 +366,68 @@ func (f *FieldConfig) validateField() error {
...
+			fieldRef = fmt.Sprintf("field `%s`", f.Name)
+		}
+
 		switch validateType {
+		case "int":
+			if _, err := strconv.Atoi(val); err != nil {
+				return fmt.Errorf("field `%s` is not a valid integer: %s", f.Name, val)
+			}
+		case "ipv6_prefix_length":
+			valInt, err := strconv.Atoi(val)
```
IDisposable, line 387: For consistency we should be using `fieldRef` instead of `f.Name` on these (lines 382-393) because it is conceivable that we could have arrays of `int`s in some future configuration (less likely to happen for prefix length, but why not stay neutral in this, since it's really used only as the `error`'s message string).

internal/sync/log.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+var defaultLogger = logging.GetSubsystemLogger("synctrace")
+
+func logTrace(msg string) {
+	if defaultLogger.GetLevel() > zerolog.TraceLevel {
+		return
+	}
+
+	logTrack(3).Trace().Msg(msg)
```
IDisposable, line 23: ~What's the meaning of the magic `3`?~

Oh, I see now, this is the number of call-stack-frames to skip... perhaps can we "reason" out or document  where the 3 (and later 4) are derived and explain where it comes from.

internal/sync/log.go
```diff
@@ -0,0 +1,149 @@
...
+			l = l.With().
+				Str("func", details.Name()).
+				Logger()
+		}
+	}
+
+	return &l
+}
+
+func logLockTrack(i string) *zerolog.Logger {
```
IDisposable, line 50: Can we get a better name for `i` throughout this implementation? Perhaps `lockAddress`?

internal/sync/log.go
```diff
@@ -0,0 +1,149 @@
...
+	l := logTrack(4).
```
IDisposable, line 51: ~What are logTracks? Why 4 here?~

So logTrack is really a "call-site" specification to get the call-site information from the right stack entry.

The magic nature of the 4 could benefit from some explanation.

internal/sync/log.go
```diff
@@ -0,0 +1,149 @@
...
+	logTrack(3).Trace().Msg(msg)
+}
+
+func logTrack(callerSkip int) *zerolog.Logger {
+	l := *defaultLogger
+	if l.GetLevel() > zerolog.TraceLevel {
+		return &l
+	}
+
+	pc, file, no, ok := runtime.Caller(callerSkip)
```
IDisposable, line 32: So this amount of stack-crawl/reflection seems very costly to inject into the logging process at every single lock/unlock/etc call... we should memoize the information for a specific call site+skip count in a better way so that we don't pay this cost every single call?

Additionally guidance is to use CallersFrames instead of FuncForPC
> Always use runtime.CallersFrames to iterate over the frames rather than manually processing the PCs (e.g., via runtime.FuncForPC), as it properly handles inlining and adjusts return PCs to call sites.

It _is_ legal to just save the pc values for later mapping back to the function information so we can only do this reflection once-per function/program counter.

So, given all that, I think we should add a wrapped pc to the logger context (instead of all this) and have the serialization of the wrapper we store do the reflection only on the logger's stringification call.

internal/sync/once.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Once is a wrapper around the sync.Once
+type Once struct {
+	mu gosync.Once
+}
+
+// Do calls the function f if and only if Do has not been called before for this instance of Once.
+func (o *Once) Do(f func()) {
+	logTrace("Doing once")
+	o.mu.Do(f)
```
IDisposable, line 17: Do we want to log the completion of the `once` callback?

internal/sync/waitgroup.go
```diff
@@ -0,0 +1,30 @@
...
+// Add adds a function to the wait group
+func (w *WaitGroup) Add(delta int) {
+	logTrace("Adding to wait group")
+	w.wg.Add(delta)
+}
+
+// Done decrements the wait group counter
+func (w *WaitGroup) Done() {
+	logTrace("Done with wait group")
+	w.wg.Done()
```
IDisposable, line 23: Should we log the pair (e.g. Calling Done, Done complete)

internal/sync/waitgroup.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// Wait waits for the wait group to finish
+func (w *WaitGroup) Wait() {
+	logTrace("Waiting for wait group")
+	w.wg.Wait()
```
IDisposable, line 29: Should we log the pair (e.g. Waiting for group, Group wait complete)?

internal/timesync/timesync.go
```diff
@@ -239,12 +264,25 @@ Orders:
 	return nil
 }
 
+// Sync triggers a manual time sync
+func (t *TimeSync) Sync() error {
+	if !t.syncLock.TryLock() {
+		t.l.Warn().Msg("sync already in progress, skipping")
+		return nil
+	}
+	t.syncLock.Unlock()
```
IDisposable, line 273: Should this be a `defer t.syncLock.Unlock()` since we haven't actually done the sync yet?

pkg/nmlite/jetdhcpc/client.go
```diff
@@ -0,0 +1,407 @@
...
+	// If not set, it will default to dhcpv6's default (546).
+	V6ClientPort *int
+
+	// V4ServerAddr can be a unicast or broadcast destination for IPv4 DHCP
+	// messages.
+	//
+	// If not set, it will default to nclient4's default (DHCP broadcast
+	// address).
+	V4ServerAddr *net.UDPAddr
+
```
IDisposable, line 74: Do we need to support a `V4ClientPort`?

pkg/nmlite/jetdhcpc/client.go
```diff
@@ -0,0 +1,407 @@
...
+type Client struct {
+	types.DHCPClient
+
+	ifaces []string
+	cfg    Config
+	l      *zerolog.Logger
+
+	ctx context.Context
+
+	// TODO: support multiple interfaces
```
IDisposable, line 96: Do we _really_ need to support multiple interfaces inside the JetKVM device? When would there be multiple?

pkg/nmlite/jetdhcpc/client.go
```diff
@@ -0,0 +1,407 @@
...
+	}, nil
+}
+
+func resetTimer(t *time.Timer, l *zerolog.Logger) {
+	l.Debug().Dur("delay", defaultTimerDuration).Msg("will retry later")
+	t.Reset(defaultTimerDuration)
+}
+
+func getRenewalTime(lease *Lease) time.Duration {
+	if lease.RenewalTime <= 0 || lease.LeaseTime > maxRenewalAttemptDuration/2 {
```
IDisposable, line 162: Why the `/2`?

pkg/nmlite/jetdhcpc/client.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+func (c *Client) Renew() error {
+	c.timer4.Reset(defaultTimerDuration)
+	c.timer6.Reset(defaultTimerDuration)
+	return nil
+}
+
+func (c *Client) Release() error {
+	// TODO: implement
```
IDisposable, line 296: Something like?:
```golang
	c.lease4Mu.Lock()
	c.currentLease4 = nil
	c.lease4Mu.Unlock()

	c.lease6Mu.Lock()
	c.currentLease6 = nil
	c.lease6Mu.Unlock()

    c.apply()

	if c.cfg.OnLease4Change != nil {
		c.cfg.OnLease4Change(nil)
	}

	if c.cfg.OnLease6Change != nil {
		c.cfg.OnLease6Change(nil)
	}
```

pkg/nmlite/jetdhcpc/client.go
```diff
@@ -0,0 +1,407 @@
...
+		c.timer6.Stop()
+	}
+
+	c.timer6.Reset(defaultTimerDuration)
+}
+
+func (c *Client) Start() error {
+	if err := c.killUdhcpc(); err != nil {
+		c.l.Warn().Err(err).Msg("failed to kill udhcpc processes, continuing anyway")
+	}
```
IDisposable, line 347: `return err` before the closing brace?

Otherwise there's no path by with Start can return anything but `nil` so it really doesn't need to return an `error` at all...

pkg/nmlite/jetdhcpc/client.go
```diff
@@ -0,0 +1,407 @@
...
+
+	// deduplicate searchList
+	searchList = slices.Compact(searchList)
+
+	if c.cfg.UpdateResolvConf == nil {
+		c.l.Warn().Msg("no UpdateResolvConf function set, skipping resolv.conf update")
+		return
+	}
+
```
IDisposable, line 390: What should we be doing if everything is nil (perhaps because neither lease was set)?

pkg/nmlite/jetdhcpc/dhcp6.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if c.cfg.V6ServerAddr != nil {
+		mods = append(mods, nclient6.WithBroadcastAddr(c.cfg.V6ServerAddr))
+	}
+
+	conn, err := nclient6.NewIPv6UDPConn(iface.Attrs().Name, clientPort)
+	if err != nil {
+		return nil, err
+	}
+
```
IDisposable, line 113: Do we need a `defer conn.Close()` here?

pkg/nmlite/jetdhcpc/lease.go
```diff
@@ -0,0 +1,313 @@
...
+		return nil
+	}
+
+	address := iana.Options.OneAddress()
+	if address == nil {
+		return nil
+	}
+
+	lease.IPAddress = address.IPv6Addr
+	lease.Netmask = net.IP(net.CIDRMask(128, 128))
```
IDisposable, line 102: Hardcoded CIDR Mask of all 128? Shouldn't this have come from the lease netmask?

pkg/nmlite/jetdhcpc/lease.go
```diff
@@ -0,0 +1,313 @@
...
+		text := scanner.Text()
+		parts := strings.Split(text, " ")
+		uptime, err = time.ParseDuration(parts[0] + "s")
+
+		if err != nil {
+			return time.Time{}, fmt.Errorf("failed to parse uptime: %w", err)
+		}
+	}
+
+	relativeLeaseRemaining := (l.Uptime + l.LeaseTime) - uptime
```
IDisposable, line 157: I'm confused as to how _/proc/uptime_ (e.g. the number of seconds the JetKVM has been up) is related to the `l.Uptime`, especially because I don't see that value getting set against this `Lease` object??

pkg/nmlite/link/consts.go
```diff
@@ -0,0 +1,13 @@
+package link
+
+const (
+	// AfUnspec is the unspecified address family constant
+	AfUnspec = 0
+	// AfInet is the IPv4 address family constant
+	AfInet = 2
```
IDisposable, line 7: We should be using these in the jetdhcpc directory's files...

pkg/nmlite/link/manager.go
```diff
@@ -0,0 +1,544 @@
...
+		logger:               logger,
+		stateChangeCallbacks: make(map[string][]StateChangeCallback),
+	}
+	n.monitorStateChange()
+	return n
+}
+
+// GetNetlinkManager returns the singleton NetlinkManager instance
+func GetNetlinkManager() *NetlinkManager {
+	netlinkManagerOnce.Do(func() {
```
IDisposable, line 46: Could this just be `InitializeNetlinkManager(nil)`

pkg/nmlite/link/manager.go
```diff
@@ -0,0 +1,544 @@
...
+			mtu = addr.MTU
+		}
+	}
+	if expectedMTU != mtu {
+		if err := link.SetMTU(expectedMTU); err != nil {
+			nm.logger.Warn().Err(err).Int("expected_mtu", expectedMTU).Int("mtu", mtu).Msg("failed to set MTU")
+		}
+	}
+
+	addrs, err := nm.AddrList(link, family)
```
IDisposable, line 468: Wouldn't it be wise to hold at least an RLock for the chunk of this trying to reconcile?

pkg/nmlite/link/netlink.go
```diff
@@ -0,0 +1,164 @@
...
+func (l *Link) Refresh() error {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	return l.refresh()
+}
+
+// Attrs returns the attributes of the link
+func (l *Link) Attrs() *netlink.LinkAttrs {
+	l.mu.Lock()
```
IDisposable, line 79: Shouldn't all the read-only operations use `RLock`/`RUnlock`?

pkg/nmlite/dhcp.go
```diff
@@ -0,0 +1,219 @@
...
+	if err := dc.client.Release(); err != nil {
+		return fmt.Errorf("failed to release DHCP lease: %w", err)
+	}
+	return nil
+}
+
+// handleLeaseChange handles lease changes from the underlying DHCP client
+func (dc *DHCPClient) handleLeaseChange(lease *types.DHCPLease, isIPv6 bool) {
+	if lease == nil {
+		return
```
IDisposable, line 204: Do we want to add so the notifier will be informed of dropping leases?
```golang
	if dc.onLeaseChange != nil {
		dc.onLeaseChange(nil)
	}
```

pkg/nmlite/hostname.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	hm.hostname = hostname
+	hm.domain = domain
+
+	return hm.reconcileHostname()
+}
+
+func (hm *ResolvConfManager) Domain() string {
+	hm.mu.Lock()
```
IDisposable, line 34: `RLock` for read-only?

pkg/nmlite/hostname.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	return fmt.Sprintf("%s.%s", hostname, domain)
+}
+
+func (hm *ResolvConfManager) getHostname() string {
+	if hm.hostname != "" {
+		return hm.hostname
+	}
+	return "jetkvm"
```
IDisposable, line 66: Would having the serial number be helpful here?

pkg/nmlite/interface.go
```diff
@@ -0,0 +1,848 @@
...
+	}
+}
+
+// applyIPv4Static applies static IPv4 configuration
+func (im *InterfaceManager) applyIPv4Static() error {
+	if im.config.IPv4Static == nil {
+		return fmt.Errorf("IPv4 static configuration is nil")
+	}
+
+	im.logger.Info().Msg("stopping DHCP")
```
IDisposable, line 474: s/stopping DHCP/disabling IPv4 on DHCP

pkg/nmlite/interface.go
```diff
@@ -0,0 +1,848 @@
...
+
+// applyIPv4DHCP applies DHCP IPv4 configuration
+func (im *InterfaceManager) applyIPv4DHCP() error {
+	if im.dhcpClient == nil {
+		return fmt.Errorf("DHCP client not available")
+	}
+
+	// Enable DHCP
+	im.dhcpClient.SetIPv4(true)
+	return im.dhcpClient.Start()
```
IDisposable, line 508: What if it's already started?

pkg/nmlite/interface.go
```diff
@@ -0,0 +1,848 @@
...
+	return im.staticConfig.DisableIPv4()
+}
+
+// applyIPv6Static applies static IPv6 configuration
+func (im *InterfaceManager) applyIPv6Static() error {
+	if im.config.IPv6Static == nil {
+		return fmt.Errorf("IPv6 static configuration is nil")
+	}
+
+	im.logger.Info().Msg("stopping DHCPv6")
```
IDisposable, line 528: s/stopping/disabling

pkg/nmlite/interface.go
```diff
@@ -0,0 +1,848 @@
...
+	// Remove static IPv6 configuration
+	l, err := im.link()
+	if err != nil {
+		return fmt.Errorf("failed to get interface: %w", err)
+	}
+
+	netlinkMgr := getNetlinkManager()
+
+	// Ensure interface is up
+	if err := netlinkMgr.EnsureInterfaceUp(l); err != nil {
```
IDisposable, line 578: Does this timeout or block?

pkg/nmlite/interface.go
```diff
@@ -0,0 +1,848 @@
...
+		im.dhcpClient.SetIPv6(false)
+	}
+
+	// Disable IPv6
+	return im.staticConfig.DisableIPv6()
+}
+
+func (im *InterfaceManager) handleLinkStateChange(link *link.Link) {
+	{
+		im.stateMu.Lock()
```
IDisposable, line 631: `RLock`?

pkg/nmlite/utils.go
```diff
@@ -0,0 +1,76 @@
...
+	// Check for infinite lifetime (0xFFFFFFFF = 4294967295)
+	// This is used for static/permanent addresses
+	// Use uint32 to avoid int overflow on 32-bit systems
+	const infiniteLifetime uint32 = 0xFFFFFFFF
+	if uint32(lifetime) == infiniteLifetime || lifetime < 0 {
+		return nil // Infinite lifetime - no expiration
+	}
+
+	// For finite lifetimes (SLAAC addresses)
+	t := time.Now().Add(time.Duration(lifetime) * time.Second)
```
IDisposable, line 24: The probably should really have the "now" passed in from whatever the time of the reply that gave us a TTL is...?

pkg/nmlite/utils.go
```diff
@@ -0,0 +1,76 @@
...
+	return &t
+}
+
+func compareStringSlices(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	sort.Strings(a)
```
IDisposable, line 33: Is this going to modify the _callers_ slices?

mdns.go
```diff
@@ -10,10 +10,14 @@ func initMdns() error {
 	m, err := mdns.NewMDNS(&mdns.MDNSOptions{
 		Logger: logger,
 		LocalNames: []string{
-			networkState.GetHostname(),
-			networkState.GetFQDN(),
+			"jetkvm", "jetkvm.local",
```
IDisposable: This seems like a big downgrade... perhaps we should start up the MDNS later in the lifecycle when we have the correct host-names and such?

network.go
```diff
@@ -1,125 +1,206 @@
...
+}
+
+func toRpcNetworkSettings(config *types.NetworkConfig) *RpcNetworkSettings {
+	return &RpcNetworkSettings{
+		NetworkConfig: *config,
+	}
+}
+
+func restartMdns() {
+	if mDNS == nil {
```
IDisposable, line 70: I suspect we should actually be deferring the start to here and handling the "MDNS is disabled" here...

network.go
```diff
@@ -1,125 +1,206 @@
...
+		ntpServers := make([]string, len(networkManager.NTPServers()))
+		for i, server := range networkManager.NTPServers() {
+			ntpServers[i] = server.String()
+		}
+		networkLogger.Info().Strs("ntpServers", ntpServers).Msg("setting NTP servers from network manager")
+		timeSync.SetDhcpNtpAddresses(ntpServers)
+	}
+
+	// sync time
+	if err := timeSync.Sync(); err != nil {
```
IDisposable: Should this be in an async go routine?

IDisposable (commented):
Lots of comments on first review, I haven't looked at the browser/client code yet.

IDisposable (commented):
Still pending UI review

.vscode/settings.json
ym: This is for development only, and the performance penalty is barely noticeable, since we don’t develop directly on the RV1106 anyway.

pkg/nmlite/utils.go
ym: This is a known side effect. I’ve renamed the function to make its purpose clearer.

mdns.go
ym: mDNS won’t start before NetworkManager is initialized

.vscode/settings.json
IDisposable: I was looking at the what library which completely disappears when you build without the tag what on the command 
https://github.com/appliedgocode/what

mdns.go
IDisposable: So we should be getting the hostnames from the lease, right?

pkg/nmlite/utils.go
ym: As the value changes dynamically every time we call `netlink.LinkAddrs`, it is better to provide an absolute value here.

ui/src/routes/devices.$id.settings.network.tsx
```diff
@@ -72,418 +76,512 @@ export function LifeTimeLabel({ lifetime }: { lifetime: string }) {
...
-  };
+  const onSubmitGate = async (data: FieldValues) => {
+    const settings = prepareSettings(data);
+    const dirty = formState.dirtyFields;
 
-  const handleLldpModeChange = (value: LLDPMode | string) => {
-    setNetworkSettings({ ...networkSettings, lldp_mode: value as LLDPMode });
-  };
+    // Build list of critical changes for display
+    const changes: { label: string; from: string; to: string }[] = [];
```
adamshiervani, line 198: Some properties need to be confirmed in a modal.

ui/src/routes/devices.$id.settings.network.tsx
```diff
@@ -72,418 +76,512 @@ export function LifeTimeLabel({ lifetime }: { lifetime: string }) {
...
+
+      {/* Critical change confirm */}
+      <ConfirmDialog
+        open={showCriticalSettingsConfirm}
+        title="Apply network settings"
+        variant="warning"
+        confirmText="Apply changes"
+        onConfirm={() => {
+          setShowCriticalSettingsConfirm(false);
+          if (stagedSettings) onSubmit(stagedSettings);
```
adamshiervani, line 526: Once confirmed, we call the actualy `onSubmit()`

network.go
```diff
@@ -169,6 +171,54 @@ func setHostname(nm *nmlite.NetworkManager, hostname, domain string) error {
...
+	}
+
+	// IPv4 mode change requires reboot when using udhcpc
+	if newConfig.IPv4Mode.String != oldConfig.IPv4Mode.String && oldDhcpClient == "udhcpc" {
+		rebootRequired = true
+		networkLogger.Info().Str("old", oldConfig.IPv4Mode.String).Str("new", newConfig.IPv4Mode.String).Msg("IPv4 mode changed with udhcpc, reboot required")
+	}
+
+	// IPv4 static config changes require reboot
+	if newConfig.IPv4Static != nil && oldConfig.IPv4Static != nil {
```
IDisposable: What about when `oldConfig.IPv4Static` is `nil` (e.g. not static) and `newConfig.IPv4Static` is **not** `nil` (e.g. setting to a static address)? (or vice-versa)?

If we just used an empty (null-object-pattern) for the `nil` cases, we could just compare away...

ui/src/routes/devices.$id.tsx
```diff
@@ -666,6 +676,12 @@ export default function KvmIdRoute() {
         window.location.href = currentUrl.toString();
       }
     }
+
+    if (resp.method === "willReboot") {
```
adamshiervani, line 680: If the device sends us a `willReboot` message, we navigate to the root route and show the reboot overlay.

Currently, we have two flows using this: 
- Click on reboot, which will send `willReboot` right before actually rebooting
- Given a certain network change, the backend will send this

In the future we can use this flow for the OTA.

ota.go
```diff
@@ -488,6 +488,15 @@ func TryUpdate(ctx context.Context, deviceId string, includePreRelease bool) err
 
 	if rebootNeeded {
 		scopedLogger.Info().Msg("System Rebooting in 10s")
+
```
adamshiervani, line 491: We can use the `willReboot` flow for the OTA too. Might need some tweaking, but the flow and interface should be enough.

network.go
adamshiervani: Simplified it with a DeepEqual

adamshiervani (approved):

IDisposable (approved):

IDisposable:
Now I have to rebase the localization :)

PR #887, opened 2025-10-15
chore: disable sleep mode when detecting video format

### LINKED ISSUE #883: EDID unable to survive a JetKVM reboot
### Disclaimer

### Application version

0.4.9-dev202510091414

### System version

0.2.6

### Device model

None

### Extension model

None

### Remote device Hardware

_No response_

### Remote device OS

_No response_

### Bug description

The EDID configuration always gets reset to "JetKVM Default" after a reboot of the JetKVM device.

I don't have any custom EDID and am just using the given options there (Acer, ASUS and Dell).

ym:


adamshiervani (approved):

internal/native/cgo/video.c
```diff
@@ -51,6 +53,34 @@ RK_U64 get_us()
...
+}
+
+static void detect_sleep_mode()
+{
+    if (access(SLEEP_MODE_FILE, F_OK) != 0) {
+        sleep_mode_available = false;
+        return;
+    }
+    sleep_mode_available = true;
+    ensure_sleep_mode_disabled();
```
shakdoesgithub, line 92: i feel like this function should just do the detection of the sleep mode and the ensuring of the sleep mode being disabled should be called by the consumer of this function. I feel like this function is doing too much.

IDisposable (approved):

internal/native/video.go
```diff
@@ -69,12 +73,32 @@ func (n *Native) VideoSleepModeSupported() bool {
 	return n.sleepModeSupported
 }
 
+// useExtraLock uses the extra lock to execute a function.
+// if the lock is currently held by another goroutine, returns an error.
+//
+// it's used to ensure that only one change is made to the video stream at a time.
+// as the change usually requires to restart video streaming
+// TODO: check video streaming status instead of using a hardcoded timeout
+func (n *Native) useExtraLock(fn func() error) error {
```
IDisposable, line 82: Could this queuing-up be better handled by pushing the change into a channel to be handled inside a go-routine that serves up the changes in-order?

internal/native/cgo/video.c
```diff
@@ -301,11 +349,29 @@ static void *venc_read_stream(void *arg)
 }
 
 uint32_t detected_width, detected_height;
-bool detected_signal = false, streaming_flag = false;
+bool detected_signal = false, streaming_flag = false, streaming_stopped = true;
```
IDisposable, line 352: Shouldn't we declare `streaming_stopped` as `volatile` so the compiler knows that another thread might be mucking with them? (`streaming_flag` is protected by mutex, so should be fine, but wouldn't hurt to declare it `volatile` as well?)

IDisposable (commented):
I know this already merged, but I had a couple questions/comments

PR #896, opened 2025-10-21
feat/Add active hostname to device display

### LINKED ISSUE #617: Custom name on display
Hi, is there any way to customize the display view?
For example, can I change the IP address shown by JetKVM to something else, like the hostname of the connected device?
I currently have 3 JetKVMs, and this would be a nice feature to have.

IDisposable:
Closes #617

### Summary

Folks with more than one JetKVM want to be able to distinguish them at a glace. The current device display Home screen has enough space to display a name, and with recent networking changes, displaying the user-defined (or defaulted) hostname will help.

### UI Changes

- None on web UI

### Checklist

- [x] Lints pass; CI green
- ~[ ] Tricky parts are commented in code~ N/A
- ~[ ] Backward compatible with existing device firmware (See `DEVELOPMENT.md` for details)~ N/A


internal/native/eez/src/ui/styles.c
```diff
@@ -302,6 +294,37 @@ void remove_style_label_font16(lv_obj_t *obj) {
...
 };
 
+//
+// Style: LabelFontBook24
+//
+
+void init_style_label_font_book24_MAIN_DEFAULT(lv_style_t *style) {
+    init_style_label_font16_MAIN_DEFAULT(style);
+    
+    lv_style_set_text_font(style, &ui_font_font_book24);
```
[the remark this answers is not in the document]
IDisposable: Was a typo in the EEZ Studio, fixed.

adamshiervani (changes requested):
The screen is too crowded with IPv4, IPv6, MAC and hostname. Let's just replace the MAC address with the hostname if we have an IPv4 IP. Once the user has an IP, the MAC address isn't that useful to have on the device. If they really need it, they can copy it from the network settings.

IDisposable:
> The screen is too crowded with IPv4, IPv6, MAC and hostname. Let's just replace the MAC address with the hostname if we have an IPv4 IP. Once the user has an IP, the MAC address isn't that useful to have on the device. If they really need it, they can copy it from the network settings.

Done, also was able to restore the 30pt font on the IPv4

IDisposable:
Did a bunch of work on the JetKVM display (EEZ Studio stuff)

- Added MAC to the _Status Screen_ (so when it disappears from the _Home Screen_ there's still some place to see it).
- Made LVGL variables for _IPv4 Address_, _IPv6 Address_, _Hostname_, and _MAC Address_, so multiple device screens can show that information and it will update in all places as they are changed.
- Sets the _Device Id_ label on the _Status Screen_.
- Forces an immediate `uiTick` every time a label value or visibility is changed.
- Massive cleanup of the flags in the EEZ file to make all the screens scroll correctly and ripple up gestures.
- Added the _StatusScreenGesture_ action and _SwitchToNetwork_ action for use in back-arrows.
- Fixed a couple identifiers to make the code more reviewable.
- Hostname will scroll circularly (like IPv6 does now) if too wide.
- Added clicked event on the "Network" button on the to switch to the _Network Screen_.
- Eliminated the FontBold24 style and font because we can show the IPv4 in the original 30 point size.
- Set the "back" button for the children of the Advanced Menu to go back to that menu instead of the main screen.
-  Fixed the wonkiness with the advanced menu child screens where they saw the menu button as the immediate press of the action (reboot/reset/change DHCP)

The way things are working correctly we want the following things in objects in EEZ Studio:

- All _Screens_ should have no behavioral Flags and not be clickable.
- All children of every screen should have (probably only) _Gesture Bubble_ so the swipe gestures are active.
- All button literals or images should only _Gesture Bubble_ and _Event Bubble_ so the screens are swipe-able and the touching the text/graphic is treated as a _Click_ on the parent _Button_.
- All button images (like the left-pointing arrow) should have _Advance Hit Test_ to ensure clicking it activates the parent button.
- All buttons should be clickable, and events should be tied to the _Click_ event (not the _Press_ event). The only exception to this is the press-for-N-seconds buttons on the `Reset`, `Reset Configuration`, and `Switch DHCP Client` buttons, which use all the press events.
- When a region needs scrolling, there are two containers. The outer one adds a 4px right padding, and is clickable, but **not** set up for scrolling. It _must_ only have the _Gesture Bubble_ flag and should be set for 85% height. The inner container should be clickable, have _Gesture Bubble_, _Scrollable_, _Scroll Momentum_, _Scroll Elastic_, and **not* have _Scroll Chain Horizontal_ **nor** _Scroll Chain Vertical_. It should have _Scrollbar Mode: auto_, _Scroll Direction: VER_, and **not** _Scroll Snap *_. The inner container should have a height of 163px to ensure the scroll region renders correctly.
- The children elements of the scrollable regions (Container/Button/Literals) **must have**  _Scroll Chain Horizontal_, _Scroll Chain Vertical_, and _Gesture Bubble_. The button literal texts also need _Event Bubble_.
- Literal fields that are not "action" triggers are given only _Gesture Bubble_ and are not _Clickable_
- When a variable is changed, we call the screen tick for each screen it's on. This is done by calling them in the  `set_var_*` methods in _vars.c_ e.g. `tick_screen_home_screen();` or `tick_screen_status_screen();`. All of the `tick_screen_*` methods are declared in _vars.h_
- Styles got cleaned up to eliminate the "redundant settings" warnings.

adamshiervani:
> on the Settings, Advanced, and Status pages, the scrolling seems to have stopped working.

That's _very_ odd, as this is what I've been fighting with all along to get consistent behavior. Can you verify that your jetkvm.eez-project file matches what is currently checked in? (I did a bunch of force-pushes right at the end to get this over the line). Could you try running EEZ Studio and clicking the Build and seeing if it tries to change any of the "output" files in internal/native/eez/src/ui directory?

Just checking back here:

- [ ] The main Menu page also has the Network and Access buttons commented out (as it was when I started), so nothing to scroll.
- [ ] The Advanced page only has the Reboot/DHCP Client and Reset Configuration enable (Developer Mode and USB Emulation were already hidden when I started), so there's nothing to scroll.
- [ ] The Status page has the Account ID hidden (as it was when I started), so it doesn't actually have enough (non-hidden) entries to need scrolling.
- [ ] The Network screen (which is not reachable right now) has only three things so it doesn't need to scroll (but will if we add more)
- [ ] The About page does, and scrolls correctly on my device.

Please note any pages above that you want me to enable what's hidden and then we'll have a lot more scrolling ;)

> Also, for visual consistency, could you make the hostname font size match the MAC address?

- [x] Sure, just was trying to make it a bit bigger, but easy change :)

> And while you’re at it, could we add a “--” placeholder when there’s no IPv4? I think this might have been lost during the cgo migration.

- [x] Sure thing, though that will make the logic for the show/hide of the MAC Address a tiny bit wonky

IDisposable:
(oops, I accidentally edited your comment instead of replying, sorry @adamshiervani)

ym (approved):

display.go
```diff
@@ -229,7 +249,7 @@ func updateStaticContents() {
 	nativeInstance.UpdateLabelAndChangeVisibility("build_date", version.BuildDate)
 	nativeInstance.UpdateLabelAndChangeVisibility("golang_version", version.GoVersion)
 
-	// nativeInstance.UpdateLabelAndChangeVisibility("boot_screen_device_id", GetDeviceID())
+	nativeInstance.UpdateLabelAndChangeVisibility("device_id", GetDeviceID())
```
ym, line 257: Should we use `UISetVar` here too and remove `UpdateLabelAndChangeVisibility` completely?
EEZ has the ability to change visibility based on a variable, but I'm not sure whether we want to complicate the EEZ code or the Golang code.

ym (approved):

display.go
IDisposable: I think absolutely that would be much simpler! I just didn't know if you folks would mind that big a change. ~I'll do it (my) Friday night.~

ym: We can do it in a separate PR instead. Let's keep this one PR from getting too big :-)

IDisposable: Sure thing.

Any thoughts about Adam's comments? Do we need to reenable and of those other screens?

IDisposable: Added to backlog #982

IDisposable:
This is ready again, updated for the split-off of native (so some gRPC changes) 

@adamshiervani @ym

IDisposable:
Resolved conflicts

IDisposable:
If we can get this merged @adamshiervani  I can start on the "set IP on the device touchscreen" you requested

russkel:
Really interested in showing the hostname on the display!

tgwaste:
Is this ever going to be a thing??

PR #898, opened 2025-10-23
refactor: ota redirecting 

adamshiervani:
- Move away from the old `useJsonRPC` API to a non-reactive interface.
	- Easier to consume
	- The new reboot URL `/settings/general/update` checks for updates on our API. The DNS might not be ready, so added timeouts and retries. 
- Add both `appversion` and `systemversion` to the OTA redirect url
- Rename to `RedirectUrl` to `RedirectTo`

### DIFF
--- a/ui/src/components/VideoOverlay.tsx
+++ b/ui/src/components/VideoOverlay.tsx
@@ -483,6 +483,7 @@ export function RebootingOverlay({ show, postRebootAction }: RebootingOverlayPro
           const targetUrl = new URL(postRebootAction.redirectTo, window.location.origin);
 
           window.location.href = targetUrl.href;
+          window.location.reload();
         }
       } catch (err) {
         // Ignore errors - they're expected while device is rebooting
--- a/ui/src/components/VideoOverlay.tsx
+++ b/ui/src/components/VideoOverlay.tsx
@@ -474,13 +474,13 @@ export function RebootingOverlay({ show, postRebootAction }: RebootingOverlayPro
 
         if (response.ok) {
           // Device is available, redirect to the specified URL
-          console.log('Device is available, redirecting to:', postRebootAction.redirectUrl);
+          console.log('Device is available, redirecting to:', postRebootAction.redirectTo);
 
           // URL constructor handles all cases elegantly:
           // - Absolute paths: resolved against current origin
           // - Protocol-relative URLs: resolved with current protocol
           // - Fully qualified URLs: used as-is
-          const targetUrl = new URL(postRebootAction.redirectUrl, window.location.origin);
+          const targetUrl = new URL(postRebootAction.redirectTo, window.location.origin);
 
           window.location.href = targetUrl.href;
         }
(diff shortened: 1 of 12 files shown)


ui/src/components/VideoOverlay.tsx
```diff
@@ -474,9 +474,15 @@ export function RebootingOverlay({ show, postRebootAction }: RebootingOverlayPro
...
-          window.location.reload();
+          console.log('Device is available, redirecting to:', postRebootAction.redirectTo);
+
+          // URL constructor handles all cases elegantly:
+          // - Absolute paths: resolved against current origin
+          // - Protocol-relative URLs: resolved with current protocol
+          // - Fully qualified URLs: used as-is
+          const targetUrl = new URL(postRebootAction.redirectTo, window.location.origin);
+
+          window.location.href = targetUrl.href;
```
IDisposable, line 485: Do you think the `window.location.reload()` is still useful when the client changes (to force the HTML to reload)

ui/src/utils/jsonrpc.ts
```diff
@@ -20,16 +22,31 @@ export interface JsonRpcCallResponse {
 
 let rpcCallCounter = 0;
 
-export function callJsonRpc(options: JsonRpcCallOptions): Promise<JsonRpcCallResponse> {
-  return new Promise((resolve, reject) => {
-    // Access the RTC store directly outside of React context
-    const rpcDataChannel = useRTCStore.getState().rpcDataChannel;
+// Helper: sleep utility for retry delays
+const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
```
IDisposable: Should we export this?

ui/src/utils/jsonrpc.ts
```diff
@@ -40,32 +57,90 @@ export function callJsonRpc(options: JsonRpcCallOptions): Promise<JsonRpcCallRes
...
+        rpcDataChannel,
+        options,
+        abortController.signal,
+      );
+
+      clearTimeout(timeoutId);
+
+      // Retry on error if attempts remain
+      if (response.error && attempt < maxRetries) {
+        await sleep(1000);
```
IDisposable: Do we want an exponential backoff like I did in this PR?
https://github.com/jetkvm/kvm/pull/838/files#diff-c5d4fe6a19259940339ea9a409ab7bc070c9a1417d2b08cb9c06a11be54e2d3eR225-R229

ui/src/utils/jsonrpc.ts
```diff
@@ -40,32 +57,90 @@ export function callJsonRpc(options: JsonRpcCallOptions): Promise<JsonRpcCallRes
...
+        continue;
+      }
```
IDisposable, line 123: _nit_ alignment seems off

ui/src/utils/jsonrpc.ts
```diff
@@ -40,32 +57,90 @@ export function callJsonRpc(options: JsonRpcCallOptions): Promise<JsonRpcCallRes
...
+
+      return response;
+    } catch (error) {
+      clearTimeout(timeoutId);
+
+      // Retry on timeout/error if attempts remain
+      if (attempt < maxRetries) {
+        await sleep(1000);
```
IDisposable: backoff?

IDisposable (commented):
Looks pretty good, this will replace #838

ui/src/utils/jsonrpc.ts
adamshiervani: Hmm, it should be formatted by Prettier. Does it look good after my latest commits?

IDisposable (approved):
There are some other little things I did in my attempt, that I can stack over this when it is merged, but this looks good to go now.

PR #899, opened 2025-10-23
More UI translations

adamshiervani:
- Add Reboot translations
- Run all language files through an LLM to (hopefully) improve them


IDisposable (approved):
Looks great, except you didn't `npm run i18n` so the language files aren't sorted, which is important for long-term GIT diff stability

adamshiervani:
Thanks - just sorted the language files!

Issue #344, opened 2025-04-10
cant downgrade from dev to stable without flashing new firmware


si458:
when you tick `dev channel updates` and click `check for updates` in the general tab, 
it upgrades from stable to dev as expected

BUT

if you untick `dev channel updates` and click `check for updates` in the general tab,
it says your on the latest version

i would have expected it to pop up saying 'upgrade to 0.3.8' even tho im on '0.3.9-devXXXXXXXX'

PR #912, opened 2025-10-29
OTA refactor

ym:
Also implements https://github.com/jetkvm/kvm/issues/344

This PR will be backported to https://github.com/jetkvm/kvm/tree/release/049 as part of `0.4.10`. Since we don’t plan to introduce i18n in the `0.4.x` series, the translation files will be removed during the backport.


ui/localization/messages/en.json
```diff
@@ -897,5 +892,23 @@
     "wake_on_lan_invalid_mac": "Invalid MAC address",
     "wake_on_lan_magic_sent_success": "Magic Packet sent successfully",
     "welcome_to_jetkvm": "Welcome to JetKVM",
-    "welcome_to_jetkvm_description": "Control any computer remotely"
+    "welcome_to_jetkvm_description": "Control any computer remotely",
```
IDisposable: @ym you overwrote the changes @adamshiervani did in [this commit](https://github.com/jetkvm/kvm/pull/912/commits/a69f7c9c504a286acaba43d03be190aee064112d#diff-5cb481ef974356d3bc9c272545e9ed3660ae2bf850987715f0706694aedad318)

Whenever you change a language file, it's important to run an `npm run i18n` to ensure the language files are properly sorted before `git add . && git commit` so that later PRs are more stable.

adamshiervani:
@ym can you add proper handling of 404? Right now, if you specify a version that doesn't exist, you get redirected to the `general/update` with a notification that says [Object Object].

ota.go
```diff
@@ -73,519 +78,146 @@ func GetLocalVersion() (systemVersion *semver.Version, appVersion *semver.Versio
...
-
-	updateStatus, err := GetUpdateStatus(ctx, deviceId, includePreRelease)
-	if err != nil {
-		otaState.Error = fmt.Sprintf("Error checking for updates: %v", err)
-		scopedLogger.Error().Err(err).Msg("Error checking for updates")
-		return fmt.Errorf("error checking for updates: %w", err)
+	currentAppTargetVersion := otaState.GetTargetVersion("app")
+	appTargetVersionChanged := currentAppTargetVersion != components.AppTargetVersion
+	updateParams.AppTargetVersion = components.AppTargetVersion
+	if err := otaState.SetTargetVersion("app", components.AppTargetVersion); err != nil {
```
adamshiervani: Why do we need to store the target version in the Go state? Couldn’t we just use ephemeral RPC calls instead? On the advanced page, we could trigger a call to validate the version, and on the update page, simply pass the desired version as a parameter. For standard upgrade paths, if no version is specified, we default to the latest version.


This makes the update path quite complex , and I can't see the reason. Am I missing something?

IDisposable (approved):

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -37,6 +42,25 @@ export default function SettingsGeneralUpdateRoute() {
...
   }, [send, setModalView, setShouldReload]);
 
+  const onConfirmCustomUpdate = useCallback((appTargetVersion?: string, systemTargetVersion?: string) => {
+    const components = [];
+    if (appTargetVersion) components.push("app");
+    if (systemTargetVersion) components.push("system");
+
+    send("tryUpdateComponents", {
+      params: {
+        components: components.join(","),
```
adamshiervani: Why use a string and not an array of strings when we have JSON?

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -37,6 +42,25 @@ export default function SettingsGeneralUpdateRoute() {
     setModalView("updating");
   }, [send, setModalView, setShouldReload]);
 
+  const onConfirmCustomUpdate = useCallback((appTargetVersion?: string, systemTargetVersion?: string) => {
+    const components = [];
+    if (appTargetVersion) components.push("app");
+    if (systemTargetVersion) components.push("system");
+
+    send("tryUpdateComponents", {
```
adamshiervani, line 53: Use the new API calling here, like you did in `checkUpdateComponents()`

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -138,6 +191,22 @@ function LoadingState({
...
+      return await getVersionInfo();
+    }
+    const params : updateParams = {
+      components: "",
+      appTargetVersion: customAppVersion,
+      systemTargetVersion: customSystemVersion,
+    };
+    if (customAppVersion) params.components += ",app";
+    if (customSystemVersion) params.components += ",system";
+    params.components = params.components?.replace(/^,+/, "");
```
adamshiervani: Why not just an array if we have JSON/

ui/src/routes/devices.$id.settings.general.update.tsx
```diff
@@ -138,6 +191,22 @@ function LoadingState({
 
   const progressBarRef = useRef<HTMLDivElement>(null);
 
+  const checkUpdate = useCallback(async () => {
+    if (!customAppVersion && !customSystemVersion) {
+      return await getVersionInfo();
```
adamshiervani, line 192: This really needs a refactor on the FE, but let's do that later. It doesn't really make sense that we call a hook that calls the new JSON API call just to set a Zustand variable.

adamshiervani: The only difference between `getVersionInfo` and `checkUpdateComponents` is that `checkUpdateComponents` can check the status for specific components and specific versions, but both return the same data - `SystemVersionInfo`. 

The thing I don't get is why we have two different methods - it feels like we're adding more API surface to support in the future - and I can't see why. Why can't we just expand the `getVersionInfo` but add `appTargetVersion` and `systemTargetVersion` as parameters?

As for backward compatibility with older devices, we simply use the already existing FeatureFlag hooks to determine whether we should check for the latest of a specific app or system version. This feature will, anyway, need to check for a version number, as the `checkUpdateComponents` won't exist on older devices.

internal/ota/state.go
```diff
@@ -0,0 +1,277 @@
...
+	SystemUpdateAvailable bool            `json:"systemUpdateAvailable"`
+	AppUpdateAvailable    bool            `json:"appUpdateAvailable"`
+
+	// for backwards compatibility
+	Error string `json:"error,omitempty"`
+}
+
+// PostRebootAction represents the action to be taken after a reboot
+// It is used to redirect the user to a specific page after a reboot
+type PostRebootAction struct {
```
adamshiervani, line 53: This is not an OTA-related structure; currently, reboot, IP and OTA are the consumers, but I'm sure this will expand.

ui/src/routes/devices.$id.settings.advanced.tsx
```diff
@@ -173,6 +185,68 @@ export default function SettingsAdvancedRoute() {
...
+  }, []);
+
+  const handleVersionUpdate = useCallback(async () => {
+    const components = updateTarget === "both" ? ["app", "system"] : [updateTarget];
+    let versionInfo: SystemVersionInfo | undefined;
+    try {
+      // we do not need to set it to false if check succeeds,
+      // because it will be redirected to the update page later
+      setVersionUpdateLoading(true);
+      versionInfo = await checkUpdateComponents({
```
adamshiervani, line 210: Same here as above, as in `/settings/general/update`.

ota.go
```diff
@@ -73,519 +77,120 @@ func GetLocalVersion() (systemVersion *semver.Version, appVersion *semver.Versio
...
-		Transport: transport,
-	}
-
-	resp, err := client.Do(req)
+func getUpdateStatus(includePreRelease bool) (*ota.UpdateStatus, error) {
+	updateStatus, err := otaState.GetUpdateStatus(context.Background(), ota.UpdateParams{
+		DeviceID:          GetDeviceID(),
+		IncludePreRelease: includePreRelease,
+	})
+	// to ensure backwards compatibility,
```
adamshiervani, line 89: The Go backend never needs to retain backward compatibility. The only backward compatibility needed is on the frontend - when the latest cloud meets an old device.

ota.go
```diff
@@ -73,519 +77,120 @@ func GetLocalVersion() (systemVersion *semver.Version, appVersion *semver.Versio
...
-	scopedLogger := otaLogger.With().
-		Str("deviceId", deviceId).
-		Bool("includePreRelease", includePreRelease).
-		Logger()
-
-	scopedLogger.Info().Msg("Trying to update...")
-	if otaState.Updating {
-		return fmt.Errorf("update already in progress")
+// rpcCheckUpdateComponents checks the update status for the given components
+func rpcCheckUpdateComponents(params updateParams, includePreRelease bool) (*ota.UpdateStatus, error) {
```
adamshiervani, line 157: I don't think this is needed. This is essentially `getUpdateStatus` which an ability to specify app and system versions.

internal/ota/ota.go
```diff
@@ -0,0 +1,350 @@
...
+			systemUpdateStatus.available = systemVersionRemote.String() != systemVersionLocal.String()
+		}
+	} else {
+		systemUpdateStatus.available = false
+	}
+
+	return nil
+}
+
+// GetUpdateStatus returns the current update status (for backwards compatibility)
```
adamshiervani, line 413: Backwards compatibility is not needed in go - new devices will always have new JS.

internal/ota/ota.go
```diff
@@ -0,0 +1,350 @@
...
+// getUpdateURL returns the update URL for the given parameters
+func (s *State) getUpdateURL(params UpdateParams) (string, error, bool) {
+	updateURL, err := url.Parse(s.releaseAPIEndpoint)
+	if err != nil {
+		return "", fmt.Errorf("error parsing update metadata URL: %w", err), false
+	}
+
+	isCustomVersion := false
+
+	appTargetVersion := s.GetTargetVersion("app")
```
adamshiervani: This component code structure with an arbitrary amount of updatable components is far too complex for our needs - and would fall under pre-optimization, especially when we have hardcoded `AppTargetVersion` in the response. 

If we ever do it, adding more updateable components will be a very deliberate act, and the updating mechanism will be only a small part of that effort.

adamshiervani (changes requested):
Unfortunately, it's still far too complex, and as this is OTA, we need to keep it simple and extremely understandable.

In general, try to make surgical changes, only change interfaces when you absolutely have to, try to reuse existing code paths, and be mindful when adding new methods. Simply adding more code paths for the same thing isn't maintainable. 

This will all be easier to reason about when we can serve old JS to old devices, but until then, we need to be mindful with our changes.

ota.go
IDisposable: While this is technically true for released software, it can get tricky for dev releases. It's best to think through what a change will do, and not _break_ anything for mixed device/UI code. In this case, we're fine returning false on error :)

internal/ota/state.go
ym: I agree. I don’t see a clearly better package for this struct right now. How about introducing a small `types` package to hold shared structs?

ota.go
ym: I’m not sure which part you’re referring to. If you mean `rpcCheckUpdateComponents`, we’d need to introduce a new RPC method when parameters change; otherwise we break backward compatibility.

For example, the cloud may run the `production` branch while a device runs `development` (newer than production). Until we implement skew protection, this mismatch is unavoidable.

internal/ota/state.go
adamshiervani: Sounds good!

IDisposable (approved):

internal/ota/ota.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	if !params.CheckOnly {
+		s.updating = true
+		s.triggerStateUpdate()
+	}
+
+	appUpdate, systemUpdate, err := s.getUpdateStatus(ctx, params)
+	if err != nil {
+		return s.componentUpdateError("Error checking for updates", err, &scopedLogger)
```
adamshiervani, line 213: If there is an error here, `s.updating` will dangle.

internal/ota/ota.go
```diff
@@ -0,0 +1,334 @@
...
+
+	return nil
+}
+
+// UpdateParams represents the parameters for the update
+type UpdateParams struct {
+	DeviceID          string            `json:"deviceID"`
+	Components        map[string]string `json:"components,omitempty"`
+	IncludePreRelease bool              `json:"includePreRelease"`
+	CheckOnly         bool              `json:"checkOnly"`
```
adamshiervani: This doesn't seem to be used anywhere. The `Settings > Advanced` update feature uses `checkUpdateComponents`. This would simplify the `doUpdate` function

adamshiervani (changes requested):

adamshiervani (commented):
The progress bars in the update UI are also broken. It shows `Rebooting to complete the update…` which likely means that some of the state is not backwards compatible as the part hasn't been changed in JS.

internal/ota/ota.go
ym: Could you specify which part is not being used? I'm a little confused.

adamshiervani: Oh, sorry! I couldn't find `CheckOnly` used as anywhere.

ym:
> The progress bars in the update UI are also broken. It shows `Rebooting to complete the update…` which likely means that some of the state is not backwards compatible as the part hasn't been changed in JS.


Fixed now. The issue was caused by the frontend logic: it checks whether `UpdatedAt` is `nil` to determine if the update has completed. My latest change set `UpdatedAt` to zero, which translates to `0001-01-01 00:00:00` instead of `nil`, so the UI treated the update as incomplete.

adamshiervani:
There is still a bug: 

1. Deploy PR 
2. Check for update
3. Click "Update now".

The update starts, but I'm taken to the Checking for updates screen with the progress bars.

ym:
> There is still a bug:
> 
> 1. Deploy PR
> 2. Check for update
> 3. Click "Update now".
> 
> The update starts, but I'm taken to the Checking for updates screen with the progress bars.

It took me some time to reproduce this issue. The root cause is that our release API is sometimes too slow and exceeds `attemptTimeoutMs`. The default timeout is 5s, but in my tests the request took around 6s, which triggers the fallback behavior and sends the UI back to the "Checking for updates" screen.

<del>I think we should move the `/releases` endpoint to a separate backend so it isn’t affected by load on the main API, which should improve both performance and reliability.</del>

_It turns out my PŸUR cable connection went down again, and the failover didn’t re-route DNS or disable IPv6 :(_

In any case, please let me know if you’re still able to reproduce the issue.
If not, please run this branch with `LOG_TRACE_SCOPES=ota` so I can see which part is taking so long to complete.

```
2025-11-18T03:09:41Z TRC jetkvm ota checkUpdateStatus: getLocalVersion appVersionLocal=0.4.9-dev202511180307 duration=0.0385 requestID=aa45c395-f2e9-4106-8bd9-87870d401420 systemVersionLocal=0.2.7
2025-11-18T03:09:41Z TRC jetkvm ota fetchUpdateMetadata: getUpdateURL duration=0.746958 requestID=aa45c395-f2e9-4106-8bd9-87870d401420
2025-11-18T03:09:47Z TRC jetkvm ota fetchUpdateMetadata: response duration=6023.059605 requestID=aa45c395-f2e9-4106-8bd9-87870d401420 status=200
2025-11-18T03:09:47Z TRC jetkvm ota fetchUpdateMetadata: completed duration=6024.226272 requestID=aa45c395-f2e9-4106-8bd9-87870d401420
2025-11-18T03:09:47Z TRC jetkvm ota checkUpdateStatus: fetchUpdateMetadata duration=6031.541564 remoteMetadata={"appHash":"220db72f14a0661d5d62aaafbe47089c706f2562649ec4e794511906d499d568","appUrl":"https://update.jetkvm.com/app/0.4.9/jetkvm_app","appVersion":"0.4.9","systemHash":"da62bc0246d84e575c719a076a8f403e16e492192e178ecd68bc04ada853f557","systemUrl":"https://update.jetkvm.com/system/0.2.7/system.tar","systemVersion":"0.2.7"} requestID=aa45c395-f2e9-4106-8bd9-87870d401420
2025-11-18T03:09:47Z TRC jetkvm ota checkUpdateStatus: remoteMetadataToComponentStatus [app] available=true availableReason="remote version 0.4.9 is greater than local version 0.4.9-dev202511180307" dependsOn=[] downloadFinishedAt=0001-01-01T00:00:00Z downloadProgress=0 hash=220db72f14a0661d5d62aaafbe47089c706f2562649ec4e794511906d499d568 localVersion=0.4.9-dev202511180307 pending=false requestID=aa45c395-f2e9-4106-8bd9-87870d401420 updateProgress=0 updatedAt=0001-01-01T00:00:00Z url=https://update.jetkvm.com/app/0.4.9/jetkvm_app verificationProgress=0 verifiedAt=0001-01-01T00:00:00Z version=0.4.9
2025-11-18T03:09:47Z TRC jetkvm ota checkUpdateStatus: remoteMetadataToComponentStatus [system] available=false availableReason="remote version 0.2.7 is greater than local version 0.2.7" dependsOn=[] downloadFinishedAt=0001-01-01T00:00:00Z downloadProgress=0 hash=da62bc0246d84e575c719a076a8f403e16e492192e178ecd68bc04ada853f557 localVersion=0.2.7 pending=false requestID=aa45c395-f2e9-4106-8bd9-87870d401420 updateProgress=0 updatedAt=0001-01-01T00:00:00Z url=https://update.jetkvm.com/system/0.2.7/system.tar verificationProgress=0 verifiedAt=0001-01-01T00:00:00Z version=0.2.7
2025-11-18T03:09:47Z TRC jetkvm ota checkUpdateStatus: completed duration=6041.999274 requestID=aa45c395-f2e9-4106-8bd9-87870d401420
```

internal/ota/ota.go
```diff
@@ -0,0 +1,432 @@
...
+	if params.RequestID != "" {
+		logger = logger.With().Str("requestID", params.RequestID).Logger()
+	}
+	t := time.Now()
+	traceLogger := func() *zerolog.Event {
+		return logger.Trace().Dur("duration", time.Since(t))
+	}
+
+	url, err, isCustomVersion := s.getUpdateURL(params)
+	traceLogger().Err(err).
```
IDisposable, line 113: Shouldn't this be inside the `if err != nil {` block?

internal/ota/ota.go
```diff
@@ -0,0 +1,432 @@
...
+		return nil, fmt.Errorf("error getting update URL: %w", err)
+	}
+
+	traceLogger().
+		Str("url", url).
+		Msg("fetching update metadata")
+
+	req, err := s.newHTTPRequestWithTrace(ctx, "GET", url, nil, traceLogger)
+	if err != nil {
+		return nil, fmt.Errorf("error creating request: %w", err)
```
IDisposable, line 125: Should we insert `traceLogger().Err(err).Msg("creating request")`

internal/ota/ota.go
```diff
@@ -0,0 +1,432 @@
...
+
+	if systemUpdate.pending {
+		if err := s.updateSystem(ctx, systemUpdate); err != nil {
+			return s.componentUpdateError("Error updating system", err, &scopedLogger)
+		}
+	} else {
+		scopedLogger.Info().Msg("System is up to date")
+	}
+
+	s.updating = false
```
IDisposable: Should we just wrap this in a `defer` up higher in the function to ensure we always clear the flag?

internal/ota/ota.go
```diff
@@ -0,0 +1,432 @@
...
+			}
+		}
+
+		scopedLogger.Info().Msg("System Rebooting due to OTA update")
+
+		redirectUrl := fmt.Sprintf("/settings/general/update?version=%s", systemUpdate.version)
+
+		if params.ResetConfig {
+			scopedLogger.Info().Msg("Resetting config")
+			if err := s.resetConfig(); err != nil {
```
IDisposable, line 271: Since this `resetConfig()` call will really run against the _current_ app's code, this means that _especially in a downgrade_ there is potential for us to create a "default" config file that is _still_ not readable by the new app version.

Would it be better to create a sentinel flag that the (new version) app-startup could detect, do the reset config itself, then delete the sentinel? That way we know the default config matches the version loaded.

Of course, we don't already have that code in-place, so downgrading to any current release would not do that work, but it would be worthwhile pushing that out (nowish) as a micro-update just so this version's OTA work would be more robust.

internal/ota/ota.go
```diff
@@ -0,0 +1,432 @@
...
+		// it means that healthCheckUrl will be called after 5 seconds that we send willReboot JSONRPC event
+		// so we need to reboot it within 5 seconds to avoid it being called before the device is rebooted
+		if err := s.reboot(true, postRebootAction, 3*time.Second); err != nil {
+			return s.componentUpdateError("Error requesting reboot", err, &scopedLogger)
+		}
+	}
+
+	// We don't need set the updating flag to false here. Either it will;
+	// - set to false by the componentUpdateError function
+	// - device will reboot
```
IDisposable, line 292: This is not strictly true, when the `if s.rebootNeeded {` is not taken, then the flag is still set and we didn't error, nor reboot.

internal/ota/ota.go
```diff
@@ -0,0 +1,432 @@
...
+	logger.Trace().
+		Str("appVersionLocal", appVersionLocal.String()).
+		Str("systemVersionLocal", systemVersionLocal.String()).
+		Dur("duration", time.Since(t)).
+		Msg("checkUpdateStatus: getLocalVersion")
+
+	// fetch the remote metadata
+	remoteMetadata, err := s.fetchUpdateMetadata(ctx, params)
+	if err != nil {
+		if err == ErrVersionNotFound || errors.Unwrap(err) == ErrVersionNotFound {
```
IDisposable, line 369: `if errors.Is(err, ErrVersionNotFound)` does the (if needed unwrap) test...

adamshiervani (approved):

IDisposable:
I apologize... I never clicked submit on these comments. This issue still concerns me as-merged https://github.com/jetkvm/kvm/pull/912#discussion_r2539467737
@ym

ym:
> I apologize... I never clicked submit on these comments. This issue still concerns me as-merged [#912 (comment)](https://github.com/jetkvm/kvm/pull/912#discussion_r2539467737) @ym

That's fine, I will open a new PR to resolve these issues.

PR #913, opened 2025-10-29
fix: Fix video quality 

adamshiervani:
We got reports that the video quality decreased in the last release.

The issue was no default stream quality factor, and therefore it was falling back on 0, which effectively puts the default stream in "Low" mode.

Additionally, when setting and getting the video quality stream, we didn't store the value in the config but updated a local quality variable that was used in the old implementation.

### DIFF
--- a/config.go
+++ b/config.go
@@ -177,7 +177,8 @@ func getDefaultConfig() Config {
 			_ = confparser.SetDefaultsAndValidate(c)
 			return c
 		}(),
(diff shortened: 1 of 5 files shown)


ym (approved):

adrianmeraz:
Thank goodness it was a bug, the artifacts were driving me crazy!

adamshiervani:
@adrianmeraz Until this is released, you can set it to low and then back to high again, and it should work.

PR #915, opened 2025-10-29
refactor: More robust handling of jsonrpc calls

adamshiervani:
**Problem**
- RPC calls timed out while waiting for WebRTC to connect
- Dangling promises continued after navigation/reconnection
- Stale data channels remained in store after cleanup

**Solution:**
- Separated RTC readiness wait (indefinite) from RPC call timeout (5s)
- Added channel instance tracking to detect stale connections
- `waitForRtcReady` now throws when connection changes or is removed.
- Added `setRpcDataChannel(null)` to cleanup effect

**Result:**
RPC calls can wait indefinitely for WebRTC while the actual request still has a timeout. Promises abort cleanly when navigating away or reconnecting.

### DIFF
--- a/ui/src/utils/jsonrpc.ts
+++ b/ui/src/utils/jsonrpc.ts
@@ -24,17 +24,47 @@ export interface JsonRpcCallResponse<T = unknown> {
 let rpcCallCounter = 0;
 
 // Helper: wait for RTC data channel to be ready
+// This waits indefinitely for the channel to be ready, only aborting via the signal
+// Throws if the channel instance changed while waiting (stale connection detected)
 async function waitForRtcReady(signal: AbortSignal): Promise<RTCDataChannel> {
   const pollInterval = 100;
+  let lastSeenChannel: RTCDataChannel | null = null;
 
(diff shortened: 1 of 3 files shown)


ym (approved):
Code LGTM but hasn't been tested yet

IDisposable (approved):

IDisposable (approved):

ui/src/utils/jsonrpc.ts
```diff
@@ -135,6 +164,10 @@ export async function callJsonRpc<T = unknown>(
       throw error instanceof Error
         ? error
         : new Error(`JSON-RPC call failed after ${timeout}ms`);
+    } finally {
+      if (timeoutId !== null) {
+        clearTimeout(timeoutId);
```
IDisposable, line 169: Would it be nice to `timeoutId = null` here?

adamshiervani: No need to, as it's declared in the for-loop scope and the timeout cleared.

Issue #702, opened 2025-07-31
Show remote public ip


animalillo:
### A note for the community

> [!NOTE]
> Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.

### Disclaimer

- [x] I have read and understood the disclaimer.
- [ ] I plan to implement the feature myself.

### Subsystem

None

### Feature description

It would be nice to have the public ip of the remote device, i have used the stun info in the past to get the remote public ip of the system. But an integrated option would be a nice to have and some how the controlled host is not responding.

It would be fine if it's hidden on a config menu, like the current ip info screen. I just miss this information option as it would make my life a little bit easy on some edge cases.

PR #944, opened 2025-11-04
feat: add remote IP address to connection stats sidebar

ym:
Implements [702](https://github.com/jetkvm/kvm/issues/702)


pkg/myip/ip.go
```diff
@@ -0,0 +1,209 @@
...
+func (ps *PublicIPState) SetIPv4AndIPv6(ipv4, ipv6 bool) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	ps.ipv4 = ipv4
+	ps.ipv6 = ipv6
+}
+
+// SetIPv4 sets if we need to track IPv4 public IP addresses
+func (ps *PublicIPState) SetIPv4(ipv4 bool) {
```
IDisposable: These extra methods seem excessive, could just call the SetIPv4AndIPv6 or not have that method and just call both of these as needed

pkg/myip/ip.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// Stop stops the timer loop
+func (ps *PublicIPState) Stop() {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	if ps.cancel != nil {
+		ps.cancel()
+	}
```
IDisposable, line 148: Shouldn't we set cancel to nil here since it's not going to just immediately get reassigned as in Start?

ui/localization/messages/de.json
```diff
@@ -561,16 +561,16 @@
...
     "mouse_jiggler_error_config": "Beim Festlegen der Jiggler-Konfiguration ist ein Fehler aufgetreten",
     "mouse_jiggler_failed_state": "Jiggler-Status konnte nicht festgelegt werden: {error}",
-    "mouse_jiggler_frequent": "Häufig – alle 30 Sek.",
+    "mouse_jiggler_frequent": "Häufig - 30er",
     "mouse_jiggler_invalid_cron": "Ungültiger Cron-Ausdruck. Bitte überprüfen Sie Ihr Zeitplanformat (z. B. „0 * * * * *“ für jede Minute).",
-    "mouse_jiggler_light": "Leicht – 5 Min",
-    "mouse_jiggler_standard": "Standard – 1 Min",
-    "mouse_jiggler_title": "Jiggler",
+    "mouse_jiggler_light": "Licht - 5m",
+    "mouse_jiggler_standard": "Standard - 1 m",
```
IDisposable: Why is 5 minutes "5m", but 1 minute is "1 m", seems off?

ui/localization/messages/zh.json
```diff
@@ -897,5 +891,11 @@
     "wake_on_lan_invalid_mac": "无效的 MAC 地址",
     "wake_on_lan_magic_sent_success": "魔包发送成功",
     "welcome_to_jetkvm": "欢迎来到 JetKVM",
-    "welcome_to_jetkvm_description": "远程控制任何计算机"
+    "welcome_to_jetkvm_description": "远程控制任何计算机",
+    "video_overlay_reboot_device_is_rebooting": "设备正在重启",
```
IDisposable: Most other language files have these video_overlay_reboot_* messages deleted in this PR, but in Chinese were adding them? Is this a merge error or an incomplete deletion?

ui/src/components/sidebar/connectionStats.tsx
```diff
@@ -106,6 +117,28 @@ export default function ConnectionStatsSidebar() {
...
+                      {m.connection_stats_remote_ip_address()}
+                    </div>
+                    <div className="flex items-center">
+                      <GridCard cardClassName="rounded-r-none">
+                        <div className="h-[34px] flex items-center text-xs select-all text-black font-mono dark:text-white px-3 ">
+                          {remoteIPAddress}
+                        </div>
+                      </GridCard>
+                      <Button className="rounded-l-none border-l-slate-800/30 dark:border-slate-300/20" size="SM" type="button" theme="light" LeadingIcon={LuCopy} onClick={async () => {
+                        const ip = remoteIPAddress || "";
```
IDisposable: Don't we know remoteIPAddress is not null/undefined because we're inside the conditional block at 120?

ui/localization/messages/zh.json
ym: It appears to be a merge error. Fixing ..

pkg/myip/ip.go
ym: Cancel function can be called multiple times and won't panic, but yes, we should set it to `nil` after it's cancelled.

pkg/myip/ip.go
ym: I would prefer to leave `SetIPv4` and `SetIPv6` here for future use, since `SetIPv4AndIPv6` only accepts a boolean value. This means that we cannot use the method to set IPv4 or IPv6 alone.

IDisposable: I meant just have `SetIPv4AndIPv6` take two booleans, (for IPv4 and IPv6)... but it's fine either way, just trying to keep the code size optimal.

ui/src/components/PublicIPCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const refreshPublicIPs = useCallback(() => {
+    send("getPublicIPAddresses", { refresh: true }, (resp: JsonRpcResponse) => {
+      setPublicIPs([]);
+      if ("error" in resp) {
+        notifications.error(m.public_ip_card_refresh_error({ error: resp.error.data || m.unknown_error() }));
+        return;
+      }
+      const publicIPs = resp.result as PublicIP[];
+      setPublicIPs(publicIPs.sort(({ ip: aIp }, { ip: bIp }) => {
+        const aIsIPv6 = aIp.includes(":");
```
IDisposable, line 44: Add comment
// sort IPv6 addresses to after the IPv4 addresses

ui/src/components/PublicIPCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+                </div>
+              </div>
+            </div>
+          ) : (
+            <div className="flex gap-x-6 gap-y-2">
+              <div className="flex-1 space-y-2">
+                {publicIPs?.map(ip => (
+                  <div key={ip.ip} className="flex justify-between border-slate-800/10 pt-2 dark:border-slate-300/20">
+                    <span className="text-sm font-medium">
+                      {ip.ip}
```
IDisposable, line 94: Would it be nice to show the `ip.last_updated` as well?

pkg/myip/check.go
```diff
@@ -0,0 +1,160 @@
...
+			continue
+		}
+		values[key] = value
+	}
+
+	ps.lastUpdated = time.Now()
+	if ts, ok := values["ts"]; ok {
+		if ts, err := strconv.ParseFloat(ts, 64); err == nil {
+			ps.lastUpdated = time.Unix(int64(ts), 0)
+		}
```
IDisposable, line 69: We silently discard any failure to parse `ts` but that's probably okay...

IDisposable (approved):

PR #946, opened 2025-11-05
fix: await sleep needs to be called inside async function

ym:
https://github.com/jetkvm/kvm/pull/916


IDisposable (approved):
Sorry about the bad redo of this.

PR #947, opened 2025-11-05
feat: reset usb when needed

### LINKED ISSUE #244: unmounting disc from web ui doesnt unmount from mac
I have a mac mini, when testing using both stable and dev channels,

i can mount an ISO as expected, but then when you say Unmount,

it says it unmounted successfully but in fact it doesnt actually unmount its still mounted on the mac

i ran a custom build on my jetkvm and console is showing the error
`Remove Mass Storage Image Error failed to set image path: write /sys/kernel/config/usb_gadget/jetkvm/functions/mass_storage.usb0/lun.0/file: device or resource busy`

also side note: when you mount an iso it doesnt actually close the window, it stays open, thats a GUI bug i think

ym:
Fixes https://github.com/jetkvm/kvm/issues/244

### DIFF
--- a/internal/usbgadget/config.go
+++ b/internal/usbgadget/config.go
@@ -24,6 +24,14 @@ type gadgetConfigItemWithKey struct {
 
 type orderedGadgetConfigItems []gadgetConfigItemWithKey
 
+type UsbResetMode uint8
+
+const (
+	UsbResetNever    UsbResetMode = 0 // never reset the usb gadget
+	UsbResetAlways   UsbResetMode = 1 // always reset the usb gadget
+	UsbResetOnDemand UsbResetMode = 2 // reset the usb gadget when needed
+)
+
 var defaultGadgetConfig = map[string]gadgetConfigItem{
 	"base": {
 		order: 0,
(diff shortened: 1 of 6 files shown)


internal/usbgadget/config.go
```diff
@@ -177,36 +185,57 @@ func (u *UsbGadget) Init() error {
...
+	err := u.WithTransaction(f(resetUsbMode == UsbResetAlways))
+	if err == nil {
 		return nil
-	})
+	}
+
+	// if the initial attempt failed, try to configure the gadget again with the resetUsb flag
+	if resetUsbMode == UsbResetOnDemand {
+		u.log.Warn().Err(err).Msg("initial attempt to configure the gadget failed, resetting USB gadget then trying again")
+		// NOTES: do not use the RebindUsb method here because it will block the transaction
```
IDisposable, line 232: ...do not use the _public_ RebindUsb method... 

I completely misread this the first time ;)

IDisposable (approved):

adamshiervani:
Unmounting an mounted image breaks the USB gadget, so there is no mouse or keyboard after unmounting. 

1. Mount any HTTP image
2. Keyboard & Mouse works.
3. Unmount image
4. Keyboard & Mouse doesn't work.

Rebooting is the only thing that solves the problem

PR #948, opened 2025-11-05
feat: lldp 

ym:
e.g. 

image


internal/lldp/afpacket.go
```diff
@@ -0,0 +1,96 @@
...
+) (
+	frameSize int,
+	blockSize int,
+	numBlocks int,
+	err error,
+) {
+	if snaplen < pageSize {
+		frameSize = pageSize / (pageSize / snaplen)
+	} else {
+		frameSize = (snaplen/pageSize + 1) * pageSize
```
IDisposable: Should there be parens around the `pageSize + 1` or for clarification that we DON'T want that, around the `snaplen / pageSize` so it's clear we want one-extra pageSize?  

I am not sure what's going on here to be honest. What are typical/expected values for `pageSize` and `snaplen` and what are we wanting the `frameSize` to reflect?

It "looks" like we want a frameSize to match either:

1. pageSize divided by the number of snaps in a page 
    - pageSize is 4K
    - snaplen is 512b
    - frameSize = 4096 / (4096 / 512) = 4096/8 = 512b
2. (A) ((snaplen / pageSize) + 1) * pageSize, for example:
    - pageSize is 1K 
    - snaplen is 2K
    - frameSize would be (2048/1024 + 1) * 1024 = (2 + 1) * 1024 = 3072 which seems odd
2. (B) (snaplen / (pageSize + 1)) * pageSize
    - pageSize is 1K 
    - snaplen is 2K
    - frameSize would be (2048/(1024 + 1)) * 1024 = (2048 / 1025) * 1024 = 
    - 1.998 * 1024 = 2046 (or  which seems odd)
    - 1 * 1024 = 2048 (assuming integer math).

Given this, the block _seems_ to be desiring the small of pageSize or snaplen... but perhaps that's because I have really bad assumptions about expected data

internal/lldp/rx.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	logger.Info().Msg("created TPacket")
+
+	// set up multicast addresses
+	// otherwise the kernel might discard the packets
+	// another workaround would be to enable promiscuous mode but that's too tricky
+	for _, mac := range multicastAddrs {
+		hwAddr, err := net.ParseMAC(mac)
+		if err != nil {
+			logger.Error().Msgf("unable to parse MAC address %s: %s", mac, err)
```
IDisposable: Shouldn't the `err` be using `%w` for the format specifier all through this module?

internal/lldp/rx.go
```diff
@@ -0,0 +1,235 @@
...
+			ttl = time.Duration(n.TTL) * time.Second
+			n.Values["ttl"] = fmt.Sprintf("%d", n.TTL)
+		case layers.LLDPTLVOrgSpecific:
+			for _, org := range info.OrgTLVs {
+				n.Values[fmt.Sprintf("org_specific_%d", org.OUI)] = string(org.Info)
+			}
+		}
+	}
+
+	if gotEnd || ttl < 1*time.Second {
```
IDisposable: The test of `gotEnd` seems backwards... shouldn't we be ignoring incomplete TLV sets (e.g. no _LLDPTLVEnd_ seen), not discarding the ones that we did get a complete value set for?

ui/src/components/LLDPNeigh.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <div className={cx("flex flex-col justify-between", className)}>
+      <span className="text-sm text-slate-600 dark:text-slate-400">
+        {label}
+      </span>
+      <span className="text-sm font-medium">{value}</span>
+    </div>
+  );
+}
+
+export default function LLDPNeighCard({
```
IDisposable: s/LLDPNeighCard/LLDPNeighborCard ?

network.go
```diff
@@ -161,6 +163,18 @@ func initNetwork() error {
 
 	networkManager = nm
 
+	lldpService = lldp.NewLLDP(&lldp.LLDPOptions{
+		InterfaceName: NetIfName,
+		EnableRx:      nc.LLDPMode.String != "disabled",
```
IDisposable: Shouldn't this care about the tx_only and rx_only settings https://github.com/jetkvm/kvm/pull/948/files#diff-461417865cac9c4c5ea6a683e5f9f810e9a3f6cff4c8da5cd726ec6896dde346R764

internal/lldp/rx.go
ym: We should use `logger.Err` instead of converting the error message manually. Fixed.

ui/src/components/LLDPNeigh.tsx
ym: Yea, I shortened it to Neigh since the original felt long, but consistency is important. I’ll rename it to `LLDPNeighborCard`.

ym:
I can't seem to fix the segfault issue when closing PacketSource, and I can see that other people are having the same issue with GoPacket. I will move the Rx process to a separate process to avoid this issue.

internal/lldp/neigh.go
```diff
@@ -1,27 +1,47 @@
...
+	PortDescription   string             `json:"port_description"`
+	SystemName        string             `json:"system_name"`
+	SystemDescription string             `json:"system_description"`
+	TTL               uint16             `json:"ttl"`
+	ManagementAddress *ManagementAddress `json:"management_address,omitempty"`
+	Capabilities      []string           `json:"capabilities"`
+	Values            map[string]string  `json:"values"`
+}
+
+func (n *Neighbor) cacheKey() string {
```
IDisposable: This does a Sprintf every time, much cheaper to do this, which is almost zero-allocating:

```go
type NeighborCacheKey struct {
    Mac    string
    Source string
}

func (n *Neighbor) cacheKey() NeighborCacheKey {
   return NeighborCacheKey {Mac: n.Mac, Source: item.Source}
}
```

internal/lldp/lldp.go
```diff
@@ -46,29 +69,77 @@ func NewLLDP(opts *LLDPOptions) *LLDP {
...
-		enableRx:      opts.EnableRx,
-		enableTx:      opts.EnableTx,
-		l:             opts.Logger,
-		neighbors:     ttlcache.New(ttlcache.WithTTL[string, Neighbor](1 * time.Hour)),
+		interfaceName:    opts.InterfaceName,
+		advertiseOptions: opts.AdvertiseOptions,
+		enableRx:         opts.EnableRx,
+		enableTx:         opts.EnableTx,
+		l:                opts.Logger,
+		neighbors:        ttlcache.New(ttlcache.WithTTL[string, Neighbor](1 * time.Hour)),
```
IDisposable: If you do [this suggestion](https://github.com/jetkvm/kvm/pull/948/commits/fd44ff49fd9b637ea88b533bb6f98b07ee2b9df4#r2500065672) then the cache is ...`.WithTTL[NeighborCacheKey, Neighbor]`

internal/lldp/neigh.go
```diff
@@ -1,27 +1,47 @@
...
+	return fmt.Sprintf("%s-%s", n.Mac, n.Source)
 }
 
-func (l *LLDP) addNeighbor(mac string, neighbor Neighbor, ttl time.Duration) {
+func (l *LLDP) addNeighbor(neighbor *Neighbor, ttl time.Duration) {
 	logger := l.l.With().
-		Str("mac", mac).
+		Str("mac", neighbor.Mac).
```
IDisposable, line 63: maybe also include `Str("source", neighbor.Source).` since that's half of the key...

internal/lldp/neigh.go
```diff
@@ -51,5 +71,10 @@ func (l *LLDP) GetNeighbors() []Neighbor {
 		neighbors = append(neighbors, item.Value())
 	}
 
+	// sort based on MAC address
+	sort.Slice(neighbors, func(i, j int) bool {
```
IDisposable: would `SliceStable(` be better here, and should we also consider `Source`?

internal/lldp/rx.go
```diff
@@ -27,65 +28,122 @@ var multicastAddrs = []string{
...
 	logger := l.l.With().Str("interface", l.interfaceName).Logger()
-	tPacket, err := afPacketNewTPacket(l.interfaceName)
+	tPacketRx, err := afPacketNewTPacket(l.interfaceName)
 	if err != nil {
 		return err
 	}
-	logger.Info().Msg("created TPacket")
+	logger.Info().Msg("created TPacketRx")
+
+	// Double-check: another goroutine might have set it up while we were creating
```
IDisposable: Aren't we under a mutex via line 33?

internal/lldp/rx.go
```diff
@@ -27,65 +28,122 @@ var multicastAddrs = []string{
...
+			Err(err).
+			Msg("unable to set BPF filter")
+		tPacketRx.Close()
 		return err
 	}
 	logger.Info().Msg("BPF filter set")
 
-	l.pktSource = gopacket.NewPacketSource(tPacket, layers.LayerTypeEthernet)
-	l.tPacket = tPacket
+	l.pktSourceRx = gopacket.NewPacketSource(tPacketRx, layers.LayerTypeEthernet)
```
IDisposable: Would it be better to assign these in the opposite order since we're testing the `l.tPacketRx` at the top of this function?

internal/lldp/rx.go
```diff
@@ -27,65 +28,122 @@ var multicastAddrs = []string{
...
 }
 
+func (l *LLDP) doCapture(logger *zerolog.Logger, rxCtx context.Context) {
+	defer func() {
+		l.mu.Lock()
+		l.rxRunning = false
+		l.mu.Unlock()
+	}()
+
+	packetChan := l.pktSourceRx.Packets()
```
IDisposable: Do we want to verify that `l.pktSourceRx` and `l.tPacketRx` are not `nil` before starting this loop, or if they get set to that while it's running?

internal/lldp/rx.go
```diff
@@ -27,65 +28,122 @@ var multicastAddrs = []string{
...
 	return nil
 }
 
+func (l *LLDP) doCapture(logger *zerolog.Logger, rxCtx context.Context) {
+	defer func() {
+		l.mu.Lock()
+		l.rxRunning = false
+		l.mu.Unlock()
+	}()
+
```
IDisposable, line 94: should this function be setting `l.rxRunning = true` here?  That's what we do in the TX code, e.g. this gets inserted above line 95...

```go
l.mu.Lock()
l.rxRunning = true
l.mu.Unlock()
```

internal/lldp/rx.go
```diff
@@ -213,23 +306,61 @@ func (l *LLDP) handlePacketCDP(mac string, raw *layers.CiscoDiscovery, info *lay
...
 	}
 
-	if l.pktSource != nil {
-		l.pktSource = nil
+	if l.tPacketRx != nil {
+		l.tPacketRx.Close()
+		l.tPacketRx = nil
 	}
 
+	if l.pktSourceRx != nil {
```
IDisposable, line 397: Should we `nil`-out in the opposite order?

internal/lldp/rx.go
```diff
@@ -213,23 +306,61 @@ func (l *LLDP) handlePacketCDP(mac string, raw *layers.CiscoDiscovery, info *lay
...
+		l.pktSourceRx = nil
+	}
+
+	time.Sleep(100 * time.Millisecond)
```
IDisposable: ... or if not, perhaps this sleep should be moved up to before we start closing and `nil`ing things out so in-flight stuff has a chance to notice and cancel the loop?

internal/lldp/tx.go
```diff
@@ -0,0 +1,270 @@
...
+		"cvlan":        layers.LLDPCapsCVLAN,
+		"svlan":        layers.LLDPCapsSVLAN,
+		"tmpr":         layers.LLDPCapsTmpr,
+	}
+)
+
+func toLLDPCapabilitiesBytes(capabilities []string) uint16 {
+	r := uint16(0)
+	for _, capability := range capabilities {
+		if _, ok := capabilityMap[capability]; !ok {
```
IDisposable: if we do this instead, we don't do the map lookup twice
```go
if mask, ok := capabilityMap[capability]; ok {
    r |= mask
}
```

internal/lldp/lldp.go
```diff
@@ -46,29 +69,77 @@ func NewLLDP(opts *LLDPOptions) *LLDP {
...
 		}
 	}
 
-	go l.neighbors.Start()
+	return nil
+}
+
+// StartRx starts the LLDP receiver if not already running
+func (l *LLDP) startRx() error {
+	l.mu.Lock()
```
IDisposable: These are under a `RLock/RUnlock` pair in the corresponding TX code (and well as grabbing the cancel, but that's later down)

internal/lldp/tx.go
```diff
@@ -0,0 +1,270 @@
...
+
+	if running || !enabled {
+		return nil
+	}
+
+	if cancel != nil {
+		cancel()
+	}
+
+	l.txCtx, l.txCancel = context.WithCancel(context.Background())
```
IDisposable: This code mutates the LLDP object, shouldn't it be under an `l.mu.Lock()` wit the corresponding `l.mu.Unlock()` just before like 239 (the doSend call)?

network.go
```diff
@@ -178,6 +187,15 @@ func initNetwork() error {
 	return nil
 }
 
+func toLLDPSysDescription(nc *types.NetworkConfig) string {
+	systemVersion, appVersion, err := GetLocalVersion()
+	if err == nil {
+		return fmt.Sprintf("JetKVM (app: %s)", GetBuiltAppVersion())
+	}
+
+	return fmt.Sprintf("JetKVM (app: %s, system: %s)", appVersion.String(), systemVersion.String())
```
IDisposable, line 249: DeviceID too?

IDisposable:
> I can't seem to fix the segfault issue when closing PacketSource, and I can see that other people are having the same issue with GoPacket. I will move the Rx process to a separate process to avoid this issue.

I would hate for us to spin up a process just for that! Can I take a gander and see if I can figure it out? There are some mutex issues I noted, but I haven't even tries to run this code... I might be able to find the issue once I do?

Can you give me the steps to make it segfault and/or the stack trace?

internal/network/types/config.go
```diff
@@ -42,7 +42,7 @@ type NetworkConfig struct {
 	IPv6Mode   null.String       `json:"ipv6_mode,omitempty" one_of:"slaac,dhcpv6,slaac_and_dhcpv6,static,link_local,disabled" default:"slaac"`
 	IPv6Static *IPv6StaticConfig `json:"ipv6_static,omitempty" required_if:"IPv6Mode=static"`
 
-	LLDPMode                null.String `json:"lldp_mode,omitempty" one_of:"disabled,basic,all" default:"basic"`
+	LLDPMode                null.String `json:"lldp_mode,omitempty" one_of:"disabled,rx_only,tx_only,rx_and_tx,basic,all" default:"rx_and_tx"`
```
IDisposable: Should we remove `,basic,all` here and shouldn't the default now be `rx_only` per comment about privacy concerns on line 57?

If we're now only taking the list of options in stores.ts line 64, do we need a config conversion that morphs basic and all to to rx_only and rx_and_tx (respectively), or do we not care, since the LLDP was never really implemented?

ym:
> Can you give me the steps to make it segfault and/or the stack trace?

I kinda fixed it, but the solution is quite dirty. If you come up with a cleaner approach, go for it.
I added the details in the latest commit message.

IDisposable:
> > Can you give me the steps to make it segfault and/or the stack trace?
> 
> I kinda fixed it, but the solution is quite dirty. If you come up with a cleaner approach, go for it. I added the details in the latest commit message.

Looks fine by me :)

ym:
<del>I reverted the change to cacheKey, since ttlcache doesn’t handle struct keys properly. It seems to use the memory address (or another non-stable reference) as the key, which causes unexpected behavior.</del>

nvm, it's stupid React key issue :-)

internal/network/types/config.go
ym: It's kept here for backwards compatibility, if the value doesn't exist in `one_off`, an error will be thrown :-(

network.go
ym: DeviceID is sent in ChassisID TLV.

internal/lldp/rx.go
ym: Removed as no longer needed

internal/lldp/rx.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	logger := l.l.With().Str("interface", l.interfaceName).Logger()
+	logger.Info().Msg("starting capture LLDP ethernet frames")
+
+	// Create a new context for this instance
+	l.rxCtx, l.rxCancel = context.WithCancel(context.Background())
+	l.rxRunning = true
+
+	go l.doCapture(&logger)
```
IDisposable: Doesn't this mean  that the go routine is running _inside_ the lock from line 140? Shouldn't we put all the "setup" work in a distinct function and run that all under a `defer` unlock, call that setup, and then start the `go` routine?

internal/lldp/tx.go
```diff
@@ -0,0 +1,239 @@
...
+		ComputeChecksums: true,
+	}, &ethFrame, &lldpFrame); err != nil {
+		l.l.Error().Err(err).Msg("unable to serialize packet")
+		return err
+	}
+
+	logger.Trace().Hex("packet", buf.Bytes()).Msg("sending LLDP packet")
+
+	// send packet
+	if err := l.tPacketTx.WritePacketData(buf.Bytes()); err != nil {
```
IDisposable: Seems like we shouldn't do I/O under the read-lock from line 95, but we are reaching down for the `tPacketTx`... can/should we split that out to ensure we do the packet write outside the lock?

internal/lldp/tx.go
```diff
@@ -0,0 +1,239 @@
...
+	enabled := l.enableTx
+	cancel := l.txCancel
+	l.mu.RUnlock()
+
+	if running || !enabled {
+		return nil
+	}
+
+	if cancel != nil {
+		cancel()
```
IDisposable: _Note to self: This cancels any **previously** running Tx loop. This code is idiomatically correct._

internal/network/types/config.go
IDisposable: Should we morph it over as we read it in (like I once did for en_US vs en-US?)

I did that in #961

internal/lldp/rx.go
IDisposable: I created a  PR #961 targeting your `feat/lldp` branch with a safer implementation of this code (in tx.go, rx.go, lldp.go) to make the locks and pointer ownership clearer.

internal/lldp/tx.go
IDisposable: ... also addressed this concern in the restructure I propose in #961

internal/lldp/tx.go
IDisposable: ... made clearer in #961

IDisposable (changes requested):

PR #952, opened 2025-11-07
feat: failsafe mode

ym:
The feature will be ported to `dev` later once this PR is merged


internal/native/native.go
```diff
@@ -92,6 +95,12 @@ func NewNative(opts NativeOptions) *Native {
 }
 
 func (n *Native) Start() {
+	if n.disable {
```
IDisposable: Why would we want to disable the native?

ui/src/components/FailSafeModeOverlay.tsx
```diff
@@ -0,0 +1,211 @@
...
+      const filename = `jetkvm-recovery-${reason}-${timestamp}.txt`;
+
+      const blob = new Blob([logContent], { type: "text/plain" });
+      const url = URL.createObjectURL(blob);
+      const a = document.createElement("a");
+      a.href = url;
+      a.download = filename;
+      document.body.appendChild(a);
+      a.click();
+      document.body.removeChild(a);
```
IDisposable, line 114: This is the normal pattern for forcing a download. I have had some issues in other apps doing the `removeChild(` and `revokeObjectURL(` immediately after the `.click()` (depending on browser and version and completion of download timing). I now just wrap those two in a `setTimeout(` for 1000 ms or so and the problems never happen.

ui/src/components/FailSafeModeOverlay.tsx
```diff
@@ -0,0 +1,211 @@
...
+**Timestamp:** ${new Date().toISOString()}
+**App Version:** ${appVersion || "Unknown"}
+**System Version:** ${systemVersion || "Unknown"}
+
+## Logs
+Please attach the recovery logs file that was downloaded to your computer:
+\`${filename}\`
+
+> [!NOTE]
+> Please omit any sensitive information from the logs.
```
IDisposable: s/omit/remove (we need to be clearer that _they_ need to do that edit... especially because if we're attaching a file to a bug report, those are publicly visible.

ui/src/components/FailSafeModeOverlay.tsx
```diff
@@ -0,0 +1,211 @@
...
+        `https://github.com/jetkvm/kvm/issues/new?` +
+        `title=${encodeURIComponent(`Recovery Mode: ${reason} process issue`)}&` +
+        `body=${encodeURIComponent(issueBody)}`;
+
+      window.open(issueUrl, "_blank");
+    });
+  };
+
+  const handleDowngrade = () => {
+    navigateTo("/settings/general/update?app=0.4.8");
```
IDisposable: Is this something we want to drive from a build-time setting instead of hard-coding in the code (and eventually forgetting or too-aggressively changing)?

ui/src/components/FailSafeModeOverlay.tsx
```diff
@@ -0,0 +1,211 @@
...
+                        disabled={!hasDownloadedLogs}
+                      />
+                    </Tooltip>
+
+                    <Tooltip text="Download logs first to unlock" show={!hasDownloadedLogs}>
+                      <Button
+                        size="SM"
+                        onClick={handleDowngrade}
+                        theme="light"
+                        text="Downgrade to v0.4.8"
```
IDisposable: Again, should the downgrade version be a build-time setting?

ui/src/hooks/stores.ts
```diff
@@ -926,3 +926,16 @@ export const useMacrosStore = create<MacrosState>((set, get) => ({
     }
   }
 }));
+
+export interface FailsafeModeState {
+  isFailsafeMode: boolean;
+  reason: string | null; // "video", "network", etc.
+  setFailsafeMode: (active: boolean, reason: string | null) => void;
```
IDisposable: Would we ever actually _set_ a failsafe with a `null` _reason_?

Seems like we would only set it to some string, so the state setter could/should be 
`setFailsafeMode: (active: boolean, reason: string) => void;`

ui/src/routes/devices.$id.settings.general.reboot.tsx
```diff
@@ -55,11 +72,16 @@ function ConfirmationBox({
...
-          <Button size="SM" theme="light" text="Yes" onClick={onYes} />
-          <Button size="SM" theme="blank" text="No" onClick={onNo} />
-        </div>
+        {isRebooting ? (
+          <div className="mt-4 flex items-center justify-center">
+            <LoadingSpinner className="h-6 w-6 text-blue-700 dark:text-blue-500" />
+          </div>
+        ) : (
+          <div className="mt-4 flex gap-x-2">
+            <Button size="SM" theme="light" text="Yes" onClick={onYes} />
```
IDisposable, line 84: localize with `text={m.Yes()}`

ui/src/routes/devices.$id.settings.general.reboot.tsx
```diff
@@ -55,11 +72,16 @@ function ConfirmationBox({
...
+            <Button size="SM" theme="blank" text="No" onClick={onNo} />
```
IDisposable, line 85: localize with `text={m.No()}`

ui/src/routes/devices.$id.tsx
```diff
@@ -113,6 +115,7 @@ const loader: LoaderFunction = ({ params }: LoaderFunctionArgs) => {
   return import.meta.env.MODE === "device" ? deviceLoader() : cloudLoader(params);
 };
 
+
```
IDisposable, line 118: _nit_ excess blank line

ui/src/routes/devices.$id.tsx
```diff
@@ -841,13 +853,15 @@ export default function KvmIdRoute() {
           />
 
           <div className="relative flex h-full w-full overflow-hidden">
-            <WebRTCVideo />
+            {!isFailsafeMode && <WebRTCVideo />}
```
IDisposable: Shouldn't this be only disabled if `isFailsafeMode` is true **and** the reason is video?

failsafe.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+)
+
+const (
+	failsafeDefaultLastCrashPath = "/userdata/jetkvm/crashdump/last-crash.log"
```
IDisposable, line 11: These constants parallel some of the ones in _main.go_, perhaps we should coalesce them?

failsafe.go
```diff
@@ -0,0 +1,107 @@
...
+			failsafeModeActive = true
+			failsafeModeReason = "failsafe_env_set"
+			return
+		}
+
+		// check if the failsafe file exists
+		if _, err := os.Stat(failsafeFile); err == nil {
+			failsafeModeActive = true
+			failsafeModeReason = "failsafe_file_exists"
+			_ = os.Remove(failsafeFile)
```
IDisposable, line 43: So the intent is to use the flag as a one-shot and delete after it's seen once?

failsafe.go
```diff
@@ -0,0 +1,107 @@
...
+		// open the last crash log file and find if it contains the string "panic"
+		content, err := os.ReadFile(lastCrashPath)
+		if err != nil {
+			l.Warn().Err(err).Msg("failed to read last crash log")
+			return
+		}
+
+		// unlink the last crash log file
+		failsafeCrashLog = string(content)
+		_ = os.Remove(lastCrashPath)
```
IDisposable, line 77: Should we delay trashing the crashlog until we've exited the failsafe mode through some action?

main.go
```diff
@@ -58,7 +64,6 @@ func Main() {
 	// Initialize mDNS
 	if err := initMdns(); err != nil {
 		logger.Error().Err(err).Msg("failed to initialize mDNS")
-		os.Exit(1)
 	}
```
IDisposable, line 67: // TODO: disable/reset mDNS option?

internal/native/cgo_linux.go
```diff
@@ -46,7 +49,17 @@ static inline void jetkvm_cgo_setup_rpc_handler() {
 */
 import "C"
 
-var cgoLock sync.Mutex
+var (
+	cgoLock     sync.Mutex
+	cgoDisabled bool
```
IDisposable, line 54: Seems like this is really `videoDisabled` so we could (perhaps), later disable mount or audio or whatever...  maybe make that rename happen _now_ before we conflate it with other disabled sub-systems.

IDisposable (changes requested):

failsafe.go
ym: Yes, otherwise it will enable fail safe mode every time you reboot. The actual log file will be preserved and we only delete the symbolic link.

IDisposable: Makes sense, thanks.

internal/native/cgo_linux.go
ym: It's a good idea, but to keep this PR simple, I'd prefer not to introduce it here.

main.go
ym: We perform a non-nil check for each call to mdns, so it is not necessary to do so now.

ui/src/hooks/stores.ts
ym: `reason` would be null if `isFailsafeMode` is `false`

failsafe.go
ym: `main.go` uses a different way to compose the file path so I would prefer managing them individually, plus we also have `JETKVM_LAST_ERROR_PATH` to allow the supervisor to pass a custom path to override the built-in one.

ui/src/hooks/stores.ts
IDisposable: Initially, yes, but why would we set it to `null` later? Wouldn't we set it to `""` (e.g. empty but not null string)?

main.go
IDisposable: I actually meant (and communicated very poorly, sorry), do we need to _reset to default_ the mDNS configuration or disable mDNS here, like we indicated we need eventually `// TODO` on line 56.

ui/src/routes/devices.$id.settings.general.reboot.tsx
adamshiervani: As this is a patch on top of the `0.4.9` we don't have any i18n yet.

ui/src/hooks/stores.ts
ym: @adamshiervani any idea?

adamshiervani: Go will always return a string, either empty or some content - let's reflect that in the FE types too.

ui/src/routes/devices.$id.settings.general.reboot.tsx
IDisposable: Ahh, sorry... I read the comment about this eventually going to `dev`, but didn't grok that this one was headed at 0.4.x instead. Sorry.

ui/src/routes/devices.$id.tsx
```diff
@@ -672,9 +672,10 @@ export default function KvmIdRoute() {
     }
 
     if (resp.method === "failsafeMode") {
+      console.log("Setting failsafe mode", resp.params);
       const { active, reason } = resp.params as { active: boolean; reason: string };
       console.debug("Setting failsafe mode", { active, reason });
-      setFailsafeMode(active, reason);
+      // setFailsafeMode(active, reason);
```
IDisposable: ~Should this still be commented-out? If so, we need to delete line 610 (and probably more).~ that's what I get for reviewing interim commits :)

IDisposable (approved):

ui/src/routes/devices.$id.tsx
adamshiervani: 🤦🏼 No, that's a relic of my testing. Reverted!

adamshiervani (approved):

PR #959, opened 2025-11-10
feat: failsafe mode (#952) dev branch

ym:


IDisposable (approved):
I fixed the merge issue that caused the UI lint to fail and pushed to this branch, make sure you pull before doing more work.

Assuming the 4.x version was good, this one is too ;)

There's some _nit_ level issues with extra lines or spaces, but not worth holding this up.

PR #960, opened 2025-11-10
Add localization of new elements

IDisposable:
**This merges into your `feat/lldp` branch**

Still need to machine_translate the other languages, but waiting until we're ready to merge (or after) to do that.


IDisposable:
This was included in #961 which has been merged, so closing

PR #961, opened 2025-11-11
Restructures LLDP service for safer state management

IDisposable:
**This targets your `feat/lldp` branch, and builds on PR #960 so merge that first or just this one**

Refactors the LLDP (Link Layer Discovery Protocol) service to improve state management, resource handling, and lifecycle control for its receive and transmit functionalities.

### Summary
*   **LLDP Service Refactoring:** Introduces a `RunningState` struct to encapsulate the operational state, context, logger, and packet resources for both LLDP RX and TX components independently. This change simplifies management, enhances resource cleanup, and clarifies concurrency handling by ensuring locks are properly released.
*   **Configuration Updates:** Handles LLDP mode options to ensure backward compatibility by remapping deprecated `enabled`, `all`, and `basic` modes to their modern equivalents in existing configurations using `rx_only`, `tx_only`, and `rx_and_tx`.
*   **Improved Error Handling:** Adds a global panic handler to ensure all uncaught panics are logged. Enhances error logging for network configuration operations, including hostname setting, DHCP client cleanup, and LLDP service initialization and updates.
*   **UI Localization:** Localizes the display of LLDP neighbor details in the user interface from PR #960


internal/confparser/confparser_test.go
```diff
@@ -39,7 +39,7 @@ type testNetworkConfig struct {
 	IPv6Mode   null.String           `json:"ipv6_mode" one_of:"slaac,dhcpv6,slaac_and_dhcpv6,static,link_local,disabled" default:"slaac"`
 	IPv6Static *testIPv6StaticConfig `json:"ipv6_static,omitempty" required_if:"IPv6Mode=static"`
 
-	LLDPMode                null.String `json:"lldp_mode,omitempty" one_of:"disabled,basic,all" default:"basic"`
+	LLDPMode                null.String `json:"lldp_mode,omitempty" one_of:"disabled,enabled,rx_only,tx_only,rx_and_tx,basic,all" default:"rx_and_tx"`
```
IDisposable, line 42: `enabled` seems to be what we default to in the original code as tested against a factory reset

internal/lldp/lldp.go
```diff
@@ -14,16 +14,18 @@ import (
 
 var defaultLogger = logging.GetSubsystemLogger("lldp")
 
+type RunningState struct {
```
IDisposable, line 17: Bundles up all the state for either Tx or Rx into a single object to ease ensuring proper running-state management.

internal/lldp/lldp.go
```diff
@@ -14,16 +14,18 @@ import (
...
+	Enabled bool
+	Running bool
+	Logger  *zerolog.Logger
```
IDisposable, line 20: Passing the Logger and TPacket as pointers to ensure we don't accidentally miss mutations.

internal/lldp/lldp.go
```diff
@@ -74,54 +70,34 @@ func NewLLDP(opts *Options) *LLDP {
 	return &LLDP{
 		interfaceName:    opts.InterfaceName,
 		advertiseOptions: opts.AdvertiseOptions,
-		enableRx:         opts.EnableRx,
-		enableTx:         opts.EnableTx,
-		rxWaitGroup:      &sync.WaitGroup{},
-		l:                opts.Logger,
 		neighbors:        make(map[neighborCacheKey]Neighbor),
 		onChange:         opts.OnChange,
+		Rx:               &RunningState{},
```
IDisposable, line 75: We don't care about the initial enabled state because we're going to call the same `SetRxAndTx` method as would be used when changing options. This gives a single path by which Rx/Tx is Start/Stop.

Also means we can eliminate the `Start()` method entirely

internal/lldp/lldp.go
```diff
@@ -74,54 +70,34 @@ func NewLLDP(opts *Options) *LLDP {
...
-	enabled := l.enableRx
+	running := l.Rx.Running
+	enabled := l.Rx.Enabled
 	l.mu.RUnlock()
 
 	if running || !enabled {
 		return nil
 	}
 
-	if err := l.setUpCapture(); err != nil {
```
IDisposable, line 114: Moved the setup inside `startCapture()`

internal/lldp/lldp.go
```diff
@@ -74,54 +70,34 @@ func NewLLDP(opts *Options) *LLDP {
...
-	}
-
 	return l.startCapture()
 }
 
 // SetAdvertiseOptions updates the advertise options and resends LLDP packets if TX is running
 func (l *LLDP) SetAdvertiseOptions(opts *AdvertiseOptions) error {
 	l.mu.Lock()
-	txRunning := l.txRunning
+	logger := l.l
```
IDisposable, line 99: Since this gets called because of change in the advertising options, we can't use either the Rx or Tx loggers, so use the top-level LLDP logger.

internal/lldp/lldp.go
```diff
@@ -130,16 +106,16 @@ func (l *LLDP) SetAdvertiseOptions(opts *AdvertiseOptions) error {
...
 	}
 
 	return nil
 }
 
 func (l *LLDP) SetRxAndTx(rx, tx bool) error {
 	l.mu.Lock()
-	l.enableRx = rx
-	l.enableTx = tx
+	l.Rx.Enabled = rx
```
IDisposable, line 117: This is now executed both when things change and one time at device startup

internal/lldp/rx.go
```diff
@@ -30,21 +32,7 @@ var multicastAddrs = []string{
 	"01:00:0C:CC:CC:CC",
 }
 
-func (l *LLDP) setUpCapture() error {
-	l.mu.Lock()
```
IDisposable, line 34: This method is now executed while already holding the lock inside `prepareCapture()` and it has setup all the Rx logger and rx TPacket before we get here.

internal/lldp/rx.go
```diff
@@ -30,21 +32,7 @@ var multicastAddrs = []string{
...
-	}
-
-	logger := l.l.With().Str("interface", l.interfaceName).Logger()
-	tPacketRx, err := afPacketNewTPacket(l.interfaceName)
-	if err != nil {
-		return err
-	}
-	logger.Info().Msg("created TPacketRx")
-
+func (l *LLDP) setUpPacketSourceUnderLock(tPacket *afpacket.TPacket, logger *zerolog.Logger) (*gopacket.PacketSource, error) {
```
IDisposable, line 35: This now is only tasked with setting up the BPF filter and packet source.

internal/lldp/rx.go
```diff
@@ -72,23 +60,31 @@ func (l *LLDP) setUpCapture() error {
...
-	return nil
+	logger.Info().Msg("BPF filter set")
+	return gopacket.NewPacketSource(tPacket, layers.LayerTypeEthernet), nil
 }
 
-func (l *LLDP) doCapture(logger *zerolog.Logger) {
-	if l.pktSourceRx == nil || l.rxCtx == nil {
+func (l *LLDP) doCapture() {
+	l.mu.Lock()
+	l.Rx.Running = true
```
IDisposable, line 76: `doCapture()` now manages the Rx running flag like we did in Tx.

internal/lldp/rx.go
```diff
@@ -72,23 +60,31 @@ func (l *LLDP) setUpCapture() error {
...
+	l.mu.Unlock()
+
+	defer func() {
+		l.mu.Lock()
+		l.Rx.Running = false
+		l.mu.Unlock()
+	}()
+
+	if l.pktSourceRx == nil || ctx == nil {
 		logger.Error().Msg("packet source or RX context not initialized")
```
IDisposable, line 88: I don't think this clause could ever happen, but left it in just in case.

internal/lldp/rx.go
```diff
@@ -101,32 +97,37 @@ func (l *LLDP) doCapture(logger *zerolog.Logger) {
...
 		if err != nil {
 			logger.Error().
 				Err(err).
 				Msg("error getting next packet")
 
 			// Immediately break for known unrecoverable errors
-			if err == io.EOF || err == io.ErrUnexpectedEOF ||
-				err == io.ErrNoProgress || err == io.ErrClosedPipe || err == io.ErrShortBuffer ||
-				err == syscall.EBADF ||
+			if errors.Is(err, io.EOF) ||
```
IDisposable, line 115: The is the preferred idiom since 1.3 because it can recognize wrapped `error`s

internal/lldp/rx.go
```diff
@@ -136,30 +137,51 @@ func (l *LLDP) doCapture(logger *zerolog.Logger) {
 	}
 }
 
-func (l *LLDP) startCapture() error {
+func (l *LLDP) prepareCapture() (bool, error) {
```
IDisposable, line 140: `prepareCapture()` does all the setup and returns true if the caller needs to call `doCapture()` (e.g. run the Rx loop).

internal/lldp/rx.go
```diff
@@ -136,30 +137,51 @@ func (l *LLDP) doCapture(logger *zerolog.Logger) {
...
 	l.mu.Lock()
 	defer l.mu.Unlock()
 
-	if l.rxRunning {
-		return nil // Already running
+	if l.Rx.Running {
+		l.l.Debug().Msg("LLDP receiver already running")
```
IDisposable, line 145: ... we haven't created the Rx logger yet, use the LLDP level one.

internal/lldp/rx.go
```diff
@@ -136,30 +137,51 @@ func (l *LLDP) doCapture(logger *zerolog.Logger) {
...
+		tPacket.Close()
+		return false, err
 	}
 
-	logger := l.l.With().Str("interface", l.interfaceName).Logger()
-	logger.Info().Msg("starting capture LLDP ethernet frames")
+	runningState := &RunningState{Enabled: true, Logger: &logger}
+	runningState.Ctx, runningState.Cancel = context.WithCancel(context.Background())
+	runningState.TPacket = tPacket
+	l.Rx = runningState
```
IDisposable, line 168: monotonic switch the new running Rx state...

internal/lldp/rx.go
```diff
@@ -335,52 +357,54 @@ func (l *LLDP) stopCapture() error {
...
 	logger.Info().Dur("duration", time.Since(start)).Msg("RX goroutine finished")
 
-	l.rxRunning = false
-
-	if l.tPacketRx != nil {
-		logger.Info().Msg("closing TPacketRx")
-		l.tPacketRx.Close()
-		l.tPacketRx = nil
-	}
+	l.Rx.Running = false
```
IDisposable, line 394: Eagerly clear the running flag because we're about to tear everything down in reverse order.

internal/lldp/tx.go
```diff
@@ -9,22 +9,17 @@ import (
...
-	"github.com/rs/zerolog"
 )
 
 var (
 	lldpDstMac    = net.HardwareAddr([]byte{0x01, 0x80, 0xc2, 0x00, 0x00, 0x0e})
 	lldpEtherType = layers.EthernetTypeLinkLayerDiscovery
 )
 
-func (l *LLDP) toPayloadValues() []layers.LinkLayerDiscoveryValue {
+func (l *LLDP) toPayloadValues(opts *AdvertiseOptions) []layers.LinkLayerDiscoveryValue {
```
IDisposable, line 19: Passing in the options we grabbed under read-lock  in `sendTxPackets()`

internal/lldp/tx.go
```diff
@@ -74,35 +69,31 @@ func (l *LLDP) toPayloadValues() []layers.LinkLayerDiscoveryValue {
...
+	txState.TPacket, err = afPacketNewTPacket(l.interfaceName)
 	if err != nil {
 		return err
 	}
-	logger.Info().Msg("created TPacket instance for sending LLDP packets")
-
-	l.tPacketTx = tPacketTx
 
+	logger.Info().Msg("created TPacket instance for sending LLDP packets")
+	l.Tx = txState
```
IDisposable, line 89: monotonically swap Tx running state

internal/lldp/tx.go
```diff
@@ -181,50 +203,45 @@ func (l *LLDP) doSendPeriodically(logger *zerolog.Logger, txCtx context.Context)
...
 	l.mu.RUnlock()
 
 	if running || !enabled {
+		logger.Trace().Bool("running", running).Bool("enabled", enabled).Msg("alrady running Tx or not enabled")
 		return nil
 	}
 
-	if cancel != nil {
-		cancel()
+	if previousCancel != nil {
```
IDisposable, line 217: Renamed for clarification that this forces the shut-down of the _previously_ running Tx loop

internal/lldp/tx.go
```diff
@@ -181,50 +203,45 @@ func (l *LLDP) doSendPeriodically(logger *zerolog.Logger, txCtx context.Context)
...
 	}
 
-	logger := l.l.With().Str("interface", l.interfaceName).Logger()
-	logger.Info().Msg("stopping LLDP transmitter")
-
 	// Cancel context to signal stop
-	txCancel := l.txCancel
-	l.txRunning = false
+	txCancel := l.Tx.Cancel
+	l.Tx.Running = false
```
IDisposable, line 240: _techinically_ we should probably let the `defer` code in `doSendPeriodically()`  clear this flag, but an early teardown indication isn't going to hurt us here.

internal/network/types/config.go
```diff
@@ -42,7 +42,7 @@ type NetworkConfig struct {
 	IPv6Mode   null.String       `json:"ipv6_mode,omitempty" one_of:"slaac,dhcpv6,slaac_and_dhcpv6,static,link_local,disabled" default:"slaac"`
 	IPv6Static *IPv6StaticConfig `json:"ipv6_static,omitempty" required_if:"IPv6Mode=static"`
 
-	LLDPMode                null.String `json:"lldp_mode,omitempty" one_of:"disabled,rx_only,tx_only,rx_and_tx,basic,all" default:"rx_and_tx"`
+	LLDPMode                null.String `json:"lldp_mode,omitempty" one_of:"disabled,enabled,rx_only,tx_only,rx_and_tx,basic,all" default:"rx_and_tx"`
```
IDisposable, line 45: Seems like the factory-reset defaults to `enabled` not any of the other options, so declare it as valid to avoid config validation errors.

pkg/nmlite/link/manager.go
```diff
@@ -414,7 +414,7 @@ func (nm *NetlinkManager) reconcileDefaultRoute(link *Link, expected map[string]
 
 	// add remaining expected default routes
 	for _, gateway := range expected {
-		nm.logger.Warn().Str("gateway", gateway.String()).Msg("adding default route")
+		nm.logger.Info().Str("gateway", gateway.String()).Msg("adding default route")
```
IDisposable, line 417: Could be `.Debug()` if we wanted less logging.

config.go
```diff
@@ -254,6 +255,16 @@ func LoadConfig() {
 		loadedConfig.KeyboardLayout = "en-US"
 	}
 
+	// fixup old lldp modes enabled and all are now rx_and_tx
```
IDisposable, line 258: Do a load-time migration of the configuration from the old legacy values to what we interpret them as now.

main.go
```diff
@@ -15,6 +15,14 @@ var appCtx context.Context
 
 func Main() {
 	logger.Log().Msg("JetKVM Starting Up")
+
```
IDisposable, line 18: Setup a top-level panic handler to catch and log unhandled panics so we at least get a log message. We might want to add an `reboot()` call here to see if we can recover?

network.go
```diff
@@ -185,12 +185,20 @@ func initNetwork() error {
 
 	nm := nmlite.NewNetworkManager(context.Background(), networkLogger)
 	networkLogger.Info().Interface("networkConfig", nc).Str("hostname", nc.Hostname.String).Str("domain", nc.Domain.String).Msg("initializing network manager")
-	_ = setHostname(nm, nc.Hostname.String, nc.Domain.String)
+
+	if err := setHostname(nm, nc.Hostname.String, nc.Domain.String); err != nil {
```
IDisposable, line 189: Seems bad manners to ignore these errors, so emit them to the log at least.

network.go
```diff
@@ -202,17 +210,17 @@ func initNetwork() error {
...
 		AdvertiseOptions: advertiseOptions,
 		OnChange: func(neighbors []lldp.Neighbor) {
 			// TODO: send deltas instead of the whole list
 			writeJSONRPCEvent("lldpNeighbors", neighbors, currentSession)
 		},
 		Logger: networkLogger,
 	})
-	if err := lldpService.Start(); err != nil {
-		networkLogger.Error().Err(err).Msg("failed to start LLDP service")
+
```
IDisposable, line 220: Instead of a distinct `Start()` method, we can just leverage the `SetRxAndTx()` function to manage the Rx and Tx loops.

network.go
```diff
@@ -233,7 +241,7 @@ func updateLLDPOptions(nc *types.NetworkConfig, ifState *types.InterfaceState) {
 	}
 
 	if err := lldpService.SetRxAndTx(nc.ShouldEnableLLDPReceive(), nc.ShouldEnableLLDPTransmit()); err != nil {
-		networkLogger.Error().Err(err).Msg("failed to set LLDP RX and TX")
+		networkLogger.Error().Err(err).Msg("failed to update LLDP RX and TX")
```
IDisposable, line 244: Disambiguates the "update" from the "initialize" path in network.go

IDisposable:
I tested this with reboots in every setting (none/rx-only/tx-only/both) and flipping around through the options and it looks like it works correctly without any panics.

ym (approved):

IDisposable:
@ym, I left this branch running on my personal JetKVM overnight and came back to a logo-only screen and unresponsive to the local IP. Not sure what happened, and in the process of trying to grab logs I broke off the HDMI connector (Doh!!) and now that guy is toast. Sorry.  I have another so continuing on with coding/testing... but wanted to warn you that this version might have an issue I'm unaware of the cause. I _did_ have the HDMI sleep enabled, so could be something there unrelated to this change...

PR #963, opened 2025-11-12
fix: wait for channel to open before triggering initial state updates

adamshiervani:


IDisposable (approved):
Awesome, so much wiser.

PR #964, opened 2025-11-13
feat: move native to a separate process, again 

ym:


IDisposable:
Why are we splitting this stuff back out to a distinct process? (not that I object, just wondering about the motivation)

ym:
cgo panics will crash the entire process. While we’ve added a failsafe mode to avoid disrupting the user’s device, in many cases a simple restart is enough to recover.

By splitting this into a separate process, we can restart it independently and provide a better user experience, with failsafe mode only triggered after repeated failures.

internal/native/grpc_client.go
```diff
@@ -0,0 +1,258 @@
...
+
+		ctx := context.Background()
+		stream, err := c.client.StreamEvents(ctx, &pb.Empty{})
+		if err != nil {
+			c.logger.Warn().Err(err).Msg("failed to start event stream, retrying ...")
+			time.Sleep(5 * time.Second)
+			continue
+		}
+
+		c.setStream(stream)
```
IDisposable: ~Don't we want to do this mutation under a mutex lock?~

Should we _retest_ c.closed after grabbing it back...

internal/native/grpc_client.go
```diff
@@ -0,0 +1,258 @@
...
+		c.handleEventStream(stream)
+		c.setStream(nil)
```
IDisposable: Seems like a good place for a defer :) to nil the stream... 

Maybe it would be clearer to have the handleEventStream set and nil it itself?

internal/native/grpc_client.go
```diff
@@ -0,0 +1,258 @@
...
+				logger.Debug().Msg("event stream closed")
+			} else {
+				logger.Warn().Err(err).Msg("event stream error")
+			}
+			break
+		}
+
+		// enrich the logger with the event type and data, if debug mode is enabled
+		if c.logger.GetLevel() <= zerolog.DebugLevel {
+			logger = logger.With().
```
IDisposable, line 113: Will this "stack up" the context each time through the loop? E.g. since we're grabbing an augmented logger, then using that same one next message, what does .With().Str("type", xxx) do?

internal/native/proxy.go
```diff
@@ -0,0 +1,679 @@
...
+	restartMu   sync.Mutex
+	restarts    uint
+	stopped     bool
+	processWait chan error
+}
+
+// NewNativeProxy creates a new NativeProxy that spawns a separate process
+func NewNativeProxy(opts NativeOptions) (*NativeProxy, error) {
+	proxyOptions := opts.toProxyOptions()
+	proxyOptions.VideoStreamUnixSocket = fmt.Sprintf("@jetkvm/native/video-stream/%s", randomId(4))
```
IDisposable, line 142: Would it be safer to use `os.CreateTemp()` to ensure we get a name that is known unused in an atomic action? Something like 
```golang
// Generate a unique socket path in /tmp with a prefix.
tmpFile, err := os.CreateTemp("/tmp", "jetkvm-native-video-stream-*.sock")
if err != nil {
    fmt.Printf("Failed to create temp file: %v\n", err)
    // pout os.Exit(1)
}
proxyOptions.VideoStreamUnixSocket := tmpFile.Name() 
tmpFile.Close()
os.Remove(socketPath) 
```

ym: We’re using an abstract socket now, so it doesn’t touch the filesystem and `os.CreateTemp()` isn’t needed anymore.

adamshiervani (approved):

PR #967, opened 2025-11-14
refactor: More robust handling of jsonrpc calls (0.4.10 backport)

ym:
Co-authored-by: Copilot <175728472+Copilot@users.noreply.github.com>
Co-authored-by: Adam Shiervani <adam.shiervani@gmail.com>

### DIFF
--- a/ui/src/utils/jsonrpc.ts
+++ b/ui/src/utils/jsonrpc.ts
@@ -1,0 +1,244 @@
+import { useRTCStore } from "@/hooks/stores";
+import { sleep } from "@/utils";
+
+// JSON-RPC utility for use outside of React components
+
+export interface JsonRpcCallOptions {
+  method: string;
+  params?: unknown;
+  attemptTimeoutMs?: number;
+  maxAttempts?: number;
+}
+
+export interface JsonRpcCallResponse<T = unknown> {
+  jsonrpc: string;
+  result?: T;
+  error?: {
+    code: number;
+    message: string;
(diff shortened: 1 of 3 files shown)


ui/src/utils/jsonrpc.ts
```diff
@@ -0,0 +1,244 @@
...
+      jsonrpc: "2.0",
+      method: options.method,
+      params: options.params || {},
+      id: requestId,
+    };
+
+    const messageHandler = (event: MessageEvent) => {
+      try {
+        const response = JSON.parse(event.data) as JsonRpcCallResponse<T>;
+        if (response.id === requestId) {
```
IDisposable, line 91: _note to self_: There's another whole instance of `messageHandler` listening for their response ID, so someone is going to catch this message, just not this instance.

ui/src/utils/jsonrpc.ts
```diff
@@ -0,0 +1,244 @@
...
+    maxAttempts: 6,
+  });
+
+  if (response.error) throw response.error;
+  return response.result;
+}
+
+export async function getLocalVersion() {
+  const response = await callJsonRpc<VersionInfo>({ method: "getLocalVersion" });
+  if (response.error) throw response.error;
```
IDisposable, line 242: ... all the other places in this file wrap this in a `new Error()` so either we don't need to do that in the other places, or do we need to do that here , to be consistent?

IDisposable (approved):

PR #969, opened 2025-11-14
fix: hidRPC handshake packet should be only sent once

### LINKED ISSUE #806: HidRPC handshake packet being sent multiple times
### Disclaimer

### Application version

0.4.7

### System version

0.2.5

### Device model

None

### Extension model

None

### Remote device Hardware

_No response_

### Remote device OS

_No response_

### Bug description

There are multiple handshakes for hidRpc (as there is no global lock to prevent other `useHidRpc` instances from sending handshake packets). This doesn’t currently have any side effects, but we need to fix it in a later version.

ym:
fixes https://github.com/jetkvm/kvm/issues/806

### DIFF
--- a/ui/src/hooks/useHidRpc.ts
+++ b/ui/src/hooks/useHidRpc.ts
@@ -1,4 +1,5 @@
 import { useCallback, useEffect, useMemo } from "react";
+import { Logger } from "tslog";
 
 import { useRTCStore } from "@hooks/stores";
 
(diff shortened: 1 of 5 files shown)


ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,125 @@ interface sendMessageParams {
...
+
+  const onClose = () => {
+    resetHandshake();
+
+    logger.info("Channel closed");
+    setRpcHidProtocolVersion(null);
+  };
+
+  rpcHidChannel.addEventListener("open", onConnected);
+  rpcHidChannel.addEventListener("close", onClose);
```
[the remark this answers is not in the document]
IDisposable: Probably good, but should clear the `handshakeInterval` _first_ to ensure we don't try processing another message while tearing things down.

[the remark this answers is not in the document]
IDisposable: Good point, and something I need to sweep the general code for this pattern.

ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,125 @@ interface sendMessageParams {
...
+      clearInterval(handshakeInterval);
+      handshakeInterval = null;
+    }
+    const timeUsed = lastSendTime ? Date.now() - lastSendTime.getTime() : 0;
+    logger.info(`Handshake completed in ${timeUsed}ms after ${attempts} attempts (Version: ${message.version} / ${HID_RPC_VERSION})`);
+
+    rpcHidChannel.removeEventListener("message", onMessage);
+  };
+
+  const resetHandshake = (newLastConnectedTime?: Date | undefined) => {
```
IDisposable: Modify to take a `completed` boolean with  the value to set `handshakeCompleted`

ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,125 @@ interface sendMessageParams {
...
+
+    setRpcHidProtocolVersion(message.version);
+
+    // clean up 
+    handshakeCompleted = true;
+    if (handshakeInterval) {
+      clearInterval(handshakeInterval);
+      handshakeInterval = null;
+    }
+    const timeUsed = lastSendTime ? Date.now() - lastSendTime.getTime() : 0;
```
IDisposable, line 106: Move this calculation to just after 107 (`handshakeCompleted = true`) and then replace 107-111 with a call to `resetHandshake(true, Date.now())` to make sure there's only on set of tear-down for the timer.

ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,125 @@ interface sendMessageParams {
...
+    attempts = 0;
+    handshakeCompleted = false;
+    if (handshakeInterval) {
+      clearInterval(handshakeInterval);
+      handshakeInterval = null;
+    }
+  };
+
+  const onConnected = () => {
+    resetHandshake(new Date());
```
IDisposable: `resetHandshake(false, Date.now());`

ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,125 @@ interface sendMessageParams {
...
+    logger.info("Channel connected");
+
+    sendHandshake(true);
+    rpcHidChannel.addEventListener("message", onMessage);
```
IDisposable, line 130: This should be removed in `onClose` to ensure we don't get called on a closed channel

ui/src/hooks/useHidRpc.ts
```diff
@@ -151,99 +271,44 @@ export function useHidRpc(onHidRpcMessage?: (payload: RpcMessage) => void) {
...
+      if (message instanceof HandshakeMessage) return; // handshake message is handled by the doRpcHidHandshake function
 
-      onHidRpcMessage?.(message);
-    };
-
-    const openHandler = () => {
-      console.info("HID RPC channel opened");
-      sendHandshake();
-    };
+      /* @__PURE__ */ logger.debug("Received message", message);
```
IDisposable, line 297: What's the PURE decoration for? Why is this indented oddly?

ui/vite.config.ts
```diff
@@ -39,7 +39,7 @@ export default defineConfig(({ mode, command }) => {
   return {
     plugins,
     esbuild: {
-      pure: ["console.debug"],
+      pure: command === "build" ? ["console.debug", "logger.debug"] : ["console.debug"],
```
IDisposable: Isn't logger.debug always going to be PURE?

ym: By setting `logger.debug` in `esbuild.pure`, all `logger.debug` calls will be removed when their return value isn’t used.
`esbuild.drop` only applies to `console.*`, so we can’t rely on it here.

https://esbuild.github.io/api/#pure

ui/src/hooks/useHidRpc.ts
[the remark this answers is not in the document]
ym: It's not a React hook, so I don't think it would work.

ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,128 @@ interface sendMessageParams {
...
+    }
+
+    setRpcHidProtocolVersion(message.version);
+
+    // clean up 
+    const timeUsed = lastSendTime ? Date.now() - lastSendTime.getTime() : 0;
+    logger.info(`Handshake completed in ${timeUsed}ms after ${attempts} attempts (Version: ${message.version} / ${HID_RPC_VERSION})`);
+    resetHandshake({ completed: true });
+
+    rpcHidChannel.removeEventListener("message", onMessage);
```
IDisposable, line 110: We might want to move this remove up to line 105 to ensure we don't get pinged again.

ui/src/hooks/useHidRpc.ts
```diff
@@ -25,6 +26,128 @@ interface sendMessageParams {
...
+  const onClose = () => {
+    resetHandshake({ lastConnectedTime: undefined, completed: false });
+
+    logger.info("Channel closed");
+    setRpcHidProtocolVersion(null);
+
+    rpcHidChannel.removeEventListener("message", onMessage);
+  };
+
+  rpcHidChannel.addEventListener("open", onConnected);
```
IDisposable, line 142:
```suggestion
  rpcHidChannel.addEventListener("close", onClose); 

  // handle case where channel is already open when the hook is mounted
  if (rpcHidChannel.readyState === "open") {
    onConnected();
  } else {
    rpcHidChannel.addEventListener("open", onConnected);
  }
```

IDisposable: If we only add the even listener if we're not already open then it won't get double-called

IDisposable (approved):
Just a little suggestion, not required.

ui/src/hooks/useHidRpc.ts
ym: My concern is that with this change, the handshake packet won’t be sent after rpcHidChannel reconnects.

IDisposable (approved):

PR #980, opened 2025-11-19
Enhance synctrace logging 

IDisposable:
This update enhances the performance and features of the `synctrace` logging of `Lock`/`Unlock`/`TryLock` for `Mutex` and `RWMutex` for improved visibility into concurrent operations and aids in debugging synchronization-related issues by providing more detailed `synctrace` logging.

## sync changes

- Switched the maps to be indexed by the `.Pointer` as a `uintptr` (not as a `string`) so `map` is smaller and faster.
- Grouped the `lockCount`, `unlockCount`, and `lastLock` into a `trackingEntry` so we can detect unlocks of something that wasn't ever locked, excessive unlocks, and also track the `firstLock` and the `lastUnlock`.
- Added `LogDangledLocks` for debugging use, just insert a call to this anywhere you might have concerns.
- Added a panic handler to the Main so we can log out panics

## other changes

Migrates from standard library `sync` package imports to an internal `internal/sync` package across almost all of  the codebase.  

Exceptions:
- _jsonrpc.go_ Way too busy*
- _usb.go_ Way too busy*
- _dc_metrics.go_ Only uses a `sync.Once` which isn't worth tracking
- _internal/logging/logger.go_ Logging logs while logging messages seems too loggish
- _internal/native/cgo_linux.go_ Very performance intensive and too busy when tracing*
- _internal/sync/log.go_ Wouldn't be worthwhile logging lock usage inside the sync logging itself
- _pkg/myip/check.go_ Stand-alone package that only uses the trackables in `checkIP` function to sync its own children
- _pkg/myip/ip.go_ Stand-alone package

`*` These are performance sensitive or very frequent lock/unlock that just emits too many logs to track. If there's a problem in those modules, you can temporarily swap over to using the logger by replacing `import "sync"` with `import "github.com/jetkvm/kvm/internal/sync"`


internal/sync/log.go
```diff
@@ -47,103 +48,160 @@ func logTrack(callerSkip int) *zerolog.Logger {
...
+	logLockTrack(ptr).Trace().Msg("locked mutex for reading")
+}
+
+// You can all this function at any time to log any dangled locks currently tracked
+// it's not an error for there to be open locks, but this can help identify any
+// potential issues if called judiciously
+func LogDangledLocks() {
+	defaultLogger.Info().Msgf("Checking %v tracked locks for dangles", len(tracking))
+
+	indexMu.Lock()
```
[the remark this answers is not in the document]
IDisposable: While a valid point, we don't want to emit our own logs while the lock is held, so we're okay with this informational only message to possibly be slightly off.

internal/sync/log.go
```diff
@@ -47,103 +48,160 @@ func logTrack(callerSkip int) *zerolog.Logger {
...
-	logLockTrack(i).Trace().Msg("locked mutex for reading")
+	ptr := getPointer(t)
+	increaseLockCount(ptr)
+	logLockTrack(ptr).Trace().Msg("locked mutex for reading")
+}
+
+// You can call this function at any time to log any dangled locks currently tracked
+// it's not an error for there to be open locks, but this can help identify any
+// potential issues if called judiciously
+func LogDangledLocks() {
```
IDisposable, line 169: @adamshiervani you could add a call to this for the new diagnostics endpoint :)

adamshiervani:
In some of the merges i think something got lost: 

I suggest something like 
```        --enable-sync-trace)
            ENABLE_SYNC_TRACE=1
            shift
            ;;
```

```# If sync trace is enabled at build time, ensure synctrace is in LOG_TRACE_SCOPES
if [ "$ENABLE_SYNC_TRACE" = 1 ]; then
    if [[ ! "$LOG_TRACE_SCOPES" =~ synctrace ]]; then
        LOG_TRACE_SCOPES="${LOG_TRACE_SCOPES},synctrace"
    fi
fi
```

IDisposable:
> In some of the merges i think something got lost

Fixed and conflict resolved! Ready for review @adamshiervani @ym

main.go
```diff
@@ -30,6 +30,13 @@ func Main() {
 
 	logger.Log().Msg("JetKVM Starting Up")
 
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Panic().Interface("error", r).Msg("Received panic")
```
adamshiervani: The `panic()` after this line wont be called as the logger will call `panic()` with the `Msg()`

See: https://pkg.go.dev/github.com/rs/zerolog#Logger.Panic

adamshiervani (changes requested):
Just panic fix and then we're good to go.

main.go
IDisposable: I wasn't intending to change the fact that it panics, just that we log out that it did before it does :)

I will switch to using `.WithLevel(`, but if we don't keep line 36, that might result in a "odd" goroutine state (but at least it'll still be responsive to the API).

IDisposable: Switched to `.WithLevel(zerolog.PanicLevel)` so the code will fall-through to line 36 (and do the `panic(r)`) if we want to delete that too... please say so (with the caveat expressed above)

adamshiervani (approved):

PR #984, opened 2025-11-20
When adding/removing routes, honor scope and protocol and manage gateways correctly

IDisposable:
While trying to figure out why the IP routing gets messed up after a lot of DHCP (or even static) changes, discovered that the gateway routes were not being correctly updated.

Added a lot more logging detail to find the route bugs which was that the gateways were being added to the map of things to update with their IP **and netmask**, then looked for by only the IP, which meant we never removed them.

Also replaced log context additions that did `.Str("foo", foo.String())` with `.Stringer("foo", foo)` so the conversion to string is lazy and not done if the log level would result in the Msg() being dropped.

### DIFF
--- a/pkg/nmlite/interface.go
+++ b/pkg/nmlite/interface.go
@@ -8,12 +8,13 @@ import (
 
 	"time"
 
-	"github.com/jetkvm/kvm/internal/sync"
-
 	"github.com/jetkvm/kvm/internal/confparser"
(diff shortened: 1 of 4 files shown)


ym (approved):

IDisposable:
@adamshiervani This is safe to merge, I've been running the code for weeks :)

PR #995, opened 2025-11-20
Fix AltGr handling for Windows clients

Riyooo:
## Issue

This PR fixes AltGr keyboard handling for **Windows clients**.

On Windows with international layouts (e.g. French), AltGr-based characters such as `@`, `€`, `|` were not working correctly. JetKVM's UI saw both `ControlLeft` and `AltRight` instead of a proper AltGr (right Alt) modifier, and forwarded both to the remote system, so the expected characters were never produced..

## Root Cause

Windows emulates AltGr by injecting a synthetic `ControlLeft` before `AltRight`. Browsers therefore see a sequence like:

```text
↓ ControlLeft
↓ AltRight
… later …
↑ ControlLeft
↑ AltRight
```
This happens for each AltGr press (and repeats when the key is held down). The remote side, however, typically expects only AltRight to be held for AltGr.

## Fix

This change adds a small, Windows-only, adjustment in the keyboard handling:

- On Windows, the component now tracks the **last keydown** event (`code` and timestamp).
- When a new `AltRight` keydown is received, it checks whether the **previous** keydown was `ControlLeft` and whether it happened within a small time window (`altGrSyntheticThresholdMs = 3 ms`).
- If both conditions are true, this is considered the **synthetic AltGr sequence** generated by Windows (CtrlLeft immediately followed by AltRight). In that case, the code sends a synthetic `ControlLeft` keyup to cancel the extra Ctrl modifier, so the effective state on the guest is “AltRight (AltGr) only”.
- In all other cases (including normal `Ctrl + AltRight` combinations with human timing), key events are forwarded unchanged.

The 3 ms threshold was chosen based on measurements made on 2 Windows PCs, where the synthetic `ControlLeft → AltRight` gap was consistently observed between 0 and ~2.1 ms. 3 ms gives a small safety margin to catch the Windows-generated pattern, while still being far below any realistic human keypress interval.

## Limitations / Trade-offs

The early guard to avoid issues when holding modifier keys ```if (e.repeat) return;``` is left unmodified.
Because of that, some less common combinations where AltRight is pressed first and a Control key is pressed while AltGr is held may not behave perfectly on Windows. In practice, these are rare compared to:
- The primary AltGr use case (typing symbols), and
- More typical Ctrl+Alt combinations (e.g. ControlLeft + AltLeft).

The goal here is to fix the main AltGr problem with minimal added complexity. If there’s real-world demand for those more exotic combinations, the logic around e.repeat can be revisited in a follow-up change.

### DIFF
--- a/ui/src/components/WebRTCVideo.tsx
+++ b/ui/src/components/WebRTCVideo.tsx
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useResizeObserver } from "usehooks-ts";
 
 import { cx } from "@/cva.config";
+import { isWindows } from "@/utils";
 import useKeyboard from "@hooks/useKeyboard";
 import useMouse from "@hooks/useMouse";
 import {
@@ -73,6 +74,12 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
     },
     [setVideoClientSize, setVideoSize]
   );
+
+  // AltGr Fix for Windows Clients
+  const altGrSyntheticThresholdMs = 3;
+  const isWindowsClient = useMemo(() => isWindows(), []);
+  const lastKeyDownRef = useRef<{ hidKey: number; time: number } | null>(null);
+  const altGrLoopRef = useRef(false);
 
   useResizeObserver({
     ref: videoElm as React.RefObject<HTMLElement>,
@@ -250,13 +257,41 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
   const keyDownHandler = useCallback(
     (e: KeyboardEvent) => {
       e.preventDefault();
-      if (e.repeat) return;
       const code = getAdjustedKeyCode(e);
       const hidKey = keys[code];
 
       if (hidKey === undefined) {
         console.warn(`Key down not mapped: ${code}`);
         return;
+      }
+
+      // Detect Windows synthetic AltGr (CtrlLeft then AltRight within ~3ms) and cancel the synthetic Ctrl
+      if (isWindowsClient) {
+        // Buffer ControlLeft briefly; if no AltRight follows within the threshold, treat it as a real ControlLeft press.
+        if (hidKey === keys.ControlLeft) {
+          const controlLeftDownTime = e.timeStamp;
+          lastKeyDownRef.current = { hidKey, time: controlLeftDownTime };
+          setTimeout(() => {
+            if (
+              lastKeyDownRef.current?.hidKey === keys.ControlLeft &&
+              lastKeyDownRef.current.time === controlLeftDownTime
+            ) {
+              lastKeyDownRef.current = null;
+              handleKeyPress(keys.ControlLeft, true);
+            }
+          }, altGrSyntheticThresholdMs);
+          return;
+        }
+
+        // If AltRight arrives shortly after ControlLeft, treat the pair as AltGr and cancel the pending ControlLeft.
+        if (
+          hidKey === keys.AltRight &&
+          lastKeyDownRef.current?.hidKey === keys.ControlLeft &&
+          e.timeStamp - lastKeyDownRef.current.time <= altGrSyntheticThresholdMs
+        ) {
+          altGrLoopRef.current = true;
+          lastKeyDownRef.current = null;
+        }
       }
 
       // When pressing the meta key + another key, the key will never trigger a keyup
@@ -282,7 +317,7 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
         }, 100);
       }
     },
-    [handleKeyPress, isKeyboardLockActive],
+    [handleKeyPress, isKeyboardLockActive, isWindowsClient],
   );
 
   const keyUpHandler = useCallback(
@@ -296,10 +331,27 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
         return;
       }
 
+      // On Windows, handle ControlLeft specially to preserve FIFO semantics with AltGr buffering.
+      if (isWindowsClient && hidKey === keys.ControlLeft) {
+
+        // Synthetic AltGr ControlLeft: never sent a down, swallow the release as well.
+        if (altGrLoopRef.current) {
+          altGrLoopRef.current = false;
+          return;
+        }
+
+        // Very fast real Ctrl tap: flush the pending down before the up.
+        if (lastKeyDownRef.current?.hidKey === keys.ControlLeft) {
+          handleKeyPress(keys.ControlLeft, true);
+        }
+
+        lastKeyDownRef.current = null;
+      }
+
       console.debug(`Key up: ${hidKey}`);
       handleKeyPress(hidKey, false);
     },
-    [handleKeyPress],
+    [handleKeyPress, isWindowsClient],
   );
 
   const videoKeyUpHandler = useCallback((e: KeyboardEvent) => {


ui/src/components/WebRTCVideo.tsx
```diff
@@ -253,12 +261,23 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
...
 
       if (hidKey === undefined) {
         console.warn(`Key down not mapped: ${code}`);
         return;
       }
 
+      // Detect Windows synthetic AltGr (CtrlLeft then AltRight within ~3ms) and cancel the synthetic Ctrl
+      if (
+        isWindowsClient &&
+        code === "AltRight" &&
```
IDisposable: Could we use the hidKey instead of the code?

ui/src/components/WebRTCVideo.tsx
```diff
@@ -74,6 +74,14 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
     [setVideoClientSize, setVideoSize]
   );
 
+  // AltGr Fix for Windows Clients
+  const altGrSyntheticThresholdMs = 3;
+  const isWindowsClient = useMemo(
+    () => typeof navigator !== "undefined" && navigator.userAgent?.includes("Win"),
```
IDisposable: Maybe this should be a client-side preference to activate instead?

IDisposable:
Thanks for putting this together. I don't _have_ a keyboard with an AltGr key, so couldn't see the issue at all. (I have one on order now!).

I think this fix is surgical and small... just wondered three things:

- Can we use the `hidKey` instead of the `code` in the event? That would be more reliable since those _can't_ be wrong or the HID spec is broken (e.g. it's more reliable)
- Can/should we set this as a preference in Settings since the Window detection is not horrible safe? If we keep it as a `navigator.platform` test, then there's _already_ a method in _utils.ts_ called `isWindows()` that you can use :)
- Should we tweak the the "last code" logic to see/accept either CtrlLeft -> AltRight _or_ AltRight ->  CtrlLeft to address your concern about the manual press of keys?

Riyooo:
Honestly I rarely contribute to projects and it’s a pleasure to be able to help here. I hadn’t noticed this issue myself until I plugged the JetKVM into a machine that had an `@` in its password. Then I saw that the virtual keyboard status bar was not matching what I pressed on the physical keyboard.

## Use of `hidKey` instead of `code`

I used `code` because the Windows AltGr quirk is usually described at the DOM level (`KeyboardEvent.code` of `ControlLeft` followed by `AltRight`). I also looked at how other projects handle this, in particular the noVNC/QEMU solution to this same issue, and they implement their AltGr workaround in terms of `KeyboardEvent.code` as well (checking for `ControlLeft` followed by `AltRight`). That’s part of why my attempt followed the same pattern.

That said, I understand your point about HID codes being the lowest‑level, layout‑independent representation. I don’t have a strong preference either way, and I’d be happy to switch `lastKeyDownRef` to store `{ hidKey, time }` and compare against `keys.ControlLeft` / `keys.AltRight` if you’d prefer the detection to live at the HID layer.

## Setting a client-side preference

My thinking is that this is more of a bug fix than a feature, so it should be enabled by default on Windows rather than something users have to discover and enable:

- Without it, AltGr is effectively broken on many non‑US layouts, which depending on the use case can be a real problem (it did prevent me from logging in).
- The logic is already gated by a Windows check and I don’t see much value for users in being able to disable it.
- The timing window is narrow (3 ms), but it’s based on tests on two Windows PCs. I did have to make a small keyboard event viewer to test this, as the ones I found online did not calculate the delta time between key events, and a found the synthetic gap to be ~0–2.1 ms. I tried to make it just large enough to cover the synthetic sequence, but still small enough not to catch a fast human `ControlLeft` → AltGr press.

For that reason I'd be inclined to keep this enabled by default on Windows and not expose it as a regular UI toggle unless we start seeing cases where it misbehaves. If you feel 3ms is too low, I can bump it to 5ms for example, and/or if you'd prefer I can wire this through a setting as suggested so that advanced users can tweak the delay if they still see issues.

Also, I didn’t notice the existing `isWindows()` helper from `utils.ts`, thanks for mentioning it. Using it is of course better than inlining a `navigator` check, and I can update the PR to use that for platform detection.


## Tweaking the logic

To be clear, the current implementation already handles the `ControlLeft → AltRight` order, but not the `AltRight → ControlLeft` one.

I did experiment with covering more combinations and interactions by modifying the `e.repeat` block and `keyUpHandler`, and also by trying a different approach that intercepted all `ControlLeft` keydown events during a small time frame. In practice that added complexity while giving little benefit compared to fixing the main bug (AltGr symbols not working at all).

For example, one of the later issues I noticed was that not only the `ControlLeft` event received when doing `AltRight → ControlLeft` was flagged as `repeat`, but it also looked very similar to the synthetic `ControlLeft` you get when holding down the AltGr key (same modifiers). That made it harder to distinguish between real and synthetic cases.

That’s why I kept this PR focused on the simplest approach. That said, if you’re okay with a more complex state machine to handle more combinations, I can certainly try to adjust the logic in that direction in a follow-up.

image

IDisposable:
Thanks for the extensive reply... I agree with everything you said.

I (personally) would rather run on the HidKey, but that's trivial. 

If you didn't see any real issues when _actually_ hitting keys in the opposite order,  then it is certainly not worth the state-machine and the 3ms timing is perfect in my eyes.

Riyooo:
Thanks for your feedback and for confirming the approach.

Then I'll update the PR to:
- switch the detection over to `hidKey` (using the HID codes for `ControlLeft` / `AltRight`),
- and use the shared helper from `utils.ts` for platform detection,

I’ll also take another quick look to see if there’s a simple way to cover a few more AltGr combinations without making the state machine too complex.

Riyooo:
Hi, I made the changes we discussed and reviewed the logic. In the end I switched to a slightly more complex approach, which should however cover all `AltGr` combinations in any order. I hope this suits you.

## Summary of the changes

- Switched the detection to `hidKey`,
- Used the shared helper from `utils.ts` for platform detection,
- Used `e.timeStamp` instead of `performance.now()` for more accurate inter-event timing,
- Switched the logic to intercept and flush all synthetic `ControlLeft` events (up and down), instead of cancelling `ControlLeft` itself.

## New approach

- Buffer `ControlLeft` keydown for a few milliseconds (`altGrSyntheticThresholdMs`) instead of sending it immediately.
- If an `AltRight` keydown arrives shortly after that buffered `ControlLeft`, treat the pair as AltGr and never send the synthetic `ControlLeft` (only `AltRight` is sent).
- If no quick `AltRight` follows, treat the buffered `ControlLeft` as real and send it normally (so `Ctrl`, `Ctrl+X`, `Ctrl` then AltGr, etc. still work).
- While in an AltGr sequence, allow modifier repeats (`ControlLeft`, `ControlRight`, `AltLeft`, `AltRight`) so we can keep cancelling all synthetic `ControlLeft` repeat events Windows generates while `AltGr` is held.
- On `ControlLeft` keyup:
  - In the real `ControlLeft` case, send the release.
  - In the AltGr synthetic case, swallow the release so the remote never sees a phantom `ControlLeft` press/release pair.

Overall, the device should now see the intended behavior:
- Pressing and holding `AltGr` only sends `AltRight`.
- Pressing `Ctrl`, then `AltGr`, or `AltGr` then `Ctrl`, works as expected without leaving `ControlLeft` stuck.
- Holding `AltGr` or using other modifier combinations should no longer produce stray or stuck `ControlLeft` states.

IDisposable:
Looks great!

Two things:

- The whole `e.repeat` check should just be deleted... I don't think we should have ever ignored those because we currently don't do "auto-repeat" correctly and I think this is partially responsible. We should process all key-down events.
- Question: What happens in this exact timing _get LeftControl **down**, buffer it, get LeftControl **up**, ignore/eat it, buffer times out, send the LeftControl **down**_, ...  don't we need to know that LeftControl was already released and schedule a synthetic LeftControl **up** for a millisecond or few later?

I think that it needs to be possible to trigger things on Control-press/release pairs (only). There are hardware KVMs, for example, that trigger on _press/release_ Ctrl+Ctrl+(port digit). If the JetKVM is plugged in as the input keyboard to them, we need to be able send these sequences.

I thing

Riyooo:
Thanks for your feedback, here are the changes I made and how they relate to your points:
- **The `e.repeat` check:** Ok I removed it completely. As you noticed I already had to poke a hole in this check to deal with all AltGr combos cause once AltGr is pressed, windows flags every other special key with the repeat flag. It's far simpler this way, and we don’t need to treat repeats specially anymore.
- **Control press/release pairs (Ctrl+Ctrl+digit, KVMs, etc.):** Interesting use-case, I hadn’t used those hardware KVMs before.
For a “pure” Ctrl press/release (no AltGr), the device still sees a normal `Ctrl` down followed by `Ctrl` up. The only time we swallow the `ControlLeft` release is when `altGrLoopRef` has been set by a quick `ControlLeft → AltRight` pair, i.e. in the AltGr synthetic case. To sanity-check this against your `Ctrl+Ctrl+(port digit)` example, I moved my keyboard event viewer (KEV) to a self-hosted server (so I could inspect timings remotely). Then, with JetKVM plugged into an Ubuntu machine and from a Windows PC, I navigated to the KEV's webpage via JetKVM'UI, and did the following:
  - Press/release `ControlLeft`
  - Press/release `ControlLeft` again
  - Press/release `8`
  - Press/release `0`  (for port 80, if I understood correctly)

  As you can see in the screenshot below, the Ctrl press/release pairs are preserved exactly as expected, and the digit presses follow normally, which should satisfy the KVM pattern you described.

2025-11-23 03 43 52 10 0 10 188 2e712eb957fc

- **Ultra-fast `ControlLeft` tap scenario:** There *was* a theoretical edge case where we could have ended up with a lone `Ctrl`. If a user (or an automated tool) managed to press and release `Ctrl` faster than the 3 ms AltGr threshold, and the keyup handler ran before the buffered timeout, we could send a `Ctrl` **up** without ever having sent the buffered `Ctrl` **down`.
I, I’ve fixed this in the current version:
  - When a real `ControlLeft` keyup arrives and there is still a pending buffered `ControlLeft`, we immediately flush the pending `ControlLeft` **down**, then send the `ControlLeft` **up**.
  - If AltGr was detected (`ControlLeft → AltRight` within the threshold), we never sent a `Ctrl` down in the first place and we still swallow the synthetic `ControlLeft` release.

  This ensures that, for non-AltGr cases, the remote machine receives all `ControlLeft` events in proper FIFO order (`down` then `up`), even for very fast presses.

  I re-tested this with the same KEV setup: by trying to do two very fast `ControlLeft` taps with < 3 ms between the first down and up, both presses were correctly registered as `Ctrl` down+up pairs on the remote side (screenshot below).

2025-11-23 05 49 37 10 0 10 188 74cb09875854

ui/src/components/WebRTCVideo.tsx
```diff
@@ -259,6 +265,36 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
...
+        if (hidKey === keys.ControlLeft) {
+          const controlLeftDownTime = e.timeStamp;
+          lastKeyDownRef.current = { hidKey, time: controlLeftDownTime };
+          setTimeout(() => {
+            if (
+              lastKeyDownRef.current?.hidKey === keys.ControlLeft &&
+              lastKeyDownRef.current.time === controlLeftDownTime
+            ) {
+              lastKeyDownRef.current = null;
+              console.debug(`Key down: ${keys.ControlLeft}`);
```
IDisposable: We _can_ kill this `console.debug` now
 You've nailed this!

ui/src/components/WebRTCVideo.tsx
```diff
@@ -259,6 +265,36 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
...
+            }
+          }, altGrSyntheticThresholdMs);
+          return;
+        }
+
+        // If AltRight arrives shortly after ControlLeft, treat the pair as AltGr and cancel the pending ControlLeft.
+        if (
+          hidKey === keys.AltRight &&
+          lastKeyDownRef.current?.hidKey === keys.ControlLeft &&
+          e.timeStamp - lastKeyDownRef.current.time <= altGrSyntheticThresholdMs
```
IDisposable, line 290: _Note to self_ It kinda looks like this needs to be `.current?.` because the current could be null. However it can't be, because the previous condition would have failed.

No change needed.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -296,10 +332,28 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
...
+        // Synthetic AltGr ControlLeft: never sent a down, swallow the release as well.
+        if (altGrLoopRef.current) {
+          altGrLoopRef.current = false;
+          return;
+        }
+
+        // Very fast real Ctrl tap: flush the pending down before the up.
+        if (lastKeyDownRef.current?.hidKey === keys.ControlLeft) {
+          console.debug(`Key down: ${keys.ControlLeft} (flushed on keyup)`);
+          handleKeyPress(keys.ControlLeft, true);
```
IDisposable, line 345: _Note to self_ no need for a delay after this as the `handleKeyPress` actually takes time to send the HID RPC message.

IDisposable (approved):

IDisposable:
:shipit:

ui/src/components/WebRTCVideo.tsx
Riyooo: Done, I’ve removed that specific `console.debug`.
Thanks! it’s great to be able to contribute.

ui/src/components/WebRTCVideo.tsx
```diff
@@ -296,10 +331,28 @@ export default function WebRTCVideo({ hasConnectionIssues }: { hasConnectionIssu
...
+
+        // Synthetic AltGr ControlLeft: never sent a down, swallow the release as well.
+        if (altGrLoopRef.current) {
+          altGrLoopRef.current = false;
+          return;
+        }
+
+        // Very fast real Ctrl tap: flush the pending down before the up.
+        if (lastKeyDownRef.current?.hidKey === keys.ControlLeft) {
+          console.debug(`Key down: ${keys.ControlLeft} (flushed on keyup)`);
```
IDisposable: ... oops, you can kill this debug too :)

IDisposable (approved):

IDisposable:
This is good to go @ym @adamshiervani

PR #997, opened 2025-11-20
fix(network): IPv6 addresses sorting was using wrong references

ym:


pkg/nmlite/utils.go
```diff
@@ -42,17 +42,19 @@ func sortAndCompareStringSlices(a, b []string) bool {
...
+	})
+}
+
 func sortAndCompareIPv6AddressSlices(a, b []types.IPv6Address) bool {
 	if len(a) != len(b) {
 		return false
 	}
 
-	sort.SliceStable(a, func(i, j int) bool {
-		return a[i].Address.String() < b[j].Address.String()
```
IDisposable, line 51: ... so should have been
`return a[i].Address.String() < a[j].Address.String()`
 (that `b` should have been `a`)... oops

pkg/nmlite/utils.go
```diff
@@ -42,17 +42,19 @@ func sortAndCompareStringSlices(a, b []string) bool {
...
-	})
-	sort.SliceStable(b, func(i, j int) bool {
-		return b[i].Address.String() < a[j].Address.String()
```
IDisposable, line 54: ... so should have been
`return a[i].Address.String() < a[j].Address.String()`
 (that `a` should have been `b`)... oops

pkg/nmlite/utils.go
```diff
@@ -42,17 +42,19 @@ func sortAndCompareStringSlices(a, b []string) bool {
...
-	})
+	sortIPv6AddressSlicesStable(a)
+	sortIPv6AddressSlicesStable(b)
```
IDisposable, line 57: Sorry I didn't notice this error :(

IDisposable (approved):

PR #998, opened 2025-11-20
fix: limit video device open attempts to prevent infinite loops

adamshiervani:
## Fix video thread infinite restart loop on signal loss

**Problem:** When display sleeps, video thread can loop indefinitely trying to restart capture, causing:
- Resource exhaustion (20+ device open/close cycles)
- Log spam
- Thread becomes unresponsive, preventing UI restart

**Root cause:** Race between fast polling video thread and slow event-driven format detection thread. Video thread sees stale `detected_signal=true` and keeps retrying before format thread updates it.

**Solution:**

1. **Failsafe counter**: Exit after 20 failed attempts to trigger supervisor restart
2. **Stale flag detection**: After 3 consecutive zero-frame timeouts (6s), clear `detected_signal` to stop the loop
4. **Thread cleanup**: Clean up stopped threads in `video_start_streaming()` to allow UI restart
5. **Add more logs**: Add detailed logging for timeout analysis

### DIFF
--- a/internal/native/cgo/video.c
+++ b/internal/native/cgo/video.c
@@ -662,8 +662,8 @@ void *run_video_stream(void *arg)
                 log_error("failure VIDIOC_QBUF: %s", strerror(errno));
         }
     cleanup:
-        log_info("cleaning up video capture device %s (attempt=%d, will loop back)",
-                 VIDEO_DEV, open_attempts);
+        log_info("cleaning up video capture device %s (attempt=%d, num_frames=%u)",
+                 VIDEO_DEV, open_attempts, num);
         if (ioctl(video_dev_fd, VIDIOC_STREAMOFF, &type) < 0)
         {
             log_error("VIDIOC_STREAMOFF failed: %s", strerror(errno));
@@ -693,6 +693,18 @@ void *run_video_stream(void *arg)
 
         log_info("closing video capture device %s (attempt=%d completed)", VIDEO_DEV, open_attempts);
         close(video_dev_fd);
+
+        // If we successfully captured frames, reset the attempt counter
+        if (num > 30) {
+            log_info("Successfully captured %u frames, resetting attempt counter", num);
+            open_attempts = 0;
+        }
+        // If we got 0 frames 3 times in a row, clear the stale signal flag
+        else if (num == 0 && open_attempts >= 3) {
+            log_warn("3 consecutive zero-frame timeouts - clearing stale detected_signal");
+            detected_signal = false;
+        }
+
         log_info("==== OUTER LOOP END: going back to start (streaming_flag=%d) ====", streaming_flag);
     }
 
@@ -736,11 +748,15 @@ void video_start_streaming()
     if (streaming_thread != NULL)
     {
         if (streaming_stopped == true) {
-            log_error("video streaming already stopped but streaming_thread is not NULL");
-            assert(streaming_stopped == true);
+            log_info("cleaning up stopped thread before starting new one");
+            pthread_join(*streaming_thread, NULL);
+            free(streaming_thread);
+            streaming_thread = NULL;
+            // Fall through to start new thread
+        } else {
+            log_warn("video streaming already started");
+            return;
         }
-        log_warn("video streaming already started");
-        return;
     }
 
     pthread_t *new_thread = malloc(sizeof(pthread_t));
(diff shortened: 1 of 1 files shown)


internal/native/cgo/video.c
IDisposable, line 800: is it worthwhile doing an `exit(1)` here too?

internal/native/cgo/video.c
```diff
@@ -395,6 +399,13 @@ void *run_video_stream(void *arg)
             continue;
         }
 
+        open_attempts++;
+        if (open_attempts > MAX_OPEN_ATTEMPTS) {
+            log_error("Tried to open video device %d times without success. "
+                     "Exiting to trigger supervisor restart.", MAX_OPEN_ATTEMPTS);
+            exit(1);
```
IDisposable, line 410: Maybe better as `return (void *)errno` (as that contains the error number from the last failed operation)

internal/native/cgo/video.c
IDisposable, line 557: By this point, all the setup work succeeded, but we still haven't gotten a valid frame, but we HAVE allocated memory that must be released.

IDisposable, line 663: `else open_attempts = 0;` because we've safely handled a frame? Otherwise we could slowly accumulate failing counts, while actually still running mostly right.

IDisposable, line 644: Do we want a cap on the attempts to send the frame too? Seems like we could get hung up here otherwise.

IDisposable, line 484: This memory allocation could be leaked in some exit paths...

IDisposable, line 518: Instead of 
```c
close(video_dev_fd);
return (void*)errno;
```

To drop out of the loop safely and ensure cleanup this should be just

```c
goto cleanup;
```

IDisposable, line 527: Same https://github.com/jetkvm/kvm/pull/998/files#r2548119345

IDisposable, line 667: To make it safe for everyone to shut-down safely with a `goto cleanup;` we need to wrap all the `ioctl(` and `close(` calls in test

`if (video_dev_fd > 0 && stream_on && ioctl(video_dev_fd, VIDIOC_STREAMOFF, &type) < 0)`

IDisposable, line 476: If we fail on allocating the second (or later) buffer after allocating some already, we leak those permanently. Would it be better to `goto cleanup;` here and ensure the tear-down has enough init-state information to free/close everything safely?

IDisposable, line 477: Doesn't the `close` call overwrite the `errno`?

IDisposable, line 457: _note to self_ input_buffer_count  == 3...

IDisposable, line 488: `goto cleanup;`

IDisposable, line 499: `goto cleanup;`

IDisposable, line 432: `goto cleanup;`

IDisposable, line 524: `stream_on = true;`

internal/native/cgo/video.c
```diff
@@ -387,6 +387,10 @@ void *run_video_stream(void *arg)
 
     streaming_stopped = false;
```
IDisposable, line 389: `bool stream_on == false;`

internal/native/cgo/video.c
IDisposable, line 640: Since this never happens if we `return` out as some paths currently do, we would leak memory if the memory was allocated and we return out...

IDisposable, line 629: _note to self_ This actually allocates a zero-count of buffers, dropping what we had above

`if (video_dev_fd > 0 &&  ioctl(video_dev_fd, VIDIOC_REQBUFS, &req_free) < 0)`

IDisposable (commented):
While I understand the idea of this change is to give up completely if we can't safely open things up, a review of this code shows a lot of issues with resources that might be partially setup not being torn down correctly.

adamshiervani:
When we enter sleep mode, we get a signal but no frames. When we request a frame, it times out after a second and then retries forever. 

The proper fix is to stop the video before we enter sleep mode, so we don't need to handle this special case. See https://github.com/jetkvm/kvm/pull/999

PR #1007, opened 2025-11-25
feat: show confirmation dialog deleting WOL target

### LINKED ISSUE #1006: Confirmation dialog deleting WOL targets
### A note for the community

Show confirmation dialog when deleting "Wake on LAN" targets

### Disclaimer

### Subsystem

None

### Feature description

Posted in feature-requests on Discord by another user.

> I just had to delete a WOL target cos it didnt end up working with WOL, but i realised it just deleted it without confirming.
Given the wake and delete buttons are close to eachother i think it would be nice to confirm to delete things, espeically due to the fact that if you're using WOL, you probably wont necessarily have access to something to check its mac address whilst you're away.

Jonher937:
Closes: #1006

### DIFF
--- a/ui/src/components/popovers/WakeOnLan/DeviceList.tsx
+++ b/ui/src/components/popovers/WakeOnLan/DeviceList.tsx
@@ -1,4 +1,5 @@
 import { LuPlus, LuSend, LuTrash2 } from "react-icons/lu";
+import { useState } from "react";
 
 import { m } from "@localizations/messages.js";
 import { Button } from "@components/Button";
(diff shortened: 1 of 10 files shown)


ui/src/components/popovers/WakeOnLan/DeviceList.tsx
```diff
@@ -55,14 +73,30 @@ export default function DeviceList({
...
         </div>
       </Card>
+      {deleteIndex !== null && (
+        <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/40">
+          <Card className="p-6 w-full max-w-xs animate-fadeIn">
+            <h3 className="text-lg font-semibold mb-2 text-slate-900 dark:text-slate-100">
+              {m.wake_on_lan_device_list_delete_device()}
+            </h3>
+            <p className="text-sm mb-4 text-slate-700 dark:text-slate-300">
+              {m.wake_on_lan_device_list_confirm_delete_message()}
```
IDisposable: Might be nice to insert the name of the device being deleted. The syntax is `{name}` in the language literals, then pass it in the message call (see any of the error uses)

IDisposable (changes requested):

ui/src/components/popovers/WakeOnLan/DeviceList.tsx
Jonher937: Good suggestion.
Added name after the questionmark in all translation strings. Verified it works correctly with multiple entries and defaults to a empty string should it fail to lookup the index (shoud never happen).

IDisposable (approved):

PR #1009, opened 2025-11-27
feat: redirect to HTTPS if TLS mode is enabled

fayusohenson:


ym:
HTTPS redirection should be optional instead of mandatory.
Enforcing it could lock users out of their device entirely if HTTPS is misconfigured.

IDisposable:
We could gate this behind an option like the "restrict to localhost" option.

PR #1035, opened 2025-12-02
Fix USB Gadget polling/writing to recover from errors

### LINKED ISSUE #128: USB not detected if not present at startup
App: 0.3.4
System: 0.2.0

My JetKVM does not recognize the USB connection if it's already powered on before the USB is connected. For example, if I power on the device using the RJ-11 jack from the ATX module and then plug in the USB, it never recognizes it. But, if I power it on using the USB first it works just fine. Strangely, shutting down the PC and then turning it back on re-establishes the USB link just fine, but simply unplugging and replugging it back in does not. I'm unsure if this is the intended behavior, but it doesn't seem like that would make much sense.

IDisposable:
Fixes #128

### Summary
- Revamped the polling and writing of the USB gadget to correct for dropped connections
- Corrected the initialization order to ensure we run even if the USB is disconnected (and we're getting power from the y-splitter/POE)
- Ensured that we restart the LED listening polling
- If we exceed too many consecutive write errors, close and reopen the gadget
- Extensive changes to logging to ensure we have complete context and honor the changing of log levels in the config.
- Fixed the script to see all JETKVM_LOG_* log-level commands.
- Handle the native C code error returns in the CGO wrapper
- MORE TO COME


internal/hidrpc/hidrpc.go
```diff
@@ -6,24 +6,6 @@ import (
 	"github.com/jetkvm/kvm/internal/usbgadget"
 )
 
-// MessageType is the type of the HID RPC message
```
IDisposable, line 9: Moved this to message.go where it really makes more sense

internal/hidrpc/message.go
```diff
@@ -11,9 +31,22 @@ type Message struct {
 	d []byte
 }
 
+func (m *Message) MarshalZerologObject(e *zerolog.Event) {
```
IDisposable: Allow zerolog to serialize the object cleanly and fast without reflection if you use `.Object()`

internal/hidrpc/message.go
```diff
@@ -11,9 +31,22 @@ type Message struct {
...
+	e.Uint8("type", uint8(m.t))
+	e.Bytes("payload", m.d)
+}
+
 // Marshal marshals the message to a byte array.
 func (m *Message) Marshal() ([]byte, error) {
-	return Marshal(m)
+	if m.t == 0 {
```
IDisposable, line 41: Just inlined the _used nowhere else_ code for doing the Marshal

internal/logging/context.go
```diff
@@ -0,0 +1,279 @@
+package logging
```
IDisposable: Please hold on commentary on this ... I did this while trying to make the whole logging system more fluent and less prone to multi-threaded problems... that said, this is kind ugly in that it duplicates a lot of the "odd" decisions that were in zerolog... I want to reduce this code down **a lot** before we merge.

internal/logging/logger.go
```diff
@@ -12,10 +12,10 @@ import (
 )
 
 type Logger struct {
-	l               *zerolog.Logger
-	scopeLoggers    map[string]*zerolog.Logger
-	scopeLevels     map[string]zerolog.Level
-	scopeLevelMutex sync.Mutex
+	defaultLogger *zerolog.Logger
```
IDisposable: Call it what it is!

internal/logging/logger.go
```diff
@@ -24,6 +24,7 @@ type Logger struct {
 
 const (
 	defaultLogLevel = zerolog.ErrorLevel
+	unset           = -2
```
IDisposable, line 28: This makes the `-2` pulled-from-whole-cloth constant  only in one place...

internal/logging/logger.go
```diff
@@ -73,125 +74,145 @@ var (
 		"INFO":    zerolog.InfoLevel,
 		"DEBUG":   zerolog.DebugLevel,
 		"TRACE":   zerolog.TraceLevel,
+		"UNSET":   unset,
```
IDisposable, line 80: You can UNSET the level via the environment variables now... so you could do `JETKVM_LOG_INFO=all` to set everything to `zerolog.InfoLevel` followed by a `JETKVM_LOG_UNSET=mdns` to make that noisy feller not force a level.

Scope-specific environment levels override the Config and "All" environment settings currently.

Note: I still need to make sure that we do the "levels" in the right order so we can decide on the hierarchy... but it's not bad now...

internal/logging/logger.go
```diff
@@ -73,125 +74,145 @@ var (
...
 	return &Logger{
-		l:                         &zerologLogger,
+		defaultLogger:             &zerologLogger,
 		scopeLoggers:              make(map[string]*zerolog.Logger),
 		scopeLevels:               make(map[string]zerolog.Level),
-		scopeLevelMutex:           sync.Mutex{},
-		defaultLogLevelFromEnv:    -2,
-		defaultLogLevelFromConfig: -2,
+		loggerMutex:               sync.Mutex{},
+		defaultLogLevelFromEnv:    unset,
```
IDisposable, line 90: Use the constant

internal/logging/logger.go
```diff
@@ -73,125 +74,145 @@ var (
...
+		defaultLogLevelFromConfig: unset,
 		defaultLogLevel:           defaultLogLevel,
 	}
 }
 
-func (l *Logger) updateLogLevel() {
-	l.scopeLevelMutex.Lock()
-	defer l.scopeLevelMutex.Unlock()
+func (l *Logger) updateLogLevels(newConfigLevel zerolog.Level) {
```
IDisposable, line 96: So this now works again... calculates the level for each "known" scope, and any explicitly declared (in environment variables) scopes... and then also determines the "base all" level for config and environment to apply when _any_scope logger is created...

internal/logging/logger.go
```diff
@@ -73,125 +74,145 @@ var (
...
-
-		if env == "" {
-			env = os.Getenv(fmt.Sprintf("PIONS_LOG_%s", name))
-		}
-
-		if env == "" {
-			continue
+		if env != "" {
+			env = strings.ToLower(env)
+			loopLogger := logger.With().
```
IDisposable, line 108: ... huh, inception... logging out the level setting of the logging infrastructure... weeeeee

internal/logging/logger.go
```diff
@@ -73,125 +74,145 @@ var (
...
-	if !ok || logger == nil {
-		scopeLogger := l.newScopeLogger(scope)
-		l.scopeLoggers[scope] = &scopeLogger
+	// double-check after acquiring the lock
+	if logger, ok := l.scopeLoggers[scope]; ok && logger != nil {
+		return logger
 	}
 
-	return l.scopeLoggers[scope]
+	scopeLevel := l.getScopeLoggerLevel(scope)
```
IDisposable, line 179: If it's a logging scope we haven't seen... we need to calculate from "defaults" then override from scope-specific env settings.

internal/logging/logger.go
```diff
@@ -73,125 +74,145 @@ var (
...
+	defer l.loggerMutex.Unlock()
 
-	if needUpdate {
-		for scope, logger := range l.scopeLoggers {
-			currentLevel := logger.GetLevel()
-			targetLevel := l.getScopeLoggerLevel(scope)
-			if currentLevel != targetLevel {
-				*logger = l.newScopeLogger(scope)
-			}
+	l.updateLogLevels(newConfigLevel)
```
IDisposable, line 207: When you change the config-driven log level (not just at startup!) we can now destroy our cache of any affected logger so we recreate a new logger when asked again ... with the correct level.

internal/logging/pion.go
```diff
@@ -48,8 +50,8 @@ func (c pionLogger) Errorf(format string, args ...any) {
 type pionLoggerFactory struct{}
 
 func (c pionLoggerFactory) NewLogger(subsystem string) logging.LeveledLogger {
-	logger := rootLogger.getLogger(subsystem).With().
-		Str("scope", "pion").
+	logger := GetSubsystemLogger("pion").
```
IDisposable: This is the first time the actual resolution to the logging level changing works... since a `zerolog.Logger` **cannot** change levels... we can't hold pointers to them for any length of time... so the sweeping change is to make all places where we would have used the global loggers to always ask for a desired logger by name... and this will be satisfied from the map cache... and when levels need to change, we just invalidate it out of the cache... so it gets created with the correct level

internal/logging/root.go
```diff
@@ -3,16 +3,16 @@ package logging
...
+		zerolog.New(defaultLogOutput).
+			With().
+			Str("scope", "jetkvm").
+			Timestamp().
+			Logger())
 )
 
-func GetRootLogger() *Logger {
-	return rootLogger
+func UpdateConfigLogLevel(logLevel string) {
```
IDisposable, line 14: Nobody should every talk to the root logger... except to change the config-driven default level... so that's all we expose now

internal/logging/utils.go
```diff
@@ -2,31 +2,17 @@ package logging
...
+	"sync"
 )
 
-var defaultLogger = zerolog.New(os.Stdout).Level(zerolog.InfoLevel)
-
-func GetDefaultLogger() *zerolog.Logger {
-	return &defaultLogger
+func (context *Context) UnlockWithTraceLog(lock *sync.Mutex, msg string, args ...any) {
+	defer lock.Unlock()
+	context.Trace().Msgf(msg, args...)
```
IDisposable: Just a basic streamlining here.

internal/mdns/log.go
```diff
@@ -0,0 +1,12 @@
+package mdns
+
+import (
+	"github.com/jetkvm/kvm/internal/logging"
+)
+
+func (m *MDNS) getMdnsLoggingContext() *logging.Context {
```
IDisposable: All internal "sub packages" now have a log.go if they do logging... and it is where the scope-specific logger name is singletoned.

Note: the `logging.NewContext(` part is probably going to be streamlined away before I mark this ready to go...

internal/mdns/mdns.go
```diff
@@ -9,15 +9,13 @@ import (
...
 	pion_mdns "github.com/pion/mdns/v2"
-	"github.com/rs/zerolog"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
 )
 
 type MDNS struct {
 	conn *pion_mdns.Conn
 	lock sync.Mutex
-	l    *zerolog.Logger
```
IDisposable, line 21: It is NOT legal to hold onto loggers for any duration... so sweeping to eliminate these and always ask from the controlling log.go

internal/mdns/mdns.go
```diff
@@ -75,8 +67,10 @@ func (m *MDNS) start(allowRestart bool) error {
 		return fmt.Errorf("listen options not set")
 	}
 
+	context := m.getMdnsLoggingContext()
```
IDisposable: going to rename all the context back to logger as I finalize... this was my evolutionary appendix.

internal/mdns/mdns.go
```diff
@@ -135,13 +123,15 @@ func (m *MDNS) start(allowRestart bool) error {
 		LoggerFactory: logging.GetPionDefaultLoggerFactory(),
 	})
 
+	context = context.Interface("mDNSConn", mDNSConn).Err(err)
+
 	if err != nil {
-		scopeLogger.Warn().Err(err).Msg("failed to start mDNS server")
+		context.Err(err).Error().Msg("failed to start mDNS server")
```
IDisposable: ... and it lead to decisions like this one... where we shouldn't really be adding context to a context unless we've `.With()` it to dupe... so this is technically wrong and will be put back to `logger.Error().Err(err).Msg(` before I'm done...

internal/native/cgo/edid.c
```diff
@@ -81,32 +81,28 @@ int set_edid(uint8_t *edid, size_t size)
     if (size != 128 && size != 256)
     {
         errno = EINVAL;
-        return -1;
+        return -2;
```
IDisposable, line 84: Make each error code distinct in the C code so the CGO code can log correct issues.

internal/native/cgo_linux.go
```diff
@@ -143,7 +142,7 @@ func videoInit(factor float64) error {
 
 	ret := C.jetkvm_video_init(factorC)
 	if ret != 0 {
-		return fmt.Errorf("failed to initialize video: %d", ret)
+		return fmt.Errorf("failed to initialize video with factor %f: %d", factor, ret)
```
IDisposable, line 145: ... just a bit more detail added to logs as I scooted past ones that stuck out

internal/native/cgo_linux.go
```diff
@@ -268,8 +269,12 @@ func uiObjAddFlag(objName string, flag string) (bool, error) {
 	defer C.free(unsafe.Pointer(objNameCStr))
 	flagCStr := C.CString(flag)
 	defer C.free(unsafe.Pointer(flagCStr))
-	C.jetkvm_ui_add_flag(objNameCStr, flagCStr)
-	return true, nil
+
+	ret := C.jetkvm_ui_add_flag(objNameCStr, flagCStr)
```
IDisposable, line 273: Handle the error returns correctly (was silently ignored)

internal/native/chan.go
```diff
@@ -36,11 +36,11 @@ func (n *Native) handleVideoStateChan() {
 func (n *Native) handleLogChan() {
 	for {
 		entry := <-logChan
-		l := n.l.With().
+		l := GetNativeLogger().
```
IDisposable, line 38: This is much closer to the final changes... mostly swapping member variables with calls to a function

internal/native/display.go
```diff
@@ -99,18 +99,21 @@ func (n *Native) DisplaySetRotation(rotation uint16) (bool, error) {
 
 // UpdateLabelIfChanged updates the label if the text has changed
 func (n *Native) UpdateLabelIfChanged(objName string, newText string) {
-	l := n.lD.Trace().Str("obj", objName).Str("text", newText)
+	logger := GetDisplayLogger().
```
IDisposable, line 102: for short durations (e.g. inside a method) it's okay to use the logger repeatedly... but we need to be vary careful for callbacks, goroutines, selects, and for {} loops... because they need to ask for their own logger at time of _use_

internal/native/grpc_client.go
```diff
@@ -42,12 +42,15 @@ type GRPCClient struct {
...
 type grpcClientOptions struct {
 	SocketPath         string
-	Logger             *zerolog.Logger
 	OnVideoStateChange func(state VideoState)
 	OnIndevEvent       func(event string)
 	OnRpcEvent         func(event string)
 }
 
+func (client *GRPCClient) getLogger() *logging.Context {
+	return getClientLogger().With().Str("target", client.conn.Target()).Bool("closed", client.closed)
```
IDisposable: This is the pattern for augmented contexts... adding it from instance values.

internal/native/grpc_client.go
```diff
@@ -121,11 +123,11 @@ func (c *GRPCClient) handleEventStream(stream pb.NativeService_StreamEventsClien
 		}
 
 		// enrich the logger with the event type and data, if debug mode is enabled
-		if c.logger.GetLevel() <= zerolog.DebugLevel {
-			logger = logger.With().
+		if logger.IsDebugLevel() {
+			logger = logger.
```
IDisposable: It's explicitly legal to replace logger with another logger to augment context... but you **have** to call `.With()` to get a copied-before-write of your own.

internal/native/grpc_server.go
```diff
@@ -6,28 +6,28 @@ import (
...
+	"github.com/jetkvm/kvm/internal/logging"
 	pb "github.com/jetkvm/kvm/internal/native/proto"
 )
 
 // grpcServer wraps the Native instance and implements the gRPC service
 type grpcServer struct {
 	pb.UnimplementedNativeServiceServer
 	native            *Native
-	logger            *zerolog.Logger
+	socketPath        string
```
IDisposable, line 19: Replace the specific long-held logger with the information needed to reconstitute a logger with that context

internal/native/log.go
```diff
@@ -1,12 +1,21 @@
...
 import (
+	"os"
+
 	"github.com/jetkvm/kvm/internal/logging"
 	"github.com/rs/zerolog"
 )
 
-var nativeLogger = logging.GetSubsystemLogger("native")
-var displayLogger = logging.GetSubsystemLogger("display")
+var pid = os.Getpid()
```
IDisposable, line 10: For the native stuff... always want the pid in the logs

internal/native/native.go
```diff
@@ -61,61 +56,49 @@ func (s VideoStreamingStatus) String() string {
...
+		videoLock:            sync.Mutex{},
 	}
 
-	onVideoFrameReceived := opts.OnVideoFrameReceived
-	if onVideoFrameReceived == nil {
-		onVideoFrameReceived = func(frame []byte, duration time.Duration) {
-			nativeLogger.Trace().Interface("frame", frame).Dur("duration", duration).Msg("video frame received")
+	if n.onVideoStateChange == nil {
+		n.onVideoStateChange = func(state VideoState) {
+			GetDisplayLogger().Info().Interface("state", state).Msg("video state changed")
```
IDisposable, line 75: Here's an example of callbacks needing to ask for the logger at callback-time instead of whatever it was when this handler was setup.

internal/native/server.go
```diff
@@ -78,82 +86,104 @@ func RunNativeProcess(binaryName string) {
 	appCtx, appCtxCancel := context.WithCancel(context.Background())
 	defer appCtxCancel()
 
-	logger := nativeLogger.With().Int("pid", os.Getpid()).Logger()
 	setProcTitle("starting")
+	logger := getServerLogger()
+
+	// for defer clean-up scoping... this is NOT a goroutine
+	func() {
```
IDisposable, line 91: I made this hunk a self-scoped function to allow proper use of the `defer` to stop/close/release resources. Also added a timeboxed attempt at graceful shutdown first.

internal/native/server.go
```diff
@@ -78,82 +86,104 @@ func RunNativeProcess(binaryName string) {
...
+		}
+		logger.Debug().Msg("wrote handshake message for supervisor")
+
+		// Set up signal handling
+		sigChan := make(chan os.Signal, 1)
+		signal.Notify(sigChan, syscall.SIGTERM, syscall.SIGINT)
+
+		// Wait for signal
+		sig := <-sigChan
+		logger.Info().Str("signal", sig.String()).Msg("received termination signal")
```
IDisposable, line 183: Seems like this sort of thing is so common there really ought to be an idiomatic way of saying "try this method for Dur time" and then give up...

internal/ota/errors.go
```diff
@@ -4,19 +4,16 @@ import (
...
 	// ErrVersionNotFound is returned when the specified version is not found
 	ErrVersionNotFound = errors.New("specified version not found")
 )
 
-func (s *State) componentUpdateError(prefix string, err error, l *zerolog.Logger) error {
-	if l == nil {
-		l = s.l
-	}
-	l.Error().Err(err).Msg(prefix)
+func (s *State) componentUpdateError(prefix string, err error, loggingContext *logging.Context) error {
```
IDisposable: This one is okay... these methods are called synchronously.

internal/ota/ota.go
```diff
@@ -349,42 +355,38 @@ func (s *State) checkUpdateStatus(
...
-	if params.RequestID != "" {
-		logger = logger.With().Str("requestID", params.RequestID).Logger()
-	}
-	t := time.Now()
-
-	logger.Trace().
-		Str("appVersionLocal", appVersionLocal.String()).
-		Str("systemVersionLocal", systemVersionLocal.String()).
+	loggingContext.
+		Stringer("appVersionLocal", appVersionLocal).
```
IDisposable, line 376: `.Stringer("foo", foo)` is the same as `.Str("foo", foo.String())` except that the latter always runs the `.String()` method, while the former can elid that call overhead if the level filter is going to toss the contextual information because we're not emitting the final message... 

BUT ONLY if you're adding it _after_ the `.Info()` (etc.) call.

internal/ota/ota_test.go
```diff
@@ -173,15 +172,8 @@ func newOtaState(d *testData, t *testing.T) *State {
 		return systemVersion, appVersion, nil
 	}
 
-	traceLevel := zerolog.InfoLevel
-
-	if os.Getenv("TEST_LOG_TRACE") == "1" {
```
IDisposable, line 178: This can really be done with `export JETKVM_LOG_TRACE=ota` before running the test now.

internal/ota/sys.go
```diff
@@ -14,42 +14,48 @@ const (
...
+		return s.componentUpdateError("Error verifying system update hash", err, loggingContext)
 	}
 	verifyFinished := time.Now()
+	loggingContext.Dur("verification_time", verifyFinished.Sub(verifyStarted)).Info().Msg("update verified")
+
 	systemUpdate.verifiedAt = verifyFinished
 	systemUpdate.verificationProgress = 1
-	systemUpdate.updatedAt = verifyFinished
-	systemUpdate.updateProgress = 1
+	systemUpdate.updatedAt = verifyFinished // TODO, this seems wrong here
```
IDisposable, line 48: **COMMENT NEEDED** this seems odd...

internal/sync/once.go
```diff
@@ -13,6 +13,10 @@ type Once struct {
 
 // Do calls the function f if and only if Do has not been called before for this instance of Once.
 func (o *Once) Do(f func()) {
-	logTrace("Doing once")
-	o.mu.Do(f)
+	g := func() {
+		logTrace("Doing once")
+		f()
```
IDisposable, line 18: I **fixed it**... now we can see when the onceler is onced.

internal/timesync/http.go
```diff
@@ -43,13 +43,12 @@ func (t *TimeSync) queryMultipleHttp(urls []string, timeout time.Duration) (now
...
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
+	loggingContext := t.loggingContext.With().Int("urls_count", len(urls)).Int("timeout_ms", int(timeout.Milliseconds()))
 
 	for _, url := range urls {
-		go func(url string) {
-			scopedLogger := t.l.With().
-				Str("http_url", url).
-				Logger()
+		loopContext := loggingContext.With().Str("http_url", url)
```
IDisposable: This is the correct way to handle context in loops... so each loop gets it's on context without pollution from the previous loop pass...

internal/usbgadget/config.go
```diff
@@ -61,26 +61,26 @@ var defaultGadgetConfig = map[string]gadgetConfigItem{
 	"mass_storage_lun0": massStorageLun0Config,
 }
 
-func (u *UsbGadget) isGadgetConfigItemEnabled(itemKey string) bool {
+func (enabledDevices *Devices) isGadgetConfigItemEnabled(itemKey string) bool {
```
IDisposable, line 64: Cleaned up the double-dotted envy here... the method doesn't care about `UsbGadget`s... it cares about `Device`s

internal/usbgadget/config.go
```diff
@@ -115,15 +115,6 @@ func (u *UsbGadget) SetGadgetDevices(devices *Devices) {
 	u.enabledDevices = *devices
 }
 
-// GetConfigPath returns the path to the config item.
```
IDisposable, line 123: Never used?

internal/usbgadget/config.go
```diff
@@ -193,19 +193,22 @@ func (u *UsbGadget) UpdateGadgetConfig() error {
...
 	return nil
 }
 
 func (u *UsbGadget) configureUsbGadget(resetUsb bool) error {
-	return u.WithTransaction(func() error {
-		u.tx.MountConfigFS()
-		u.tx.CreateConfigPath()
-		u.tx.WriteGadgetConfig()
+	return u.WithTransaction(func(u *UsbGadget, tx *UsbGadgetTransaction) error {
+		tx.MountConfigFS()
```
IDisposable, line 205: This logging crashed hard in some weird timing related ways... turns out that we really need to not have the transaction as a member of the gadget... rather treat it as it's own entity.

internal/usbgadget/config_tx.go
```diff
@@ -6,68 +6,42 @@ import (
...
 )
 
 // no os package should occur in this file
 
 type UsbGadgetTransaction struct {
-	c *ChangeSet
-
-	// below are the fields that are needed to be set by the caller
-	log                       *zerolog.Logger
-	udc                       string
```
IDisposable, line 19: Copying all this down from the gadget isn't desirable... we just pass both the gadget and the transaction into the methods that need that data.

internal/usbgadget/config_tx.go
```diff
@@ -6,68 +6,42 @@ import (
...
-		return err
-	}
-	if err := fn(); err != nil {
-		u.log.Error().Err(err).Msg("transaction failed")
+	context := u.getUsbGadgetLoggingContext().Str("udc", u.udc)
+	context.Info().Msg("starting USB gadget transaction")
+
+	tx := u.newUsbGadgetTransaction(context)
+
+	if err := fn(u, tx); err != nil {
```
IDisposable, line 35: ... here..... passing both tx and u

internal/usbgadget/config_tx.go
```diff
@@ -131,84 +105,71 @@ func (tx *UsbGadgetTransaction) MountConfigFS() {
...
 		"gadget",
-		tx.configC1Path,
+		configC1Path,
 		"create config path",
 		[]string{configFSPath},
 	)
 }
 
-func (tx *UsbGadgetTransaction) WriteGadgetConfig() {
+func (tx *UsbGadgetTransaction) WriteGadgetConfig(kvmGadgetPath string, configC1Path string, udc string, orderedConfigItems orderedGadgetConfigItems, enabledDevices *Devices) {
```
IDisposable, line 127: ... as Gary would say... "tell, don't ask"

internal/usbgadget/changeset.go
```diff
@@ -194,15 +195,15 @@ func (fc *FileChange) checkIfDirIsMountPoint() error {
 }
 
 // GetActualState returns the actual state of the file at the given path.
-func (fc *FileChange) getActualState() error {
-	l := defaultLogger.With().Str("path", fc.Path).Logger()
+func (fc *FileChange) getActualState(context *logging.Context) error {
```
IDisposable: AHHHHH, into the meat of the actual fix!

The transactions around tearing down and setting up the UsbGadget were not terribly stable, _especially_ in the power-on before usb-connected case... I basically streamline this entire base to get it to not crash out the process... ironically, it was all my other changes to logging that helped me find the nil-pointer accesses and corrupt sequences.

internal/usbgadget/config_tx.go
```diff
@@ -131,84 +105,71 @@ func (tx *UsbGadgetTransaction) MountConfigFS() {
...
 		gadgetItemDir := tx.mkdirAll(component, gadgetItemPath, "create gadget item directory", files)
 		files = append(files, gadgetItemDir)
 	}
 
 	beforeChange := make([]string, 0)
 	disableGadgetItemKey := fmt.Sprintf("disable-%s", item.device)
+
 	if item.configPath != nil && item.configAttrs == nil {
-		beforeChange = append(beforeChange, tx.getDisableKeys()...)
+		beforeChange = append(beforeChange, disableGadgetItemKey)
```
IDisposable, line 182: This was one of the NPEs...

internal/usbgadget/hid_keyboard.go
```diff
@@ -93,6 +96,10 @@ type KeyboardState struct {
 	raw        byte
 }
 
+func (s KeyboardState) MarshalZerologObject(e *zerolog.Event) {
```
IDisposable, line 101: So zerolog can `.Object(` marshal these...

internal/usbgadget/hid_keyboard.go
```diff
@@ -115,19 +122,25 @@ func (u *UsbGadget) updateKeyboardState(state byte) {
 	u.keyboardStateLock.Lock()
 	defer u.keyboardStateLock.Unlock()
 
+	context := u.getHidKeyboardLoggingContext().Uint8("state", state)
+
 	if state&^ValidKeyboardLedMasks != 0 {
-		u.log.Warn().Uint8("state", state).Msg("ignoring invalid bits")
-		return
+		context.Warn().Msg("ignoring invalid bits")
+		state &= ValidKeyboardLedMasks
```
IDisposable, line 131: This is sort of an unrelated change... but if the host OS sends us extra bits... we should ignore them... not ignore the entire status.

internal/usbgadget/hid_keyboard.go
```diff
@@ -115,19 +122,25 @@ func (u *UsbGadget) updateKeyboardState(state byte) {
...
 		return
 	}
-	u.log.Trace().Uint8("old", u.keyboardState).Uint8("new", state).Msg("keyboardState updated")
+
 	u.keyboardState = state
+	context.Trace().Msg("updated keyboardState")
 
-	if u.onKeyboardStateChange != nil {
-		(*u.onKeyboardStateChange)(getKeyboardState(state))
+	if cb := u.onKeyboardStateChange; cb != nil {
```
IDisposable, line 144: HUGE change here... we need to ensure that when we trigger report-back transactions that they get emitted asynchronously so the host and device can get on with life while the network and/or client is tardy.

internal/usbgadget/hid_keyboard.go
```diff
@@ -149,6 +162,17 @@ func (u *UsbGadget) GetKeysDownState() KeysDownState {
 	return u.keysDownState
 }
 
+func (u *UsbGadget) ResetRollover() {
```
IDisposable, line 167: This code was corrupted when we did the auto-release changes. We need to _keep_ the hidErrorRollOver bytes intact until _after_ we've written them to the host **and** reported them back to the client ... so this needs to be `defer`-called from two places... hence extracted the logic.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
 // DefaultAutoReleaseDuration is the default duration for auto-release of a key.
 const DefaultAutoReleaseDuration = 100 * time.Millisecond
 
-func (u *UsbGadget) scheduleAutoRelease(key byte) {
+func (u *UsbGadget) DelayAutoReleaseWithDuration(resetDuration time.Duration) {
+	context := u.getHidKeyboardAutoReleaseLoggingContext().Dur("reset_duration", resetDuration)
+
 	u.kbdAutoReleaseLock.Lock()
-	defer unlockWithLog(&u.kbdAutoReleaseLock, u.log, "autoRelease scheduled")
+	defer context.UnlockWithTraceLog(&u.kbdAutoReleaseLock, "auto-release delayed")
```
IDisposable: I really like this method!

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
-	defer unlockWithLog(&u.kbdAutoReleaseLock, u.log, "autoRelease cancelled")
+	timer, ok := u.kbdAutoReleaseTimers[key]
+	if ok {
+		if timer != nil {
+			timer.Stop()
+		}
 
-	if timer := u.kbdAutoReleaseTimers[key]; timer != nil {
-		timer.Stop()
-		u.kbdAutoReleaseTimers[key] = nil
```
IDisposable, line 189: We don't need to store a nil, we're under lock and going to delete the entry in the next line.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
-	// We currently hardcode the duration to 100ms
-	// However, it should be the same as the duration of the keep-alive reset called baseExtension.
-	u.kbdAutoReleaseTimers[key] = time.AfterFunc(100*time.Millisecond, func() {
-		u.performAutoRelease(key)
-	})
 }
 
-func (u *UsbGadget) cancelAutoRelease(key byte) {
+// note: lock must be freeed by caller
+func (u *UsbGadget) popAutoReleaseTimer(key byte) bool {
```
IDisposable, line 203: This is the common code in two paths... we grab the entry and if found, stop it... and then in either case remove if from the map.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
-		if timer != nil {
-			timer.Reset(resetDuration)
+func (u *UsbGadget) cancelAutoRelease(key byte) {
+	ok := u.popAutoReleaseTimer(key)
+	context := u.getHidKeyboardAutoReleaseLoggingContext().Uint8("key", key)
+	defer context.UnlockWithTraceLog(&u.kbdAutoReleaseLock, "auto-release cancelled")
+
+	if ok {
+		// Reset keep-alive timing when key is actually released
+		if cb := u.onKeepAliveReset; cb != nil {
```
IDisposable, line 238: Async-safe callback pattern...

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
 
 func (u *UsbGadget) listenKeyboardEvents() {
-	var path string
-	if u.keyboardHidFile != nil {
-		path = u.keyboardHidFile.Name()
-	}
-	l := u.log.With().Str("listener", "keyboardEvents").Str("path", path).Logger()
-	l.Trace().Msg("starting")
-
-	go func() {
```
IDisposable, line 278: moved the goroutine call to the caller of this method as it's much clearer.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
 
-	var err error
-	u.keyboardHidFile, err = os.OpenFile("/dev/hidg0", os.O_RDWR, 0666)
-	if err != nil {
-		return fmt.Errorf("failed to open hidg0: %w", err)
-	}
-
 	if u.keyboardStateCancel != nil {
 		u.keyboardStateCancel()
+		u.keyboardStateCancel = nil
```
IDisposable, line 310: HUGE change.

We need to clear-down the context cancel on the previous instance _before_ we setup for the new file open.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
+
+	if keyboardFile, err := os.OpenFile("/dev/hidg0", os.O_RDWR, 0666); err == nil {
+		u.keyboardHidFile = keyboardFile
+	} else {
+		return fmt.Errorf("failed to open keyboard on hidg0: %w", err)
 	}
 
 	u.keyboardStateCtx, u.keyboardStateCancel = context.WithCancel(context.Background())
-	u.listenKeyboardEvents()
+	go u.listenKeyboardEvents()
```
IDisposable, line 320: This is the async go-routine change mentioned above.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +184,165 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
 		u.keyboardHidFile.Close()
 		u.keyboardHidFile = nil
 		return err
 	}
-	u.resetLogSuppressionCounter("keyboardWriteHidFile")
 	return nil
 }
 
 func (u *UsbGadget) UpdateKeysDown(modifier byte, keys []byte) KeysDownState {
-	// if we just reported an error roll over, we should clear the keys
```
IDisposable, line 429: This is the broken rollover logic... we can't clear the buffer yet... we haven't scheduled the bad news to the client.

internal/usbgadget/hid_keyboard.go
```diff
@@ -471,30 +479,34 @@ func (u *UsbGadget) keypressReport(key byte, press bool) (KeysDownState, error)
...
 			} else {
 				// If we are releasing a key, and we didn't find it in a slot, who cares?
-				l.Warn().Msg("key not found in buffer, nothing to release")
+				context.Trace().Msg("key not found in buffer, nothing to release")
 			}
 		}
 	}
 
 	err := u.keyboardWriteHidFile(modifier, keys)
+	defer u.ResetRollover()
```
IDisposable: ... so we clear the roll-over _after_ we send the UpdateKeysDown

internal/usbgadget/hid_mouse_absolute.go
```diff
@@ -97,12 +94,9 @@ func (u *UsbGadget) AbsMouseReport(x int, y int, buttons uint8) error {
 		byte(y),      // Y Low Byte
 		byte(y >> 8), // Y High Byte
 	})
-	if err != nil {
-		return err
-	}
 
 	u.resetUserInputTime()
-	return nil
+	return err
```
IDisposable, line 99: ... oops... dropped that err!

internal/usbgadget/hid_mouse_relative.go
```diff
@@ -58,20 +58,17 @@ var relativeMouseCombinedReportDesc = []byte{
 func (u *UsbGadget) relMouseWriteHidFile(data []byte) error {
 	if u.relMouseHidFile == nil {
 		var err error
-		u.relMouseHidFile, err = os.OpenFile("/dev/hidg2", os.O_RDWR, 0666)
-		if err != nil {
-			return fmt.Errorf("failed to open hidg1: %w", err)
+		if u.relMouseHidFile, err = os.OpenFile("/dev/hidg2", os.O_RDWR, 0666); err != nil {
+			return fmt.Errorf("failed to open hidg2: %w", err)
```
IDisposable, line 62: logging message was wrong :) hidg1 instead of hidg2

internal/usbgadget/usbgadget.go
```diff
@@ -172,6 +167,12 @@ func (u *UsbGadget) Close() error {
 	u.kbdAutoReleaseTimers = make(map[byte]*time.Timer)
 	u.kbdAutoReleaseLock.Unlock()
 
+	// Cancel keyboard state context
```
IDisposable: TODO This looks like a merge error... no reason that this block should be done twice

internal/usbgadget/utils.go
```diff
@@ -2,44 +2,15 @@ package usbgadget
...
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
-
-	"github.com/rs/zerolog"
 )
 
-type ByteSlice []byte
```
IDisposable, line 19: Moved for more public consumption

internal/usbgadget/utils.go
```diff
@@ -71,14 +42,21 @@ func hexToOctal(hex string) (string, error) {
 	return octal, nil
 }
 
+func compareIgnoreTrailingLF(shorter []byte, longer []byte) bool {
```
IDisposable, line 45: Clearer on intent.

internal/usbgadget/utils.go
```diff
@@ -112,67 +90,28 @@ func compareFileContent(oldContent []byte, newContent []byte, looserMatch bool)
...
+	fileName := file.Name()
+
 	if err := file.SetWriteDeadline(time.Now().Add(hidWriteTimeout)); err != nil {
 		return -1, err
 	}
 
 	n, err = file.Write(data)
 	if err == nil {
-		return
+		u.resetLogSuppressionCounter("writeWithTimeout_" + fileName)
```
IDisposable, line 101: We need to reset the counter when we succeed

internal/usbgadget/utils.go
```diff
@@ -112,67 +90,28 @@ func compareFileContent(oldContent []byte, newContent []byte, looserMatch bool)
...
-	l := logger.With().Int("counter", u.logSuppressionCounter[counterName]).Logger()
-
-	if u.logSuppressionCounter[counterName]%every == 0 {
-		if err != nil {
-			l.Error().Err(err).Msgf(msg, args...)
-		} else {
-			l.Error().Msgf(msg, args...)
+		if exceeded := u.logWithSuppression(context, "writeWithTimeout_"+fileName, 100, err, "write timed out"); exceeded {
+			// we've exceeded the suppression interval, so return an error so we can close and re-open the file
+			return 0, err
```
IDisposable, line 114: HUGE change

We need to eventually give up on trying to write, fall out of the goroutine and let the caller close and reopen the HID file.

internal/usbgadget/utils.go
```diff
@@ -112,67 +90,28 @@ func compareFileContent(oldContent []byte, newContent []byte, looserMatch bool)
...
-			"write timed out: %s",
-			file.Name(),
-		)
-		err = nil
-	}
-
-	return
-}
-
-func (u *UsbGadget) logWithSuppression(counterName string, every int, logger *zerolog.Logger, err error, msg string, args ...any) {
```
IDisposable, line 146: moved to usbgadget/log.go

internal/websecure/log.go
```diff
@@ -1,9 +0,0 @@
-package websecure
-
-import (
-	"os"
-
-	"github.com/rs/zerolog"
-)
-
-var defaultLogger = zerolog.New(os.Stdout).With().Str("component", "websecure").Logger()
```
IDisposable, line 9: No. we don't do runtime-startup loggers :)

internal/websecure/selfsign.go
```diff
@@ -59,34 +54,36 @@ func (s *SelfSigner) createSelfSignedCert(hostname string) *tls.Certificate {
...
 
 	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
-		s.log.Error().Err(err).Msg("Failed to generate private key")
+		logger.Error().Err(err).Msg("Failed to generate private key")
 		return nil
 	}
 
-	notBefore := time.Now()
+	notBefore := time.Now().AddDate(0, 0, -1) // ensure we don't have issues with clock skew
```
IDisposable, line 82: I've seen the SSL certificates fail if the client's clock is just a bit behind the device's clock (after correcting for timezones). This way we declare the certificate valid for bit into "the past" in case the client is still living there.

scripts/dev_deploy.sh
```diff
@@ -277,27 +279,9 @@ else
     SKIP_UI_BUILD=${SKIP_UI_BUILD_RELEASE} \
     ENABLE_SYNC_TRACE=${ENABLE_SYNC_TRACE}
 
-	# Kill any existing instances of the application
```
IDisposable, line 301: This all looks a bit like a merge error from a while back... there's duplication that should not be here.

scripts/dev_deploy.sh
```diff
@@ -320,7 +316,7 @@ cd "${REMOTE_PATH}"
 chmod +x jetkvm_app_debug
 
 # Run the application in the background
-PION_LOG_TRACE=${LOG_TRACE_SCOPES} ./jetkvm_app_debug | tee -a /tmp/jetkvm_app_debug.log
+./jetkvm_app_debug | tee -a /tmp/jetkvm_app_debug.log
```
IDisposable, line 319: This doesn't work to pass the desired environment variables...

scripts/dev_deploy.sh
```diff
@@ -312,6 +296,18 @@ while [ \$i -le 10 ]; do
...
 done
+EOF
+
+    # Copy the binary to the remote host
+    msg_info "▶ Copying the application to the remote host"
+    sshdev "cat > ${REMOTE_PATH}/jetkvm_app_debug" < bin/jetkvm_app
+
+    # Deploy and run the application on the remote host
+    msg_info "▶ Starting the application on the remote host"
+    logs=$(printenv | grep '^JETKVM_LOG' | sed "s/=/='/; s/$/'/")
```
IDisposable: We need to pass the _coder's_ env to the SSH command to pass on to the app.

ui/localization/messages/da.json
```diff
@@ -71,10 +71,22 @@
     "advanced_error_reset_config": "Konfigurationen kunne ikke nulstilles: {error}",
     "advanced_error_set_dev_channel": "Kunne ikke indstille udviklerkanaltilstand: {error}",
     "advanced_error_set_dev_mode": "Kunne ikke indstille udviklertilstand: {error}",
+    "advanced_error_set_log_level": "Kunne ikke indstille logniveau: {error}",
```
IDisposable, line 74: UI added to allow choosing the LOG level in Settings.

display.go
```diff
@@ -75,7 +76,7 @@ func updateDisplay() {
 		nativeInstance.UpdateLabelIfChanged("hdmi_status_label", "Disconnected")
 		_, _ = nativeInstance.UIObjClearState("hdmi_status_label", "LV_STATE_CHECKED")
 	}
-	nativeInstance.UpdateLabelIfChanged("cloud_status_label", fmt.Sprintf("%d active", actionSessions))
+	nativeInstance.UpdateLabelIfChanged("cloud_status_label", fmt.Sprintf("%d active", getActiveSessions()))
```
IDisposable: The active session count was getting corrupted while I was watching closely... so switched to atomic

hidrpc.go
```diff
@@ -7,99 +7,108 @@ import (
...
-			return
-		}
-		if err := session.HidChannel.Send(message); err != nil {
-			logger.Warn().Err(err).Msg("failed to send handshake message")
-			return
-		}
-		session.hidRPCAvailable = true
-	case hidrpc.TypeKeypressReport, hidrpc.TypeKeyboardReport:
-		rpcErr = handleHidRPCKeyboardInput(message)
+		return handleHidRPCHandshake(session)
```
IDisposable, line 18: Moved all the handlers to distinct functions as this hurt my brain.

hidrpc.go
```diff
@@ -7,99 +7,108 @@ import (
...
+		logger = logger.Str("message", message.String())
 	}
 
 	t := time.Now()
 
 	r := make(chan interface{})
 	go func() {
-		handleHidRPCMessage(message, session)
-		r <- nil
+		r <- handleHidRPCMessage(message, session)
```
IDisposable, line 104: HUGE change

We pass back the result from the handleHidRPCMessage (an error) so it can be logged.

hidrpc.go
```diff
@@ -7,99 +7,108 @@ import (
...
 	}()
 	select {
 	case <-time.After(1 * time.Second):
-		scopedLogger.Warn().Msg("HID RPC message timed out")
-	case <-r:
-		scopedLogger.Debug().Dur("duration", time.Since(t)).Msg("HID RPC message handled")
+		logger.Warn().Msg("HID RPC message took too long")
```
IDisposable, line 108: Changed the wording here... we don't actually abandon the processing of the message... we just complain that it was long-running.

hidrpc.go
```diff
@@ -147,14 +153,11 @@ func handleHidRPCKeypressKeepAlive(session *Session) error {
 				// This is likely a retransmit stall or ordering delay.
 				// We reject the tick entirely and DO NOT extend,
 				// so the auto-release still fires on time.
-				validTick = false
+				return nil
```
IDisposable, line 158: Simplify

hidrpc.go
```diff
@@ -216,23 +229,24 @@ func reportHidRPC(params any, session *Session) {
...
 		return
 	}
 
-	if err := session.HidChannel.Send(message); err != nil {
-		if errors.Is(err, io.ErrClosedPipe) {
-			logger.Debug().Err(err).Msg("HID RPC channel closed, skipping reportHidRPC")
-			return
+	// fire and forget...
+	logger.Debug().Msg("sending HID RPC report")
+	go func() {
```
IDisposable, line 239: HUGE HUGE change

We need to send the report messages back asynchronously. Shove the response out, but don't stop processing at the device end.

log.go
```diff
@@ -1,34 +0,0 @@
...
-	"github.com/jetkvm/kvm/internal/logging"
-	"github.com/rs/zerolog"
-)
-
-func ErrorfL(l *zerolog.Logger, format string, err error, args ...any) error {
-	return logging.ErrorfL(l, format, err, args...)
-}
-
-var (
-	logger          = logging.GetSubsystemLogger("jetkvm")
```
IDisposable, line 13: NO NO NO. Bad Marc... this broke all ability to change logging levels without a restart :(

main.go
```diff
@@ -92,62 +93,24 @@ func Main() {
...
-		logger.Warn().Err(err).Msg("failed to init images folder")
+		mainLogger.Warn().Err(err).Msg("failed to init images folder")
 	}
+
 	initJiggler()
 
 	// start video sleep mode timer
 	startVideoSleepModeTicker()
 
-	go func() {
```
IDisposable, line 129: moved this code out to a distinct method, and also now start it up later.

main.go
```diff
@@ -159,26 +122,25 @@ func Main() {
...
 	go RunWebsocketClient()
-	initPublicIPState()
 
+	initPublicIPState()
 	initSerialPort()
 
 	setProcTitle("ready")
+	mainLogger.Log().Msg("JetKVM ready")
+
+	go RunAutoUpdateCheck()
```
IDisposable, line 132: ... no don't start up the AUP until we're fully ready...

main.go
```diff
@@ -159,26 +122,25 @@ func Main() {
...
-	//if fuseServer != nil {
-	//	err := setMassStorageImage(" ")
-	//	if err != nil {
-	//		logger.Infof("Failed to unmount mass storage image: %v", err)
-	//	}
-	//	err = fuseServer.Unmount()
-	//	if err != nil {
-	//		logger.Infof("Failed to unmount fuse: %v", err)
-	//	}
+	if err := rpcUnmountImage(); err != nil {
```
IDisposable, line 138: Have to unmount the drive or the UsbGadget gets wonky... and it takes _reboot_ to get it back.

main.go
```diff
@@ -159,26 +122,25 @@ func Main() {
...
+		mainLogger.Warn().Err(err).Msg("failed to eject virtual media on shutdown")
+	}
+
+	gadget.Close()
```
IDisposable, line 142: Explicitly close the gadget so we don't have all the "setup" still pretending the devices exist to the client

ota.go
```diff
@@ -178,8 +191,53 @@ func rpcTryUpdateComponents(params updateParams, includePreRelease bool, resetCo
...
 		err := otaState.TryUpdate(context.Background(), updateParams)
 		if err != nil {
-			otaLogger.Warn().Err(err).Msg("failed to try update")
+			logging.GetSubsystemLogger("ota").Warn().Err(err).Msg("failed to try update")
 		}
 	}()
 	return nil
 }
+
+func RunAutoUpdateCheck() {
```
IDisposable, line 195: moved here from the main

usb.go
```diff
@@ -44,10 +37,21 @@ func initUsbGadget() {
 		}
 	})
 
-	// open the keyboard hid file to listen for keyboard events
-	if err := gadget.OpenKeyboardHidFile(); err != nil {
-		usbLogger.Error().Err(err).Msg("failed to open keyboard hid file")
-	}
+	go func() {
```
IDisposable, line 40: ... and the MAIN REASON we're here...

We need to do this last and we need to do two things... one, return the gadget to the main so it can `.Close()` it... and two.... automatically _reopen the keyboard HID file_ (which restarts the listeners) whenever the USB goes "ready"... so effectively this loop is where we notice the plug-in and "resume/start" the handling correctly.

NOTE: the open is basically a NOP if it was already open... and this loop is what _reopens_ it if we close it on some error... kicking things back off... so we can when too many writes timeout, or when too many reads (for the LED state) just close the whole gadget down and open 'er back up to get past them unplug/replug of the USB

usb_mass_storage.go
```diff
@@ -286,22 +302,28 @@ func rpcMountWithHTTP(url string, mode VirtualMediaMode) error {
 		URL:    url,
 		Size:   n,
 	}
-	virtualMediaStateMutex.Unlock()
```
IDisposable: dangled lock if we hit any errors in this method... ugh.

usb_mass_storage.go
```diff
@@ -350,7 +372,7 @@ func rpcGetStorageSpace() (*StorageSpace, error) {
 	var stat syscall.Statfs_t
 	err := syscall.Statfs(imagesFolder, &stat)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get storage stats: %v", err)
+		return nil, fmt.Errorf("failed to get storage stats: %w", err)
```
IDisposable, line 375: Errorf likes %w to know it should unwrap the error for more detail  (but NOT `Msgf`)

webrtc.go
```diff
@@ -39,39 +40,25 @@ type Session struct {
...
-)
-
-func incrActiveSessions() int {
-	activeSessionsMutex.Lock()
-	defer activeSessionsMutex.Unlock()
+var activeSessions atomic.Int32
 
-	actionSessions++
-	return actionSessions
+func incrActiveSessions() int32 {
```
IDisposable, line 47: As noted... was seeing the session count get wonky... so just went atomic

adamshiervani:
Can you factor these out to their own PR: 
> - Revamped the polling and writing of the USB gadget to correct for dropped connections
> - Corrected the initialization order to ensure we run even if the USB is disconnected (and we're getting power from the y-splitter/POE)
> - Ensured that we restart the LED listening polling
> -  If we exceed too many consecutive write errors, close and reopen the gadget

IDisposable:
> * Revamped the polling and writing of the USB gadget to correct for dropped connections
> * Corrected the initialization order to ensure we run even if the USB is disconnected (and we're getting power from the y-splitter/POE)
> * Ensured that we restart the LED listening polling
> * If we exceed too many consecutive write errors, close and reopen the gadget

It's going to be a bit... the logging changes were kind of invasive and required to actually find the bugs. I will extract the "functional" fixed when I get a bit more time.

internal/usbgadget/hid_keyboard.go
```diff
@@ -160,183 +186,169 @@ func (u *UsbGadget) SetOnKeepAliveReset(f func()) {
...
+	defer keyboardHidFileLock.Unlock()
+
+	if err := u.openKeyboardHidFileUnderMutex(); err != nil {
 		return err
 	}
 
 	_, err := u.writeWithTimeout(u.keyboardHidFile, append([]byte{modifier, 0x00}, keys[:hidKeyBufferSize]...))
 	if err != nil {
-		u.logWithSuppression("keyboardWriteHidFile", 100, u.log, err, "failed to write to hidg0")
 		u.keyboardHidFile.Close()
```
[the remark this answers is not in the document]
IDisposable: This is an error recovery situation. There's nothing we can do short of close/reopen.

IDisposable:
Going to revisit this after PR #1062 is merged as this was the origin of many of those changes, but also has all the log-cleanup. I'm switching this to Draft until I've crafted the "log cleanup" PR and then I'll close this one.

PR #1048, opened 2025-12-08
fix: handle versioned cloud URLs in reboot flow

adamshiervani:


ui/src/components/VideoOverlay.tsx
```diff
@@ -398,124 +399,126 @@ export function PointerLockBar({ show }: PointerLockBarProps) {
...
-    return () => {
-      clearTimeout(timeoutId);
-    };
+    const id = setTimeout(() => setHasTimedOut(true), 30_000);
+    return () => clearTimeout(id);
   }, [show]);
 
+  // Redirect helper - navigates and forces reload
+  const redirectTo = useCallback(async (url: string) => {
+    window.location.href = url;
```
IDisposable, line 435: Glad you extracted this pattern out to a utils.ts function... I think I've see it before ;)

ui/src/components/VideoOverlay.tsx
```diff
@@ -398,124 +399,126 @@ export function PointerLockBar({ show }: PointerLockBarProps) {
...
-      abortControllerRef.current = abortController;
-      isFetchingRef.current = true;
+      abortControllerRef.current?.abort();
+      const controller = new AbortController();
+      abortControllerRef.current = controller;
+      isCheckingRef.current = true;
 
-      console.log('Checking post-reboot health endpoint:', postRebootAction.healthCheck);
-      const timeoutId = window.setTimeout(() => abortController.abort(), 2000);
+      const timeout = setTimeout(() => controller.abort(), 2000);
```
IDisposable, line 452: Is 2 seconds enough?

ui/src/components/VideoOverlay.tsx
```diff
@@ -398,124 +399,126 @@ export function PointerLockBar({ show }: PointerLockBarProps) {
...
-      if (abortControllerRef.current) {
-        abortControllerRef.current.abort();
+      abortControllerRef.current?.abort();
+      isCheckingRef.current = false;
+    };
+  }, [show, postRebootAction, hasSeenDisconnect, redirectTo]);
+
+  // Cloud mode: wait for WebRTC reconnection via RPC, then redirect with versioned URL
+  useEffect(() => {
+    if (isOnDevice || !postRebootAction || !deviceId || !show || !hasSeenDisconnect) return;
```
IDisposable: I think this is muddled enough to warrant two statements:
```typescript
// cloud mode only
if (isOnDevice) return;

// ensure we actually should be waiting
if (!postRebootAction || !deviceId || !show || !hasSeenDisconnect) return;
```

IDisposable: ... or even the second line could be 
```typescript
if (false === (postRebootAction && deviceId && show && hasSeenDisconnect)) return;
```

ui/src/components/VideoOverlay.tsx
```diff
@@ -398,124 +399,126 @@ export function PointerLockBar({ show }: PointerLockBarProps) {
...
+    let cancelled = false;
+
+    const waitForReconnectAndRedirect = async () => {
+      if (isCheckingRef.current) return;
+      isCheckingRef.current = true;
+
+      try {
+        // getLocalVersion internally waits for RTC channel to be ready
+        const { appVersion } = await getLocalVersion();
+        if (cancelled) return;
```
IDisposable, line 509: _note to self_: do this second so that during the `await` if it gets cancelled...

ui/src/components/VideoOverlay.tsx
IDisposable, line 435: Should we stick in a console.log here?

ui/src/routes/devices.$id.tsx
```diff
@@ -699,15 +699,13 @@ export default function KvmIdRoute() {
...
-      console.debug("Setting reboot state", postRebootAction);
-
+      const action = resp.params as PostRebootAction | undefined;
       setRebootState({
         isRebooting: true,
         postRebootAction: {
-          healthCheck: postRebootAction?.healthCheck || `${window.location.origin}/device/status`,
-          redirectTo: postRebootAction?.redirectTo || window.location.href,
-        }
+          healthCheck: action?.healthCheck ?? "/device/status",
```
IDisposable: ... the this simplifies to 
```typescript
postRebootAction: action || defaultAction,
```

ui/src/routes/devices.$id.tsx
```diff
@@ -699,15 +699,13 @@ export default function KvmIdRoute() {
...
+          redirectTo: action?.redirectTo ?? "/",
```
IDisposable: ... cautionary note that if we get an `action` without one of the properties, this could be wrong... but I would contend that the handler that uses the `postRebootAction` should already be hardened against that...

ui/src/routes/devices.$id.tsx
```diff
@@ -823,7 +821,7 @@ export default function KvmIdRoute() {
 
     // Rebooting takes priority over connection status
     if (rebootState?.isRebooting) {
-      return <RebootingOverlay show={true} postRebootAction={rebootState.postRebootAction} />;
+      return <RebootingOverlay show={true} postRebootAction={rebootState.postRebootAction} deviceId={params.id} />;
```
IDisposable, line 824: _unrelated nit_ I would love to see the various "overlays" all have  close/dismiss in case they're wrong/intrusive to see what's going on in the video...

IDisposable (changes requested):

PR #1052, opened 2025-12-09
fix: correct Chinese translation errors in zh.json

ryiann:
This PR improves the Chinese translation in `zh.json`.


IDisposable (approved):
I'm not sure changing `URL` to `Address` is actually warranted/better... but it's not my language :) Looks fine.

PR #1064, opened 2025-12-12
Add sl_SI (Slovenian) keyboard layout

bzgec:


ui/src/keyboardLayouts/sl_SI.ts
```diff
@@ -0,0 +1,157 @@
...
+  "]": { key: "KeyG", AltGr: true },
+  "ł": { key: "KeyK", AltGr: true },
+  "Ł": { key: "KeyL", AltGr: true },
+  "ß": { key: "Quote", AltGr: true },
+  "¤": { key: "Backslash", AltGr: true },
+  "@": { key: "KeyV", AltGr: true },
+  "{": { key: "KeyB", AltGr: true },
+  "}": { key: "KeyN", AltGr: true },
+  "§": { key: "KeyM", AltGr: true },
+  // "<": { key: "Comma", AltGr: true },
```
IDisposable: Should these actually be commented-out?

bzgec: The symbols "<" and ">" can be typed in two different locations. I included them in the code for that reason. However, I commented them out because I received an error that said, "An object literal cannot have multiple properties with the same name." 
Should I remove them and update the pull request?

IDisposable: Please leave those in there, with a note, so I can take that situation into account as I do the keyboard layout work next on my schedule.

IDisposable: Please leave those in there, with a note, so I can take that situation into account as I do the keyboard layout work next on my schedule.

IDisposable: Please leave those in there, with a note, so I can take that situation into account as I do the keyboard layout work next on my schedule.

IDisposable (approved):

PR #1093, opened 2025-12-23
Add option to enforce TLS by redirecting HTTP requests to HTTPS (#1092)

### LINKED ISSUE #1092: Enforce TLS when HTTPS Mode is enabled
### A note for the community

> [!NOTE]

### Disclaimer

### Subsystem

{"UI" => "Application"}

### Feature description

Currently, when HTTPS mode is enabled, the server still responds normally via HTTP.

It is preferred that there is an option to redirect all HTTP requests to HTTPS to prevent any credential or token leakage.

This would help prevent any MITM attacks, both via sniffing traffic or downgrading HTTPS to HTTP.

Currently, the initial login sends the login password in plaintext in a POST request. This is easily obtainable without the use of TLS.

Combined with HSTS headers, this would force browsers to always connect to the JetKVM securely when a user accesses the host via plain HTTP.

DigitalDJ:
Closes #1092 

### Summary

- An option to redirect all HTTP requests to HTTPS when HTTPS mode is enabled.
- When enabled, the HTTP router always responds with a redirect to HTTPS. 
- A HSTS header is included to force future connections to TLS.
- Exposes an "Enforce TLS" option in the Access options UI to enable
- The default is disabled.

### Checklist

- [ X ] Lints pass; CI green
- [ X ] Tricky parts are commented in code


ui/src/routes/devices.$id.settings.access._index.tsx
```diff
@@ -187,9 +193,17 @@ export default function SettingsAccessIndexRoute() {
     setTlsKey(value);
   };
 
+  const handleTlsEnforceChange = (value: boolean) => {
+    setTlsEnforce(value);
+    // For "disabled" and "self-signed" modes, immediately apply the settings
+    if (tlsMode !== "custom") {
```
IDisposable: Would it be better to only immediately apply setting to `"disabled"`.

Also we should **not** apply the force option being turned on until an explicit Save click (like we do for many of the Network settings), so that we don't immediately lock someone out of their device because of an errant click?

ui/src/routes/devices.$id.settings.access._index.tsx
```diff
@@ -148,12 +152,13 @@ export default function SettingsAccessIndexRoute() {
...
   // Function to update TLS state - accepts a mode parameter
   const updateTlsState = useCallback(
-    (mode: string, cert?: string, key?: string) => {
+    (mode: string, enforce: boolean, cert?: string, key?: string) => {
       const state = { mode } as TLSState;
       if (cert && key) {
         state.certificate = cert;
         state.privateKey = key;
       }
+      state.enforce = enforce;
```
IDisposable: Shouldn't be able to set enforce to `true` if the `mode` is `"disabled"`, so we need to make sure that is enforced here.

web.go
```diff
@@ -83,6 +84,22 @@ func setupRouter() *gin.Engine {
 		}),
 	))
 
+	if secureRedirect {
+		r.Use(secure.Secure(secure.Options{
+			AllowedHosts:          []string{},
+			SSLRedirect:           true,
+			SSLHost:               "",
+			SSLProxyHeaders:       map[string]string{"X-Forwarded-Proto": "https"},
+			STSSeconds:            315360000,
```
IDisposable: We should *not* be setting STS requirements. It can be normal best-practice for very strongly secured websites, but if you set STS on a domain, the browser is going to cache that in an almost unclearable way (it's a **real PITA** to undo) and might make getting back to the device very difficult if there's some cert setup failure.

web.go
```diff
@@ -564,8 +581,12 @@ func basicAuthProtectedMiddleware(requireDeveloperMode bool) gin.HandlerFunc {
 	}
 }
 
+var (
+	updateWebRouter = make(chan struct{})
+)
+
 func RunWebServer() {
-	r := setupRouter()
+	r := setupRouter(config.TLSMode != "" && config.TLSEnforce)
```
IDisposable: If we ensure that `config.TLSEnforce` can never be `true` unless the setup is completely valid (in _web_tls.go_'s `getTLSState()`, then we only need to test the flag. Given the potential lockout nature of this setting, we should be **really** defensive in the setting of TLS requirements.

web.go
```diff
@@ -564,8 +581,12 @@ func basicAuthProtectedMiddleware(requireDeveloperMode bool) gin.HandlerFunc {
 	}
 }
 
+var (
+	updateWebRouter = make(chan struct{})
```
IDisposable, line 585: We need to close this `chan` at web server shutdown.

web_tls.go
```diff
@@ -68,6 +69,9 @@ func getCertificate(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
 
 func getTLSState() TLSState {
 	s := TLSState{}
+
+	s.Enforce = config.TLSEnforce
```
IDisposable: Move this down below the mode and cert setup and make it _also_ matter that the the config is valid.

Something like 
```go
canEnforce := false

switch config.TLSMode {
	case "disabled":
		s.Mode = "disabled"
	case "custom":
		s.Mode = "custom"
		cert := certStore.GetCertificate(webSecureCustomCertificateName)
		if cert != nil {
			var certPEM []byte
			// convert to pem format
			for _, c := range cert.Certificate {
				block := pem.Block{
					Type:  "CERTIFICATE",
					Bytes: c,
				}

				certPEM = append(certPEM, pem.EncodeToMemory(&block)...)
			}
			s.Certificate = string(certPEM)
			canEnforce = true
		}
	case "self-signed":
		s.Mode = "self-signed"
        canEnforce = true
	}

s.Enforce = canEnforce && config.TLSEnforce
```

web.go
```diff
@@ -73,7 +74,7 @@ var cachableFileExtensions = []string{
 	".jpg", ".jpeg", ".png", ".svg", ".gif", ".webp", ".ico", ".woff2",
 }
 
-func setupRouter() *gin.Engine {
+func setupRouter(secureRedirect bool) *gin.Engine {
```
IDisposable: Let's not pass this in... should just test `config.TLSEnforce` at [line 87](https://github.com/jetkvm/kvm/pull/1093/changes#diff-a0a6130738465fc41d08d7b30558d8f1e63d7ac34466343db619b1d020487225R87)

web_tls.go
```diff
@@ -159,7 +172,7 @@ func runWebSecureServer() {
 		tlsStarted = false
 	}()
 
-	r := setupRouter()
+	r := setupRouter(false)
```
IDisposable: This will start up the router without redirection on... if nobody changes the setting, what will cause it to be enabled?

See [this comment](https://github.com/jetkvm/kvm/pull/1093/changes#r2644305908)

IDisposable (changes requested):
This looks really great! Thanks!

Just a couple concerns about defensively protecting against accidental lockout.

web.go
```diff
@@ -592,7 +613,19 @@ func RunWebServer() {
...
 	logger.Info().Str("bindAddress", bindAddress).Bool("loopbackOnly", config.LocalLoopbackOnly).Msg("Starting web server")
-	if err := r.Run(bindAddress); err != nil {
+	server := &http.Server{
+		Addr:    bindAddress,
+		Handler: r,
+	}
+
+	go func() {
+		for range updateWebRouter {
+			server.Handler = setupRouter(config.TLSMode != "" && config.TLSEnforce)
```
IDisposable: I wonder if we should make the handler swap delayed slightly (like a few seconds) when turning _on_ enforcement (and do NOT delay in turning off) so that the in-flight WebRTC setting changes have a chance to settle before ripping the rug out from under them...

IDisposable: We should _not_ be passing the config into `setupRouter()`. See [this comment](https://github.com/jetkvm/kvm/pull/1093/changes#diff-a0a6130738465fc41d08d7b30558d8f1e63d7ac34466343db619b1d020487225R77)

ui/src/routes/devices.$id.settings.access._index.tsx
DigitalDJ: I have made it so that Enforce TLS can only be enabled after acknowledging a pop-up and the user applies TLS settings (in "custom" / "self-signed") mode.

ui/src/routes/devices.$id.settings.access._index.tsx
DigitalDJ: Resolved by checking "disabled" and forcing false if so.

web.go
DigitalDJ: Makes me a bit sad... but understand the user experience side. I've made it so HSTS is disabled.

web.go
DigitalDJ: I've reworked setupRouter to check the config in the function itself, instead of passing it down. setupRouter still needs to know which server is calling it though (the HTTP or HTTPS server). Since the HTTPS server should never enable the redirect. I've tried to make the argument more clear.

web.go
DigitalDJ: I've added a call to close() after the web server returns. I'm not quite sure if this is right though. To be honest, I don't think this matters too much? Since the HTTP server is always running?  I think the correct place to close() (if you really wanted to) was before the entire process was about to exit?

However, this did uncover a little bug with regards to the startTLS/stopTLS channels. stopTLS channel would block the TLS server from fully shutting down. I've resolved this in a commit part of this pull.

With regards to startTLS channel, same as updateWebRouter, I don't think it makes sense to close unless the process is exiting?

The process exit is pretty forceful at the moment, so I haven't gone that deep.

web_tls.go
DigitalDJ: I've implemented basically what you have here. Only enforce TLS if TLS gets enabled successfully.

Also small bug here -- config.TLSMode is never "disabled". It uses empty string "" to indicate disabled....so the case statement is confusing the JSON RPC state vs. config schema. I've fixed this to default to TLS disabled.

web.go
DigitalDJ: I implemented a 3 second sleep for the enforce TLS swap over.

DigitalDJ:
I've tried to address all the issues here. If there are any more please let me know. I've tried to focus on enforce TLS not being enabled unless everything is working okay.... and making the user jump through some hoops to make it happen.

DigitalDJ:
I should note explicitly there are two other semi-related problems fixed here. That need attention (maybe). I don't really know Go (or TS for that matter).

1. [[7fbd475](https://github.com/jetkvm/kvm/pull/1093/commits/7fbd475d2ff505ed288fc4d4a9f2f551008e387a)] Fixes the TLS server not fully shutting down when going from disabled -> enabled -> disabled. The `stopTLS` channel seems to block `server.ListenAndServeTLS` from returning when `server.Shutdown` is called? Not sure about the inner-workings of this, but closing `stopTLS` when we stop the server and remaking it for next start resolves the issue.
2. `getTLSState` confuses the JSON RPC schema with the config schema. It compares `config.TLSMode` to "disabled" to check the state of TLS, but the config schema uses an empty string to indicate TLS disabled. I fixed this in [[2cef0ea](https://github.com/jetkvm/kvm/pull/1093/commits/2cef0eae09ed77e09a2607536de538972e57e98b)]

ui/localization/messages/en.json
```diff
@@ -19,6 +19,13 @@
     "access_description": "Manage the Access Control of the device",
     "access_disable_protection": "Disable Protection",
     "access_enable_password": "Enable Password",
+    "access_enforce_tls_label": "Enforce TLS",
```
IDisposable: Let's change this to `Require TLS` and adjust the other prompts.

ui/localization/messages/en.json
```diff
@@ -19,6 +19,13 @@
...
+    "access_enforce_tls_description": "Redirect all HTTP requests to HTTPS",
+    "access_enforce_tls_warning_before": "Before enabling this feature, make sure you have:",
+    "access_enforce_tls_warning_confirm": "I Understand, Enable Anyway.",
+    "access_enforce_tls_warning_description": "WARNING: This will restrict web interface access to HTTPS only.",
+    "access_enforce_tls_warning_https": "HTTPS mode is enabled and working.",
+    "access_enforce_tls_warning_title": "Enable Enforce TLS?",
```
IDisposable: Change to `Force TLS/HTTPS?` ??

ui/localization/messages/en.json
IDisposable: Perhaps `"Enforce TLS/HTTPS"`?

ui/localization/messages/en.json
IDisposable: `"Are You Sure You Want to Enforce TLS/HTTPS?"` might be clearer

config.go
```diff
@@ -188,6 +189,7 @@ func getDefaultConfig() Config {
 		// This is the "Standard" jiggler option in the UI
 		JigglerConfig: func() *JigglerConfig { c := defaultJigglerConfig; return &c }(),
 		TLSMode:       "",
+		TLSEnforce:    false,
```
IDisposable, line 192: Outside the scope of this PR, but I can see this eventually morphing into what TLS version is required... e.g. TLS 1.2

web.go
```diff
@@ -83,6 +84,22 @@ func setupRouter() *gin.Engine {
...
 
+	if !isSecureServer && config.TLSEnforce {
+		r.Use(secure.Secure(secure.Options{
+			AllowedHosts:          []string{},
+			SSLRedirect:           true,
+			SSLHost:               "",
+			SSLProxyHeaders:       map[string]string{"X-Forwarded-Proto": "https"},
+			STSSeconds:            0,
+			STSIncludeSubdomains:  false,
+			FrameDeny:             true,
```
IDisposable: I don't know of anyone currently `IFRAME`ing the current device control UI... but this would disable that (intentionally)... might consider a white-list instead?

web_tls.go
```diff
@@ -86,23 +89,29 @@ func getTLSState() TLSState {
...
 
 	switch s.Mode {
 	case "disabled":
 		if config.TLSMode != "" {
 			isChanged = true
 		}
 		config.TLSMode = ""
+		config.TLSEnforce = false
 	case "custom":
 		if config.TLSMode == "" {
```
IDisposable, line 115: This isn't something you actually changed, but looks like this line should really be `if config.TLSMode != "custom" {` instead... as the current code runs, if someone goes from custom to self-signed, it's not noted as `isChanged = true` which means that we don't restart the secure server... so would it actually notice the certificate changing? I suspect this line and line 131 should both be checking that we didn't change modes.

web_tls.go
```diff
@@ -117,15 +126,22 @@ func setTLSState(s TLSState) error {
...
 		}
 		config.TLSMode = "self-signed"
+		config.TLSEnforce = s.Enforce
 	default:
 		return fmt.Errorf("invalid TLS mode: %s", s.Mode)
 	}
 
+	if oldEnforce != config.TLSEnforce {
+		logger.Info().Msg("Rerouting web server, as TLS enforcement changed")
+		updateWebRouter <- struct{}{}
```
IDisposable: Should we do this _after_ we have started the secure web server down there around line 157 so we don't start redirects until the HTTPS will be served?

web_tls.go
DigitalDJ: I'm pretty sure the change does affect the certificate without restarting the secure web server, but I'd have to verify. Variable should probably read something like "restartHttps" rather than "isChanged"

ui/localization/messages/en.json
DigitalDJ: In this case, I am copying the pattern used by "Enable Loopback-Only Mode". This string is used in the warning window dialog with the confirm being "I Understand, Enable Anyway". So just trying to keep it consistent with that.

web_tls.go
DigitalDJ: I've verified. The reason for this is because you can change the certificate without restarting the entire Web Secure server, so it would be an unnecessary restart when going to and from "custom" <--> "self-signed".

web_tls.go
DigitalDJ: I've moved this below the web secure signalling, and made it more clear what isChanged signifies here.

web.go
DigitalDJ: I've set this back to false for now. Maybe an additional "harden" option could blanket enable additional security features like this and HSTS.

ui/localization/messages/en.json
DigitalDJ: I've updated (and made consistent) the UI to say "Enforce HTTPS (TLS)"...making it consistent with the UI's general use of "HTTPS Mode", but signaling to the user that it is indeed TLS.

IDisposable (approved):
Looks good to me :shipit:

adamshiervani (changes requested):
"Enforce HTTPS (TLS)" doesn't work for me. It gets checked in the UI, but after a refresh of the page, it's unchecked, and I can access it through HTTP.

Could you please also add a E2E test for this? AI can be very helpful writing the E2E tests.

DigitalDJ:
I just pushed a change that should fix what you're seeing in "self-signed" mode. The button should now appear, to apply TLS settings, when "self-signed" mode is used. The intent is for the user to agree to the popup AND click the apply button to enforce TLS.

As for E2E tests, I'll probably pass on that sorry. If someone wants to take that up, they are more than welcome. If that means this sits here, that's fine. I'll maintain my own fork rebasing against release versions for those that want the feature, for as long as I own a JetKVM.

Unfortunately, the complete lack of response from JetKVM's support channels has made me re-think the investment here. Sorry about that.

PR #1095, opened 2025-12-23
Prevent mDNS queries by ICE when mDNS is disabled in config (#1094)

### LINKED ISSUE #1094: Disabling mDNS does not fully disable mDNS queries
### Disclaimer

### Application version

0.5.1

### System version

0.2.5

### Device model

JetKVM

### Extension model

None

### Remote device Hardware

_No response_

### Remote device OS

_No response_

### Bug description

When mDNS is set to disabled and a client is connected to JetKVM, mDNS queries are still regularly performed (4/second approximately)

This is due to pion/ice library defaulting to using mDNS for ICE.

DigitalDJ:
Fixes #1094

### Summary
- Pass whether mDNS is enabled down to the WebRTC session
- If it is disable set ICE Multicast DNS Mode to disabled
- Prevents mDNS query spam when a client is connected to JetKVM and mDNS is set to disabled

### Checklist
- [ X ] Lints pass; CI green

### DIFF
--- a/cloud.go
+++ b/cloud.go
@@ -454,7 +454,7 @@ func handleSessionRequest(
 		LocalIP:    req.IP,
 		ICEServers: req.ICEServers,
 		Logger:     scopedLogger,
-		NoMDNS:     config.NetworkConfig.MDNSMode.String == "disabled",
+		MDNSMode:   config.NetworkConfig.MDNSMode.String,
 	})
 	if err != nil {
 		_ = wsjson.Write(context.Background(), c, gin.H{"error": err})
--- a/cloud.go
+++ b/cloud.go
@@ -454,6 +454,7 @@ func handleSessionRequest(
 		LocalIP:    req.IP,
 		ICEServers: req.ICEServers,
 		Logger:     scopedLogger,
+		NoMDNS:     config.NetworkConfig.MDNSMode.String == "disabled",
 	})
 	if err != nil {
 		_ = wsjson.Write(context.Background(), c, gin.H{"error": err})
--- a/web.go
+++ b/web.go
@@ -212,7 +212,7 @@ func handleWebRTCSession(c *gin.Context) {
 		return
 	}
 
-	session, err := newSession(SessionConfig{})
+	session, err := newSession(SessionConfig{NoMDNS: config.NetworkConfig.MDNSMode.String == "disabled"})
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, gin.H{"error": err})
 		return
--- a/web.go
+++ b/web.go
@@ -212,7 +212,7 @@ func handleWebRTCSession(c *gin.Context) {
 		return
 	}
 
-	session, err := newSession(SessionConfig{NoMDNS: config.NetworkConfig.MDNSMode.String == "disabled"})
+	session, err := newSession(SessionConfig{MDNSMode: config.NetworkConfig.MDNSMode.String})
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, gin.H{"error": err})
 		return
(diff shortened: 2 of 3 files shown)


cloud.go
```diff
@@ -454,6 +454,7 @@ func handleSessionRequest(
 		LocalIP:    req.IP,
 		ICEServers: req.ICEServers,
 		Logger:     scopedLogger,
+		NoMDNS:     config.NetworkConfig.MDNSMode.String == "disabled",
```
IDisposable: s/`NoMDNS`/`MDNSMode` so we have the IPV4/IPV6/none setting.

web.go
```diff
@@ -212,7 +212,7 @@ func handleWebRTCSession(c *gin.Context) {
 		return
 	}
 
-	session, err := newSession(SessionConfig{})
+	session, err := newSession(SessionConfig{NoMDNS: config.NetworkConfig.MDNSMode.String == "disabled"})
```
IDisposable: We should pass the entire `config.NetworkConfig.MDNSMode` string into the SessionConfig so it can _also_ drive a call to `SetNetworkTypes`.

webrtc.go
```diff
@@ -123,6 +124,7 @@ type SessionConfig struct {
 	IsCloud    bool
 	ws         *websocket.Conn
 	Logger     *zerolog.Logger
+	NoMDNS     bool
```
IDisposable: `MDNSMode string`

webrtc.go
```diff
@@ -249,6 +251,11 @@ func newSession(config SessionConfig) (*Session, error) {
 	webrtcSettingEngine := webrtc.SettingEngine{
 		LoggerFactory: logging.GetPionDefaultLoggerFactory(),
 	}
+
```
IDisposable, line 254: Here we should (as you did!) call `webrtcSettingEngine.SetICEMulticastDNSMode` to enable/disable
We _also_ should call `webrtcSettingEngine.SetNetworkTypes()` to reflect the MDNSMode setting 

At this time the pion webrtc only supports UDP, so we can just use `NetworkTypeUDP4` and `NetworkTypeUDP6` for the options.
 
Since `MDNSMode` can be _disabled,auto,ipv4_only,ipv6_only_, then we just end up with something like:

```go
mdnsNetworkTypes := make([]NetworkType, 0)
if mdnsMode == "auto" || mdnsMode == "ipv4_only" {
  mdnsNetworkTypes = append(mdnsNetworkTypes, ice.NetworkTypeUDP4)
}
if mdnsMode == "auto" || mdnsMode == "ipv6_only" {
  mdnsNetworkTypes = append(mdnsNetworkTypes, ice.NetworkTypeUDP6)
}

if len(mdnsNetworkTypes) > 0 {
  webrtcSettingEngine.SetNetworkTypes(mdnsModes)
  webrtcSettingEngine.SetICEMulticastDNSMode(ice.MulticastDNSModeQueryOnly)
} else {
  webrtcSettingEngine.SetICEMulticastDNSMode(ice.MulticastDNSModeDisabled)
}
```

When deciding if we should be using `MulticastDNSModeQueryOnly` or `MulticastDNSModeQueryAndGather` [*](https://github.com/pion/ice/blob/master/mdns.go#L20), it looks like in _pion/webrtc/agent.go_ it [defaults if not set to  `MulticastDNSModeQueryOnly`](https://github.com/pion/ice/blob/master/agent.go#L430)... or we could set it to `0` (the uninitialized state) and let the defaulting happen as now... but that seems less clear.

(Warning: code above was written in the GitHub comment feature and likely is syntactically wrong)

IDisposable (changes requested):
Please also call `webrtcSettingEngine.SetNetworkTypes()` to reflect the `MDNSMode` setting (disabled, ipv4, ipv6, auto) where auto means both...

DigitalDJ:
I've updated to include your suggested changes. Thanks!

IDisposable (approved):

Issue #645, opened 2025-06-25
Support offline JetKVM devices


hertg:
> [!NOTE]
> I am creating this as a meta issue, to track the progress of JetKVM becoming usable and user-friendly in offline settings. Please don't hesitate  to close this, if you (the maintainers) don't like having meta issues in your repository.

Some people may prefer their KVM devices to have limited or no access to the internet. From a security perspective, with the level of access the KVM device may grant to internal infrastructure, it makes sense to a) not connect it to a cloud service and b) even completely isolate the device from the outside world.

Right now, JetKVM is _usable_ without WAN access, but there's still things that need to be addressed, I am listing them below.

- [ ] #516 and #534 
  Timesync does not work because JetKVM uses static public NTP servers; which also leads to HTTPS-mode not working because the self-signed certificates are issued for timespans in the past
- [ ] #96 
  It is currently **very** complex to do app and firmware updates offline
- [x] #368
  Not necessarily about WAN access, but it should be possible to set a static IP in environments where there is no, or an unstable DHCP server

Maybe there's something I missed, but those are the things I stumbled over while configuring my offline JetKVM devices, please feel free to add to he list.

pmhausen:
I switched to the beta release to finally be able to configure a proper NTP server. Thank you for the work that went into this feature.

But ... why doesn't JetKVM "just" use the NTP servers it gets via DHCP? Any reasonable managed enterprise network has that. It even displays the server(s) it received in the UI but does not use them! That's why I took for granted it would just work like any other device in my network ... e.g. for Mikrotik gear, you tick "NTP client - enable" and that's that. If it gets one via DHCP it will use it.

It came as quite a surprise that JetKVM displays the server but does not use it. I block access to external NTP for all devices but my own NTP servers.

Is there any chance that could be implemented, too?

<img width="631" height="519" alt="Image" src="https://github.com/user-attachments/assets/1326970c-0866-4c20-bbbc-d237c4b3842c" />

Kind regards,
Patrick

fernvenue:
I really have no idea what JetKVM is doing with their product design, from not supporting static IPs early on, to now not allowing settings to be modified without an internet connection, why the fuck do you think I need to change the settings in the first place?! Because I need to debug and configure the network, obviously! If you're only on a local network, even the settings are grayed out and inaccessible, why on earth is it like this? WTF??! What the hell are you guys thinking? This is the single most regretted purchase I've ever made, bar none. It's bullshit on design.

PR #1102, opened 2025-12-26
Custom NTP Configuration UI

### LINKED ISSUE #516: Add option to configure NTP server(s) used by JetKVM device
Some networks have local NTP servers which may be preferred or mandatory. There is also a DHCP option for NTP servers which the device could use, although support for this in general isn't widespread.

DigitalDJ:
Closes #516, #645, #59

### Summary

- There's no web user interface for the custom NTP settings that are implemented in the back-end Go application
- This is a first-cut implementation of a "Custom" Time Synchronization option
- The options are mapped directly to the back-end configuration options
- An effort has been made to validate all fields. i.e. NTP servers must be specified when the ordering includes user-specified NTP servers

### Checklist

- [ X ] Lints pass; CI green
- [ X ] Tricky parts are commented in code


adamshiervani (changes requested):
Thanks for implementing the time sync configuration and also all your other PRs!! 

The functionality is solid, but I think the UI needs a remodel. The current design reflects our internal data model rather than what users actually need.

Usually there are two groups with a 95/5 ratio:

1. **"Just work"** - Automatic time sync, don't care how
2. **"Use my servers"** - Enterprise/air-gapped environments with specific NTP servers

### Proposed Simplification

**Two modes: Automatic | Custom**

```
Time Synchronization
└─ [Automatic ▼]
```

When "Custom":

```
┌────────────────────────────────────────────────────────┐
│ Your NTP Servers                                       │
│ [ntp.internal.corp.com                     ] [×]       │
│ [+ Add Server]                                         │
│                                                        │
│ Your HTTP URLs (optional)                              │
│ [http://time.internal.corp.com             ] [×]       │
│ [+ Add URL]                                            │
│                                                        │
│ ☑ Fall back to built-in servers if yours fail         │
│ ☐ Also try NTP servers from DHCP (for isolated networks)│
└────────────────────────────────────────────────────────┘
```

### Backend Mapping

| UI State | Backend Values |
|----------|----------------|
| Automatic | `time_sync_mode: "ntp_and_http"`, default ordering |
| Custom + fallback | `mode: "custom"`, `ordering: [user_ntp, user_http, ntp, http]`, `disable_fallback: false` |
| Custom + fallback + DHCP | Same but ordering includes `ntp_dhcp` after user sources |
| Custom, no fallback | `ordering: [user_ntp, user_http]`, `disable_fallback: true` |

DigitalDJ:
Hi there! I agree with your comments, and that's what I thought of doing. I implemented this as a first-pass.... just to basically understand what the model was behind the scenes. At least this gives power-users access to the advanced features.

From here, yeah, I agree. Things like "Disable Fallback" shouldn't be possible with no specified NTP/HTTP URLs.... A lot of the things should be "automatic" unless specified. Then even the "specified" things should only be the servers / URLs. As long as things behind the scenes only hit the servers specified, I think offline users / air-gaped users would be satisfied.

I might get around to simplifying this over the next couple of weeks....but not being a React / TS dev...this makes it a bit of a struggle :P

IDisposable:
> Custom + fallback + DHCP	Same but ordering includes `ntp_dhcp` after user sources

As a network engineer, I would expect DHCP to be honored _first_

Torstein-Eide:
> As a network engineer, I would expect DHCP to be honored _first_

I disagree, and it depends on wording, for the [SNTP](https://en.wikipedia.org/wiki/Network_Time_Protocol#SNTP) client. (I don't think this i a full stack NTP client, since it doesn't need it.)
Similar to how `systemd-timesyncd.service`, handles it.  

> ### NTP=
> A space-separated list of NTP server host names or IP addresses. During runtime this list is combined with any per-interface NTP servers acquired from [systemd-networkd.service(8)](https://www.freedesktop.org/software/systemd/man/latest/systemd-networkd.service.html#). systemd-timesyncd will contact all configured system or per-interface servers in turn, until one responds. When the empty string is assigned, the list of NTP servers is reset, and all prior assignments will have no effect. This setting defaults to an empty list.


> ### FallbackNTP=
> A space-separated list of NTP server host names or IP addresses to be used as the fallback NTP servers. Any per-interface NTP servers obtained from [systemd-networkd.service(8)](https://www.freedesktop.org/software/systemd/man/latest/systemd-networkd.service.html#) take precedence over this setting, as do any servers set via NTP= above. This setting is hence only relevant if no other NTP server information is known. When the empty string is assigned, the list of NTP servers is reset, and all prior assignments will have no effect. If this option is not given, a compiled-in list of NTP servers is used.

Ref: [timesyncd.conf(5)](https://www.freedesktop.org/software/systemd/man/latest/timesyncd.conf.html)

PR #1128, opened 2026-01-03
feat: allows disabling IPv4 or IPv6

### LINKED ISSUE #1039: Restore setting to disable IPv6
### A note for the community

> [!NOTE]

### Disclaimer

- [ ] I plan to implement the feature myself.

### Subsystem

None

### Feature description

Same as issue https://github.com/jetkvm/kvm/issues/685 but that seems to never have been completed (or was undone at some point).  Basically, I'd like to be able to explicitly disable IPv6 support.

I am on app 0.5.0-dev202511271020, system 0.2.7, and the only available options for the **Network -> IPv6 Mode** setting are **SLAAC**, **Static**, and **Local-link only**.

An explicit **Disabled** option existed at some point, but it was removed in PR https://github.com/jetkvm/kvm/pull/878.  PR https://github.com/jetkvm/kvm/pull/864 (which was merged afterwards) added it back in code, but commented out.  It remains that way today: https://github.com/jetkvm/kvm/blob/15352a7796f5d795a93dcaa88bc9f095775ab5f9/ui/src/routes/devices.%24id.settings.network.tsx#L518

Please restore this functionality.

BPplays:
Closes #1039

### Summary

allows disabling IPv4 or IPv6

known issues:
- disabling IPv4 causes DNS to stop working because there is no RDNSS support
- you can disable IPv6 and IPv4 at the same time, this is probably never wanted
- the button to disable IPv4 currently uses the localization label to disable IPv6
- there is no warning if the user is disabling the IP version they're currently using to connect

i've mostly been testing disabling IPv4 but testing disabling IPv6 does seem to work too

### Checklist

- [ ] Lints pass; CI green
- [ ] Tricky parts are commented in code

### DIFF
--- a/internal/native/eez/src/ui/screens.h
+++ b/internal/native/eez/src/ui/screens.h
@@ -32,7 +32,6 @@ typedef struct _objects_t {
     lv_obj_t *cloud_status_icon;
     lv_obj_t *cloud_status_label;
     lv_obj_t *home_info_container;
-    lv_obj_t *ipinfo;
     lv_obj_t *home_info_ipv4_addr;
     lv_obj_t *home_info_ipv6_addr;
     lv_obj_t *home_info_mac_addr;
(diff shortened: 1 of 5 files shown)


IDisposable (approved):

86dd:
I think it would be important to add IPv6 RDNSS support before this is merged, as people will likely complain about broken DNS when disabling IPv4, and maybe add a warning or just refuse when the User disables both stacks, as this is well, an IP(v4/v6) KVM :grin:

PR #1137, opened 2026-01-07
Change production LOGLEVEL

adamshiervani:
### Summary
- Change default loglevel from INF to WARN in production as this could cause insanely large log files
- Migrate old configurations

### Checklist
- [x] Lints pass; CI green

### DIFF
--- a/config.go
+++ b/config.go
@@ -195,7 +195,7 @@ func getDefaultConfig() Config {
 			_ = confparser.SetDefaultsAndValidate(c)
 			return c
 		}(),
-		DefaultLogLevel:    "INFO",
+		DefaultLogLevel:    "WARN",
 		VideoQualityFactor: 1.0,
 	}
 }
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -272,6 +272,11 @@ func LoadConfig() {
 		loadedConfig.KeyboardLayout = "en-US"
 	}
 
+	// Migrate old verbose log level to sensible default
```
IDisposable, line 275: This will effectively prevent people from selecting `INFO` won't it? Seems like this will run every boot.

adamshiervani: Yeah, that's right. Luckily, the config is not exposed for the user yet, and I had to change the log level for all because users were getting insanely large logs files. One log I saw, was 100MB in 7 days.

Once we expose this configuration to people we can remove this line, as most users will be on 'WARN' in their configuration.

PR #1138, opened 2026-01-07
Enhance diagnostics logging with context-aware timeouts

adamshiervani:
### Summary
- Append diagnostics to log file when supervisor detects a crash

### Checklist

- [x] Lints pass; CI green


internal/diagnostics/helpers.go
```diff
@@ -12,7 +12,11 @@ const defaultCmdTimeout = 2 * time.Second
 
 // runCmdLog runs a command and logs its output.
 func (d *Diagnostics) runCmdLog(label string, cmd string, args ...string) {
-	ctx, cancel := context.WithTimeout(context.Background(), defaultCmdTimeout)
+	parentCtx := d.ctx
```
IDisposable, line 15: I usually use a coalescing helper (in a utils package) for this 
````go
func orBackground(ctx context.Context) context.Context {
	if ctx == nil {
		return context.Background()
	} else {
		return ctx
	}
}

PR #1161, opened 2026-01-27
Add missing languages and correct translations of languages.

IDisposable:
### Summary
- When languages are added to the localization options, we need to always add them to each language. This was not done when Japanese and Portuguese were added so we didn't have a self-localized display string for those in the English locale.
- When languages are added to the localization options, we need to add  **all other** languages in the language of the **new** localization. For all _locale_*.json_ files, this means the translation of a `"locale_*":` should be in the language OF the containing locale file. For example, in the _locale_pt.json_, we should have `"locale_en": "Inglês"`. This is different from the _locale_en.json_ which always shows the name in the native language of the locale being represented, e.g. `"locale_es": "Español",` is the Spanish way of saying spanish.

### Changes
- Add missing _locale_jp_ and _locale_pt_ to all the existing languages (were missed when those PRs were merged)
  - In the _en.json_ added them in their own native language
  - In the rest added them in the native language of the containing file
- Corrected the locale translations in _ja.json_ to be in Japanese (e.g. `"locale_en": "英語",` for English)
- Corrected the locale translations in _zh-tw.json_ to be in Traditional Chinese (e.g. `"locale_es": "西班牙語",` for Spanish)
- Correct the Chinese locale option translations in all the other _ui/localization/messages/_ files as they should be in the native language of the containing file (e.g, `"locale_zh": "Kinesisk (forenklet)",` for the Danish translation of Chinese)
- Resorted translations using `npm run i18n:resort` for diff-stability later
- Updated documentation of the process of adding localizations to spell out the need to resort, and how to correctly create a new localization target.
- Added utility function `getLocalizedMessage` that allows dynamic access to the localized strings and updated the `map_locale_code_to_name` to use it so we don't need to also update the previous `switch` statement.

### Checklist
- [X] Ran `make test_e2e` locally and passed
- [ ] Linked to issue(s) above by issue number (e.g. `Closes #<issue-number>`)
- [X] One problem per PR (no unrelated changes)
- [X] Lints pass; CI green


ui/localization/messages/ja.json
```diff
@@ -438,15 +438,18 @@
...
-    "locale_sv": "Svenska",
-    "locale_zh": "中文 (简体)",
+    "locale_da": "デンマーク語",
+    "locale_de": "ドイツ語",
+    "locale_en": "英語",
+    "locale_es": "スペイン語",
+    "locale_fr": "フランス語",
+    "locale_it": "イタリア語",
+    "locale_ja": "日本語",
+    "locale_nb": "ノルウェー語 (ブックモール)",
```
IDisposable: @saltfield, please confirm if in Japanese we should be spelling out `(bokmål)` like this or would it be better to follow the other languages and just append the literal `(bokmål)` string?

IDisposable:
The design spelled out here are intended to ensure that when the UI is in English (the default), we show each language in their native language and when the UI is in any other language we show the languages in both the current language and in their native language.

Thus the language selector to looks like this in German:
image
(notice in 0.5.2 that the Chinese is incorrectly displaying only Chinese... this is because the line in the _de.json_  for `"locale_zh"`was not entered in Chinese, not as a German phrase.)

...and this for for English (the default language)
image

DEVELOPMENT.md
```diff
@@ -496,20 +496,24 @@ If you enable the [Sherlock](https://inlang.com/m/r7kp499g/app-inlang-ideExtensi
...
 
 ### Adding a new language
 
 1. Get the [ISO 3166-1 alpha-2 country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) (for example AT for Austria)
-2. Create a new file in the _ui/localization/messages_ directory (example _at.json_)
-3. Add the new country code to the _ui/localizations/settings.json_ file in both the `"locales"` and the `"languageTags"` section (inlang and Sherlock aren't exactly current to each other, so we need it in both places).
-4. That file also declares the baseLocale/sourceLanguageTag which is `"en"` because this project started out in English. Do NOT change that.
-5. Run `npm run i18n:machine-translate` to do an initial pass at localizing all existing messages to the new language.
+2. Create a **new empty** file in the _ui/localization/messages_ directory (example _at.json_)
+3. Add the new country code to the _ui/localization/settings.json_ file in both the `"locales"` and the `"languageTags"` section (inlang and Sherlock aren't exactly current to each other, so we need it in both places).
```
adamshiervani: `ui/localization/jetKVM.UI.inlang/settings.json`

adamshiervani (approved):
Looks good to me. Just one small docs change.

DEVELOPMENT.md
IDisposable: Good catch :)

Fixed

IDisposable:
Here's the results of the changes
image
image

ui/localization/messages/ja.json
saltfield: I recommend keeping it as (bokmål) to maintain consistency with other languages. In Japanese contexts, it's common to leave specific language variants or technical labels in their original script if that's the established pattern.

adamshiervani (approved):
Just need that Japanese bokmål fix and then we'r good to go! Great catch

IDisposable:
> Just need that Japanese bokmål fix and then we'r good to go! Great catch

Updated and pushed.

ui/localization/messages/ja.json
IDisposable: Done. THANKS for the quick review.

Issue #528, opened 2025-05-26
Show JetKVM Hostname in WebGUI


csuno:
Hy, as the subject tells...

i have two jetkvms and sometime heavy switching between these both... so sometimes it is difficult to identify the right one nad dont confuse myself ;)

So can you add an option tho display and customise the Heading of the Wbgui?

IDisposable:
Mostly done in https://github.com/jetkvm/kvm/pull/611

JackWa:
Hello all, as #617 has been marked a duplicate could displaying the HostName on the physical screen be then brought into scope for this Feature? (or #617 ticket reopened if inappropriate here)

We are utilizing multiple devices in an environment where it would be useful to see at a glance which JetKVM we are looking at.

Thank you.

adamshiervani:
Thanks for the request — this has been implemented in [#1163](https://github.com/jetkvm/kvm/pull/1163). The WebGUI now shows your device hostname in the header and browser tab title when a custom hostname is set in **Settings → Network**.

This is available on the dev channel (`0.5.5-dev202603170942`) and will be included in the next stable release.

@JackWa — displaying the hostname on the physical device screen isn't included here. Could you open a separate issue for that?

PR #1163, opened 2026-01-28
feat: display hostname in browser tab and nav bar

### LINKED ISSUE #83: Ability to Rename a JetKVM

DiamondCortex:
Closes #83 #528 

### Summary

This feature helps users with multiple JetKVM devices quickly identify which device a browser tab belongs to. The hostname appears in the browser tab title (format: "{hostname} - JetKVM") and in the navigation bar next to the logo. Users can enable this via a checkbox in Network settings, which is off by default to avoid disrupting existing workflows.

### Checklist

- [x] Lints pass; CI green

Image
Screenshot 2026-01-27 at 8 55 18 PM

### DIFF
--- a/ui/src/routes/devices.$id.tsx
+++ b/ui/src/routes/devices.$id.tsx
@@ -141,6 +141,7 @@ export default function KvmIdRoute() {
   const location = useLocation();
   const isLegacySignalingEnabled = useRef(false);
   const [connectionFailed, setConnectionFailed] = useState(false);
+  const [displayHostname, setDisplayHostname] = useState<string | null>(null);
 
   const navigate = useNavigate();
   const { otaState, setOtaState, setModalView } = useUpdateStore();
@@ -773,6 +774,17 @@ export default function KvmIdRoute() {
       console.debug("Setting HDMI state", hdmiState);
       setHdmiState(hdmiState);
     });
+
+    console.log("Requesting network settings");
+    send("getNetworkSettings", {}, (resp: JsonRpcResponse) => {
+      if ("error" in resp) return;
+      const result = resp.result as { hostname?: string };
+      if (result.hostname) {
+        setDisplayHostname(result.hostname);
+      } else {
+        setDisplayHostname(null);
+      }
+    });
   }, [rpcDataChannel?.readyState, send, setHdmiState]);
 
   const [needLedState, setNeedLedState] = useState(true);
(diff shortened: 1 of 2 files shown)


ui/src/routes/devices.$id.tsx
```diff
@@ -773,6 +774,17 @@ export default function KvmIdRoute() {
       console.debug("Setting HDMI state", hdmiState);
       setHdmiState(hdmiState);
     });
+
+    console.log("Requesting network settings");
```
IDisposable, line 778: This should probably be deleted or changed to `console.debug(`

IDisposable (approved):

ui/src/routes/devices.$id.tsx
DiamondCortex: I was mimicking the same behavior as the request right above at line 770. I can remove the log statement if you think its unnecessary though

BDworak:
Looking forward to this change merging :). Thanks for implementing @DiamondCortex

adamshiervani:
Thanks for taking the time to making the PR!

Let's simplify this: remove the configuration, and set the document title and show it in the header, if the user has overridden the default (jetkvm-xxxxx) hostname. It's not a complete solution, but good until we have time to implement a proper solution.

The proper solution would be to implement a separate on-device "name" property and remove the cloud-api device name, which is causing a real conflict. We currently use the cloud-api device name to list the devices, but instead of using the cloud-api database, we could simply just ask the device what the on-device "name" is and just render that. We already ask for the version number every time we list the devices in the cloud dashboard, so shouldn't be a big problem. To tie it all together, we could add an onboarding step asking the user to name the device from the start.

This would be a "proper" implementation and clean up some debt, but as this touches the cloud-api, websocket signaling, UI, I understand if third-party contributors don't have time to tackle this. Unfortunately, I don't either, right now.

Anyway, just do the removal of the configuration, and I'll merge it into dev. Thanks!

DiamondCortex:
@adamshiervani Sorry for the delay, I've simplified based on your feedback

IDisposable (approved):
This is ready to merge @adamshiervani

adamshiervani (approved):

PR #1185, opened 2026-02-03
Rate limit login access

adamshiervani:
### Summary

- Rate limit login attempts, to protect malicious local attackers or badly configured networks that expose the JetKVM.

### Checklist

- [x] Lints pass; CI green


web.go
```diff
@@ -639,6 +659,11 @@ func handleCreatePassword(c *gin.Context) {
 		return
 	}
 
+	if len(req.Password) < MinPasswordLength {
```
IDisposable, line 662: Won't this prevent someone who (foolishly, perhaps) set a short password from changing it to a longer more-compliant one?

IDisposable (approved):

web.go
adamshiervani: This is only for new password, login and old password should be able to still have fewer chars than `MinPasswordLength`

bt90:
I don't see any kind of subnet normalisation for IPv6. Even the smallest subnet assigned by providers has 2^64 addresses. Any rate limiting that targets a single IPv6 address is basically flawed.

https://adam-p.ca/blog/2022/02/ipv6-rate-limiting/

IDisposable:
> Any rate limiting that targets a single IPv6 address is basically flawed.

Indeed, but _perfect is the enemy of good enough to start_... 

PR to add IPv6 subset normalization is welcome ;) @bt90

PR #1187, opened 2026-02-04
feat: Add detached window with optional toolbar and hostname display

tvinhas:
### Summary

  Adds a "Detach" button to open the KVM video stream in a separate window. The detached window can optionally show/hide toolbars (configurable in Settings > Appearance), with the window size automatically adjusting. The window title
  displays "JetKVM: {hostname}" for easy identification.

  ### Checklist

  - [x] Lints pass; CI green


tvinhas:
image

tvinhas:
This PR adds a "Detach" button to the top bar that when clicked opens a browser popup with minimal UI so the user can have multiple KVMs open with minimal distractions, without wasting space in the monitor.

image

ui/src/components/DetachedToolbar.tsx
```diff
@@ -0,0 +1,73 @@
+import { LuX } from "react-icons/lu";
```
adamshiervani: Remove this and simply use the replace the fullscreen buttons in the Actionbar with a Close button

ui/src/routes/devices.$id.detached.tsx
```diff
@@ -0,0 +1,550 @@
+import { useCallback, useEffect, useRef, useState } from "react";
```
adamshiervani: By creating a new `/detached` route, you're missing a bunch of important code and duplicating the most complex part of the application, which is the signaling. Just re-use the `device.$id.tsx` route and simply add a query parameter for the detached mode.

ui/src/hooks/useDetachedWindow.ts
```diff
@@ -0,0 +1,53 @@
+import { isOnDevice } from "@/main";
+import { useSettingsStore } from "@hooks/stores";
+
+// Module-level Map to track windows (avoids serialization issues)
+const windowMap = new Map<string, Window>();
+
+// Height constants for window sizing
+const VIDEO_HEIGHT = 720;
+const DETACHED_TOOLBAR_HEIGHT = 32; // h-8 = 32px
+const ACTION_BAR_HEIGHT = 40; // min-h-[39.5px] ≈ 40px
```
adamshiervani: This wont last long, I would suggest just set an arbitrary window size as it's the initial window size. The user can resize according to their needs if they don't like it.

adamshiervani (changes requested):
Thanks for taking the time implementing this!

Overall feedback - can be much simpler and duplicate less code. The Action bar will get too crowded with another button, but I'll fix that once this is merged, as I need to rethink the entire navigation.

ui/src/hooks/useDetachedWindow.ts
tvinhas: Simplified, now just opens at a fixed 1280x720. Removed all the height constant calculations, that was really awkward. The user can resize to their liking. See commit: https://github.com/jetkvm/kvm/pull/1187/changes/a7c29c491a1e43c68595f2ccc130d8a9f08a91ca

ui/src/routes/devices.$id.detached.tsx
tvinhas: Good call. I deleted the separate detached route entirely and switched to a ?detached=true query param on the existing device route. The detached window now shares the full render tree with !isDetachedWindow guards to hide the navbar, sidebar, focus trap, modals, and terminals. All signaling, HID, and connection logic is reused with zero duplication. See commit https://github.com/jetkvm/kvm/pull/1187/changes/9e3fcde855eaba9d8be722201c0fb9f71488cf0a

ui/src/components/DetachedToolbar.tsx
tvinhas: Done. Deleted DetachedToolbar.tsx and replaced the Detach/Fullscreen buttons in ActionBar with a Close button when in detached mode. Removed the showDetachedToolbar setting since the ActionBar is now always visible. See commit: https://github.com/jetkvm/kvm/pull/1187/changes/6d55e5f0803d7764241f7093e2623c5e92cbc8c3

IDisposable (approved):

adamshiervani (changes requested):
- Opening the Virtual Media modal, will make it go away from detached mode
- Clicking on Connection Stats doesn't open the sidebar

tvinhas:
> * Opening the Virtual Media modal, will make it go away from detached mode
> * Clicking on Connection Stats doesn't open the sidebar

> * Opening the Virtual Media modal, will make it go away from detached mode
> * Clicking on Connection Stats doesn't open the sidebar

Good catch! I never used neither of those functions so I didn't realize they were broken. Fixed both:
 1. Virtual Media: navigateTo was dropping the ?detached=true query param during navigation, and the modal was guarded by !isDetachedWindow. Fixed the navigation helper to preserve the detached param and removed
 the modal guard.
 2. Connection Stats: The sidebar container wasn't rendered in detached mode. Removed the guard so it renders.

adamshiervani (changes requested):
Web Terminal doesn't work. Oddly enough, it works i you open the Virtual Media modal, but then the main big header pops up too.

tvinhas:
> Web Terminal doesn't work. Oddly enough, it works i you open the Virtual Media modal, but then the main big header pops up too.

Fixed both issues:

1. Web Terminal: The Terminal components were guarded by !isDetachedWindow, preventing them from rendering in detached mode. Removed the guards.
2. Header appearing after Virtual Media: The mount route used raw navigate("..") which drops the ?detached=true search param, causing isDetachedWindow to become false. Switched to navigateTo("/") which preserves the detached param, and refactored the Dialog's internal navigation to use its onClose callback for consistency.

tvinhas:
@adamshiervani I just rebased this against dev since you've been busy merging a lot of stuff to it.

adamshiervani (approved):

PR #1188, opened 2026-02-04
Signed Releases

adamshiervani:
### Summary

- Ensure that updates are signed by trusted key

### Checklist

- [x] Lints pass; CI green


internal/ota/gpg.go
```diff
@@ -0,0 +1,275 @@
...
+
+	return key, nil
+}
+
+// updateMemoryCache updates the in-memory key cache
+func (g *GPGVerifier) updateMemoryCache(key []byte) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	g.cachedKey = make([]byte, len(key))
```
IDisposable, line 167: Should we be storing the key in the cache before verifying it's valid (with the `ReadArmoredKeyRing`)... should this come lower down after `g.keyring = keyring`?

IDisposable (commented):
Looks pretty good... will this make local development testing of upgrades harder?

internal/ota/gpg.go
adamshiervani: Higher up in the call stack, in `fetchFromSingleKeyserver` we already validate the key:

```go
key, err := g.fetchFromSingleKeyserver(ctx, url)
if err != nil {
	g.logger.Debug().Err(err).Str("url", url).Msg("keyserver failed")
	errors = append(errors, fmt.Errorf("%s: %w", url, err))
	continue
}

g.logger.Info().Str("url", url).Msg("successfully fetched public key")
return key, nil
}
```

However, given we need to call `openpgp.ReadArmoredKeyRing` regardless of previous validation, I'll just call it before we set it to the cache.

adamshiervani:
The PR has a bunch of new OTA tests and also fixes a bunch of flaky tests. You can just skim through those, but focus the review mainly on the OTA code path. Would love to have your eyes on that to iron out any issues before we actually hit production. <3

PR #1207, opened 2026-02-18
feat: Added copy text to clipboard from kvm using OCR

tvinhas:
### Summary

- Adds a "Copy text" button to that allows the user to select any text in the window, and OCR will transform that in text in a modal popup in the browser that the user can copy to his local clipboard. Supports indentation for copying code as well.

### Checklist

- [X] Ran `make test_e2e` locally and passed
- [X] Linked to issue(s) above by issue number (e.g. `Closes #<issue-number>`)
- [X] One problem per PR (no unrelated changes)
- [X] Lints pass; CI green
- [X] Tricky parts are commented in code


tvinhas:
image

tvinhas:
image

ui/src/components/OcrOverlay.tsx
```diff
@@ -0,0 +1,474 @@
...
+          </div>
+        )}
+      </motion.div>
+
+      {/* OCR Result panel — portaled to document.body to escape the video
+          container's stacking context and select-none inheritance */}
+      {status === "result" &&
+        createPortal(
+          <div className="pointer-events-none fixed inset-0 z-50 flex items-center justify-center">
+            <div className="pointer-events-auto mx-4 flex max-h-[80%] w-full max-w-md flex-col rounded-lg border border-slate-200 bg-white shadow-xl dark:border-slate-700 dark:bg-slate-800">
```
adamshiervani: Please use `<ConfirmDialog>`, which uses `<Modal>` which contains the mounting animations too, so no react motion needed here.

ui/src/components/OcrOverlay.tsx
```diff
@@ -0,0 +1,474 @@
...
+                </span>
+                <button
+                  className="rounded-sm p-1 text-slate-400 hover:bg-slate-100 hover:text-slate-600 dark:hover:bg-slate-700 dark:hover:text-slate-300"
+                  onClick={() => setOcrMode(false)}
+                >
+                  <LuX className="h-4 w-4" />
+                </button>
+              </div>
+              <div className="px-4 py-3">
+                <TextArea
```
adamshiervani: If possible, make this readonly and highlight all the text. I know `Ctrl+C` works, but it feels weird if i'm copying a non-selected text. 

If you can highlight the text, you don't don't need to the copy hint below

ui/src/components/OcrOverlay.tsx
```diff
@@ -0,0 +1,474 @@
...
+        )}
+
+        {/* Selection rectangle with size indicator */}
+        {selectionRect && selectionStyle && status !== "result" && (
+          <div
+            className="absolute border-2 border-dashed border-blue-400 bg-blue-400/10"
+            style={selectionStyle}
+          >
+            {selectionRect.width >= 10 && selectionRect.height >= 10 && (
+              <div className="absolute right-0 -bottom-6 rounded bg-black/70 px-1.5 py-0.5 text-[10px] font-medium text-white tabular-nums">
```
adamshiervani: Use the `<Card />`

ui/src/components/OcrOverlay.tsx
```diff
@@ -0,0 +1,474 @@
...
+              </div>
+              <div className="px-4 py-3">
+                <TextArea
+                  ref={resultRef}
+                  value={ocrResult}
+                  onChange={e => setOcrResult(e.target.value)}
+                  rows={Math.min(10, ocrResult.split("\n").length + 1)}
+                />
+              </div>
+              <div className="border-t border-slate-200 px-4 py-3 dark:border-slate-700">
```
adamshiervani: The `<ConfirmDialog>` pushes for this, but just as a note - it doesn't have any X, and uses CTA's for the main actions.

I would suggest removing this label, and have two CTAs - `Cancel` and `Copy`. They both also get corresponding shortcuts - `Esc` and `CMD+C`/ `Ctrl+C`

ui/src/components/OcrOverlay.tsx
```diff
@@ -0,0 +1,474 @@
...
+        style={{ cursor: status === "result" ? "default" : "crosshair" }}
+        onMouseDown={handlePointerDown}
+        onMouseMove={handlePointerMove}
+        onMouseUp={handlePointerUp}
+        onTouchStart={handlePointerDown}
+        onTouchMove={handlePointerMove}
+        onTouchEnd={handlePointerUp}
+      >
+        {/* Semi-transparent background */}
+        <div className="absolute inset-0 bg-black/20" />
```
adamshiervani: This doesn't go all the way to the border end of the viewport

ui/src/components/OcrOverlay.tsx
```diff
@@ -0,0 +1,474 @@
...
+        {status === "idle" && (
+          <div className="pointer-events-none absolute inset-x-0 top-4 flex justify-center">
+            <div className="rounded-md bg-black/70 px-3 py-1.5 text-xs font-medium text-white">
+              {m.ocr_drag_to_select()}
+            </div>
+          </div>
+        )}
+
+        {/* Processing indicator */}
+        {status === "processing" && (
```
adamshiervani: The OCR can take some time, this is too tiny to indicate processing. Additionally, users might want to cancel the processing. You could use the same `<ConfirmDialog>` as the `result` status, but with a:
- Skelleton Loading with label on top to indicate processing
- Disabled `Copy` CTA - only `Cancel`

ui/src/components/ActionBar.tsx
```diff
@@ -63,6 +76,14 @@ export default function Actionbar({
               onClick={() => setTerminalType(terminalType === "kvm" ? "none" : "kvm")}
             />
           )}
+          <Button
```
adamshiervani, line 79: Note: The Action bar is already too crowded. You don't need to change anything here, but once I get time I'll rethink the action bar, so. just letting you know that I'll probably move after it merges with `dev`.

ui/localization/messages/en.json
```diff
@@ -49,6 +49,7 @@
     "access_update_tls_settings": "Update TLS Settings",
```
adamshiervani, line 49: Once the Review comments have been addressed make sure to machine translate to other languages too.  See [DEVELOPMENT.md](https://github.com/jetkvm/kvm/blob/dev/DEVELOPMENT.md#localization ) for details.

adamshiervani (changes requested):
Thanks for taking the time developing this!

ui/src/components/ActionBar.tsx
tvinhas: I agree, which is why I have another PR to detach the WebRTC screen. This way I have a screen without any clutter or information, just like I'm using RDP in a window.

ui/src/components/OcrOverlay.tsx
tvinhas: Thanks! Fixed in this commit:
https://github.com/jetkvm/kvm/pull/1207/changes/f523878e5ffa04f139e181d29afa61727214e220

ui/src/components/OcrOverlay.tsx
tvinhas: Done. Swapped the size indicator to use <Card />. See commit: https://github.com/jetkvm/kvm/pull/1207/changes/184f1c4d2c5904c7dec486618f592d209cc04757

ui/src/components/OcrOverlay.tsx
tvinhas: Replaced the processing indicator with a <ConfirmDialog> that shows a skeleton loading animation, a disabled "Copy text" button, and an active "Cancel" button. Clicking Cancel exits the overlay and the in-flight OCR result is discarded. See commit: https://github.com/jetkvm/kvm/pull/1207/changes/0c8daeb86b4a23ca60176561ef90e446af399603

ui/src/components/OcrOverlay.tsx
tvinhas: Replaced with `<ConfirmDialog>`. It now handles the modal backdrop, mounting animations, and Cancel/Copy CTAs. Removed createPortal, react-dom, and LuX imports that are no longer needed. See commit: https://github.com/jetkvm/kvm/pull/1207/changes/02a641cdf95ff45af0116c5839e96d8fed042118

ui/src/components/OcrOverlay.tsx
tvinhas: This was one I had my doubts when I was testing it. My goal in making it editable is that sometimes OCR can make little mistakes and this would allow me to correct the text prior to copying it, but you are right, it's weird and shouldn't be done at the KVM screen.

tvinhas: Dome. `<TextArea>` is now readOnly with all text auto-selected on mount. The copy hint was already removed in the ConfirmDialog refactor. See commit https://github.com/jetkvm/kvm/pull/1207/changes/39a3056f476d538685a98fbfbe7dd1c457b1b040

ui/src/components/OcrOverlay.tsx
tvinhas: Done. The X button is gone (ConfirmDialog doesn't render one), replaced by Cancel and Copy CTAs. Esc closes the dialog, Cmd+C/Ctrl+C copies and closes. Removed the old copy hint label and added a dedicated ocr_copy key for the shorter "Copy" button text. See commit https://github.com/jetkvm/kvm/pull/1207/changes/2a5ad15daee4ae9eaf4042f11565e939853152ad

ui/localization/messages/en.json
tvinhas: Done. Ran the i18n pipeline (resort, machine-translate, compile, validate). The new ocr_copy key is translated across all 12 languages and the removed ocr_copy_hint key has been cleaned up. Build passes. See commit https://github.com/jetkvm/kvm/pull/1207/changes/5c6c276ef2d8c3e14342b5924c0d0510069be751

adamshiervani (changes requested):
- The "Copy" button is clickable in the `processing` state. I think the <ConfirmDialog/> needs to remove the `cursor-pointer` on disabled state. 
- Be explicit in the CTA. "Copy" "Copy text".
- The "Select area" state breaks, when clicking on in the video, and then moving the mouse outside of the viewport and then coming back.
- Extend the `<ConfirmDialog />` with a `children` prop and use the description to say something about that it might take a few seconds. 
- Clicking "Copy" doesn't work. I'm getting:
```
index-C3OdXBs2.js:5 Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')
    at onConfirm (index-C3OdXBs2.js:5:68)
```

tvinhas:
> * The "Copy" button is clickable in the `processing` state. I think the  needs to remove the `cursor-pointer` on disabled state.
> * Be explicit in the CTA. "Copy" "Copy text".
> * The "Select area" state breaks, when clicking on in the video, and then moving the mouse outside of the viewport and then coming back.
> * Extend the `<ConfirmDialog />` with a `children` prop and use the description to say something about that it might take a few seconds.
> * Clicking "Copy" doesn't work. I'm getting:
> 
> ```
> index-C3OdXBs2.js:5 Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')
>     at onConfirm (index-C3OdXBs2.js:5:68)
> ```

  Thanks Adam! All 5 addressed:

  1. Wrapped the ConfirmDialog confirm button to apply cursor-not-allowed when disabled.
  2. Changed CTA from "Copy" to "Copy text".
  3. Added onMouseLeave on the overlay to reset the selection when the pointer leaves during a drag.
  4. Extended ConfirmDialog with a children prop. Processing dialog uses description for "This may take a few seconds" and children for the skeleton. Same for the result dialog.
  5. Added fallback for navigator.clipboard, it now  uses document.execCommand('copy') when the Clipboard API is unavailable.

adamshiervani (changes requested):
- The loading modal card shouldn't be in separate Modal as it animates in and out super quickly if the text short.
- The selection area size pill shows a dark text in dark mode - it should be white.
- The copy to clipboard toast is shown twice
- The exit animation on the copy modal doesn't work - see the modal in Network settings.

tvinhas:
> * The loading modal card shouldn't be in separate Modal as it animates in and out super quickly if the text short.
> * The selection area size pill shows a dark text in dark mode - it should be white.
> * The copy to clipboard toast is shown twice
> * The exit animation on the copy modal doesn't work - see the modal in Network settings.

Thanks for the feedback! All four items addressed in the latest commit:

1. Loading modal flicker: Merged the processing and result states into a single <ConfirmDialog> that stays open across both states. The title, description, and children update dynamically, so there's no close→reopen animation when OCR completes quickly.
2. Dark mode size pill text: Added dark:text-white to the selection size indicator <Card>.
3. Double toast: The document.execCommand("copy") fallback was showing its own toast while also triggering the copy event listener which showed another. Removed the duplicate from the fallback since the copy event listener handles it.
4. Exit animation: The issue was that setOcrMode(false) immediately unmounted the component, preventing HeadlessUI from playing the Dialog leave transition. Added a closeOverlay() function that first sets the dialog's open to false (via status reset), waits for the 200ms leave transition, then unmounts.

adamshiervani (changes requested):
- When closing the modal the text quickly flashed back to "Recognizing text". It should stay the same while exiting the modal

tvinhas (commented):
Good catch! The issue is that `closeOverlay` sets `status` to `"idle"` to trigger the dialog's leave transition, but the dialog content (title, description) also depends on `status`. So during the 200ms exit animation, the ternary `status === "result" ? "Copy text" : "Recognizing text..."` falls through to "Recognizing text..." since status is now "idle".                                                                                                              │

Fixed by adding an `isClosing` flag that closes the dialog without resetting `status`, keeping the content stable during the animation.

adamshiervani (approved):

PR #1244, opened 2026-03-07
added noauth streaming for multiple clients

LandonRawson:
Please go to the `Preview` tab :point_up: and select the appropriate sub-template:

* [Feature](?expand=1&template=feature.md)
* [Bugfix](?expand=1&template=bugfix.md)

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -1,0 +1,487 @@
+package kvm
+
+import (
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/pion/webrtc/v4/pkg/media"
+)
+
+var (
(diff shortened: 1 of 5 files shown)


stream.go
```diff
@@ -0,0 +1,487 @@
...
+			continue
+		}
+
+		if err := session.VideoTrack.WriteSample(media.Sample{Data: frame, Duration: duration}); err != nil {
+			nativeLogger.Warn().Err(err).Msg("error writing sample to stream viewer session")
+		}
+	}
+}
+
+const streamViewerHTML = `<!doctype html>
```
IDisposable, line 129: Embedding this HTML file as a constant is not the best. Can we cache an ACTUAL HTML file instead?

IDisposable:
Can you give a rationale for why we would switch to this?

PR #1250, opened 2026-03-08
chore: Add .nvmrc for Node.js version management

gentslava:
Added a .nvmrc file to define the recommended Node.js version for the project. This makes environment setup easier, allows developers to quickly switch to the correct version with nvm use.


IDisposable (approved):
This matches the current .devcontainer so I think it's risk-free to merge @adamshiervani

PR #1267, opened 2026-03-11
fix(security): add missing HTTP security headers

flightlesstux:
## Summary

Adds a security headers middleware to all responses:

- `X-Frame-Options: DENY` — prevents clickjacking
- `X-Content-Type-Options: nosniff` — prevents MIME type sniffing
- `X-XSS-Protection: 1; mode=block` — enables browser XSS filter

## Test plan
- [ ] All responses include the three new headers
- [ ] Existing functionality unaffected


adamshiervani (approved):

patschi:
Please note that `X-XSS-Protection` has been deprecated - also seen [as per Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-XSS-Protection). Looks like Claude isn't aware of that yet. I'd suggest removing it.

To quote:
>Warning: Even though this feature can protect users of older web browsers that don't support [CSP](https://developer.mozilla.org/en-US/docs/Glossary/CSP), in some cases, X-XSS-Protection can create XSS vulnerabilities in otherwise safe websites. See the [Security considerations](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-XSS-Protection#security_considerations) section below for more information.

PR #1268, opened 2026-03-11
fix(security): enforce WebSocket origin check

flightlesstux:
## Summary

The local WebRTC signaling WebSocket handler had `InsecureSkipVerify: true` explicitly set, which allows connections from **any** origin. This means a malicious website could establish a WebSocket connection to the device using the victim's authenticated session cookie (CSRF via WebSocket).

**Fix**: Replace with `OriginPatterns` restricted to the device's own host and loopback addresses, so only same-origin and local connections are accepted.

## Test plan
- [ ] WebSocket connection from the device UI (same origin) works normally
- [ ] WebSocket connection from a foreign origin is rejected with 403
- [ ] Cloud connections (separate handler, unaffected) still work


agh:
The CSRF-via-WebSocket vector here is real in principle but worth examining the actual threat model and breakage risk before merging.

**The vulnerability:**

`InsecureSkipVerify: true` means the WebSocket endpoint at `/webrtc/signaling/client` accepts connections regardless of the `Origin` header. A malicious webpage could open a WebSocket to `ws://<device-ip>/webrtc/signaling/client` — the browser would attach the `authToken` cookie automatically (it's not `SameSite` restricted and domain is empty, so it defaults to the device's host), and the device would accept the connection. The attacker could then perform WebRTC signaling and potentially establish a full KVM session.

This is a legitimate concern. However:

**Exploitation prerequisites are narrow:**
- Attacker must know the victim's JetKVM IP address on their LAN
- Victim must be authenticated (have a valid `authToken` cookie for that IP)
- Victim must visit the attacker's page while on the same network
- If `localAuthMode` is `noPassword`, the auth middleware passes everything through anyway (line 535-537), so origin checking doesn't help — anyone on the LAN already has full access

**The proposed fix has breakage risks:**

The fix restricts `OriginPatterns` to `[c.Request.Host, "localhost", "127.0.0.1", "[::1]"]`. This checks the `Origin` header against these patterns. Scenarios where this breaks:

1. **User accesses by IP, browser sends hostname as Origin** — user types `http://192.168.1.100`, browser's `Origin` is `http://192.168.1.100`, `c.Request.Host` is `192.168.1.100`. This should match. OK.

2. **User accesses by mDNS name** — `http://jetkvm.local` → Origin is `http://jetkvm.local`, Host is `jetkvm.local`. Should match. OK.

3. **User accesses by hostname, device has a different canonical name** — if DNS returns a different name than what the user typed, `Host` and `Origin` could mismatch. Edge case but possible in enterprise DNS environments.

4. **Reverse proxy / port forwarding** — user accesses via `https://kvm.home.example.com:8443` which proxies to the device. The `Origin` header would be `https://kvm.home.example.com:8443` but `c.Request.Host` at the device would be the proxy's forwarded host or the device's own IP. This is a common home lab setup and would break.

5. **Cloud path** — the PR description says "Cloud connections (separate handler, unaffected)" which is correct — cloud uses a different WebSocket path. But if the cloud UI ever needs to fall back to direct local WebSocket signaling, it would be blocked (`app.jetkvm.com` origin vs device IP host).

6. **Port mismatch** — `nhooyr.io/websocket`'s `OriginPatterns` matching includes the port. If the user accesses on a non-standard port (common behind reverse proxies), the match could fail even with the right hostname.

**A more robust fix would be:**

Set `SameSite=Lax` (or `Strict`) on the `authToken` cookie. This prevents the browser from sending the cookie on cross-origin requests in the first place — no origin checking needed on the WebSocket side. Currently the cookie has no `SameSite` attribute, which defaults to `Lax` in modern browsers but isn't explicitly set.

This approach is simpler, doesn't risk breaking any access patterns, and addresses the root cause (cookie being sent cross-origin) rather than the symptom (WebSocket accepting cross-origin connections).

adamshiervani:
All valid points by @agh. SameSite=Lax is the better approach

PR #1269, opened 2026-03-11
fix(security): remove wildcard CORS from /device/status

flightlesstux:
## Summary

`/device/status` had `Access-Control-Allow-Origin: *` set, allowing any website to query whether the device has been set up. While the endpoint only returns `{isSetup: bool}`, exposing device state to arbitrary origins is unnecessary — the frontend is served from the same origin and requires no CORS headers to access it.

**Fix**: Remove the wildcard CORS headers and OPTIONS preflight handler entirely.

## Test plan
- [ ] `/device/status` still returns correct response from the device UI
- [ ] No `Access-Control-Allow-Origin` header present in response
- [ ] Cross-origin requests from foreign sites no longer get the response


agh:
A few concerns with this change.

`/device/status` returns `{isSetup: bool}` — a single boolean. The wildcard CORS was likely added deliberately to support cross-origin access patterns:

1. **Cloud dashboard** — `app.jetkvm.com` queries device status from the device's LAN IP. That's a cross-origin request (cloud domain → device IP). Removing `Access-Control-Allow-Origin: *` would cause the browser to block these responses, potentially breaking the cloud registration/adoption flow.

2. **Local access via different origins** — users access JetKVM by IP (`10.x.x.x`), mDNS hostname (`jetkvm.local`), or custom DNS. If a tool or integration checks setup status from a different origin than the one serving the UI, it would silently fail.

3. **The endpoint is already public and unauthenticated by design** — it has to be, because it runs before auth exists (the whole point is to check whether setup has happened). The information it exposes — a single boolean — has no security value to an attacker. Knowing a device on your LAN is or isn't configured doesn't enable any exploit.

The CORS headers here aren't papering over a vulnerability. They're enabling a public endpoint to be callable cross-origin, which is the correct behavior for its purpose. Removing them tightens something that doesn't need tightening and risks breaking the cloud integration path.

If the concern is CORS hygiene across the codebase, a better approach would be to audit which endpoints genuinely need cross-origin access and document why, rather than removing it from an endpoint that was intentionally designed to be public.

adamshiervani:
All valid points by @agh

PR #1276, opened 2026-03-15
feat(network): add Tailscale status to Settings > Network

agh:
Tailscale runs on JetKVM devices but has no visibility in the web UI. This adds a read-only status card to the network settings page that surfaces the output of `tailscale status --json` over the existing JSON-RPC transport.

## Backend (Go)

**tailscale.go** -- `TailscaleStatus`/`TailscalePeer` structs returned by a single RPC handler (`getTailscaleStatus`). `isTailscaleInstalled` uses `exec.LookPath`; `getTailscaleStatus` shells out via `exec.CommandContext` with a 10s timeout and parses the JSON response. When the binary is absent the handler returns `{installed: false}` without error. When the daemon is unreachable it returns `{installed: true, running: false}`.

**tailscale_test.go** -- Table-driven tests for `parseTailscaleStatus` covering Running, NeedsLogin, Stopped, Starting, malformed JSON, and empty object inputs. Integration-level tests for `getTailscaleStatus` verify the exec mock path and confirm the function never returns an error regardless of environment state.

**jsonrpc.go** -- `getTailscaleStatus` registered in `rpcHandlers`.

**log.go** -- `tailscaleLogger` subsystem added.

## Frontend (TypeScript/React)

**TailscaleCard.tsx** -- `GridCard` component that calls `getTailscaleStatus` on mount. Renders nothing when `installed=false`. Shows a status badge (Connected / Needs Login / Stopped), Tailscale IPv4/IPv6, hostname, DNS name, auth URL when NeedsLogin, and health warnings when present.

**stores.ts** -- `TailscaleStatus` and `TailscalePeer` interfaces.

**devices.$id.settings.network.tsx** -- `TailscaleCard` rendered after `PublicIPCard`.

## Graceful degradation

- Card is hidden entirely when Tailscale is not installed (no empty state, no placeholder)
- Daemon unreachable: shows installed-but-stopped state
- RPC error: card does not render
- Network settings page is never blocked by Tailscale status failures

### DIFF
--- a/ui/src/components/TailscaleCard.tsx
+++ b/ui/src/components/TailscaleCard.tsx
@@ -67,14 +67,14 @@ export default function TailscaleCard() {
               {ipv4 && (
                 <div className="flex justify-between border-t border-slate-800/10 pt-2 dark:border-slate-300/20">
                   <span className="text-sm text-slate-600 dark:text-slate-400">IPv4</span>
-                  <span className="font-mono text-sm font-medium">{ipv4}</span>
+                  <span className="font-mono text-[13px] font-medium">{ipv4}</span>
                 </div>
               )}
 
               {ipv6 && (
                 <div className="flex justify-between border-t border-slate-800/10 pt-2 dark:border-slate-300/20">
                   <span className="text-sm text-slate-600 dark:text-slate-400">IPv6</span>
-                  <span className="font-mono text-sm font-medium">{ipv6}</span>
+                  <span className="font-mono text-[13px] font-medium">{ipv6}</span>
                 </div>
               )}
 
@@ -110,16 +110,6 @@ export default function TailscaleCard() {
               Tailscale is installed but not running.
               {status.backendState && ` State: ${status.backendState}`}
             </p>
-          )}
-
-          {status.health && status.health.length > 0 && (
-            <div className="space-y-1 border-t border-slate-800/10 pt-2 dark:border-slate-300/20">
-              {status.health.map((warning, idx) => (
-                <p key={idx} className="text-xs text-amber-600 dark:text-amber-400">
-                  {warning}
-                </p>
-              ))}
-            </div>
           )}
         </div>
       </div>
--- a/ui/src/components/TailscaleCard.tsx
+++ b/ui/src/components/TailscaleCard.tsx
@@ -1,0 +1,150 @@
+import { LuRefreshCcw } from "react-icons/lu";
+import { useCallback, useEffect, useState } from "react";
+
+import { Button } from "@components/Button";
+import { GridCard } from "@components/Card";
+import { TailscaleStatus } from "@hooks/stores";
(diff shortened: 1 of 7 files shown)


agh:
Shout if you'd like anything changed, I tried to follow the same style and approach as the rest of the project!

ui/src/components/TailscaleCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+          )}
+
+          {!status.running && status.backendState !== "NeedsLogin" && (
+            <p className="pt-2 text-sm text-slate-600 dark:text-slate-400">
+              Tailscale is installed but not running.
+              {status.backendState && ` State: ${status.backendState}`}
+            </p>
+          )}
+
+          {status.health && status.health.length > 0 && (
```
adamshiervani: Remove the raw health warnings section - these are internal Tailscale diagnostics (e.g. nftables errors) that aren't actionable or meaningful to end users.

Image

ui/src/components/TailscaleCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+                <div className="flex justify-between border-t border-slate-800/10 pt-2 dark:border-slate-300/20">
+                  <span className="text-sm text-slate-600 dark:text-slate-400">IPv4</span>
+                  <span className="font-mono text-sm font-medium">{ipv4}</span>
+                </div>
+              )}
+
+              {ipv6 && (
+                <div className="flex justify-between border-t border-slate-800/10 pt-2 dark:border-slate-300/20">
+                  <span className="text-sm text-slate-600 dark:text-slate-400">IPv6</span>
+                  <span className="font-mono text-sm font-medium">{ipv6}</span>
```
adamshiervani: The monospace font looks larger, set to `text-[13px]`

ui/src/components/TailscaleCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+                <div className="flex justify-between border-slate-800/10 pt-2 dark:border-slate-300/20">
+                  <span className="text-sm text-slate-600 dark:text-slate-400">Hostname</span>
+                  <span className="text-sm font-medium">{status.self.hostName}</span>
+                </div>
+              )}
+
+              {ipv4 && (
+                <div className="flex justify-between border-t border-slate-800/10 pt-2 dark:border-slate-300/20">
+                  <span className="text-sm text-slate-600 dark:text-slate-400">IPv4</span>
+                  <span className="font-mono text-sm font-medium">{ipv4}</span>
```
adamshiervani: The monospace font looks larger, set to `text-[13px]`

adamshiervani (changes requested):

agh (commented):
Pushed — dropped the health warnings block entirely and set `text-[13px]` on both IP spans. Good catch on the mono sizing; it was definitely rendering taller than the adjacent labels.

adamshiervani (approved):

PR #1289, opened 2026-03-15
feat(network): add custom NTP/HTTP time sync configuration UI

### LINKED ISSUE #516: Add option to configure NTP server(s) used by JetKVM device

agh:
Closes #516, #645, #59

## Problem

The backend supports custom NTP servers, HTTP URLs, source ordering, parallel queries, and fallback control for time synchronization, but the frontend only exposes three presets (NTP only, NTP and HTTP, HTTP only). Users who need to specify their own NTP server have no way to do so through the UI.

## Why a new PR instead of landing #1102

PR #1102 implemented full custom time sync UI but stalled for 3 months (last activity Jan 2026). It exposed every backend field as a direct UI control — source ordering with a drag/drop Combobox, parallel query count, disable-fallback toggle — which drew UX concerns and a debate about DHCP priority ordering. The author acknowledged it was a "first-cut" and planned to simplify but did not return.

This PR ships the functionality users actually requested — custom NTP servers and HTTP URLs — with a minimal UI surface that is easy to review and extend later:

| | #1102 | This PR |
|---|---|---|
| Additions | 753 | ~270 (130 non-localization) |
| Changed files | 15 | 18 (13 are localization) |
| New components | 1 (388 lines) + Combobox modifications | 1 (140 lines), no shared component changes |
| Scope | Full source ordering, parallel queries, fallback toggle | NTP server list + HTTP URL list |

The advanced fields (`TimeSyncOrdering`, `TimeSyncParallel`, `TimeSyncDisableFallback`) retain their backend defaults and can be surfaced in a follow-up PR if there is demand.

## Changes

**Backend (Go):**

- `confparser.go` — add `hostname_or_ipv4_or_ipv6` validation type so NTP server fields accept hostnames like `pool.ntp.org`, not just raw IPs
- `config.go` — change `TimeSyncNTPServers` validation from `ipv4_or_ipv6` to `hostname_or_ipv4_or_ipv6`

**Frontend (TypeScript/React):**

- `CustomTimeSyncCard.tsx` — new component following the same list-with-add/remove pattern as `StaticIpv4Card` DNS fields. NTP server inputs validated as hostname or IP, HTTP URL inputs validated with `validator.isURL`
- `stores.ts` — add optional `time_sync_ordering`, `time_sync_ntp_servers`, `time_sync_http_urls`, `time_sync_disable_fallback`, `time_sync_parallel` to `NetworkSettings` interface
- `devices.$id.settings.network.tsx` — uncomment Custom dropdown option, conditionally render `CustomTimeSyncCard` when selected

**Localization:**

8 new keys translated across all 13 supported languages (da, de, en, es, fr, it, ja, nb, pt, ru, sv, zh, zh-tw).

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes time sync source ordering/validation and triggers an automatic re-sync on network config updates, which can affect devices’ system time behavior after settings changes.
> 
> **Overview**
> Adds support for **custom time synchronization** by extending network config to accept user-provided NTP servers and HTTP time URLs, including a new `custom` mode with explicit source ordering.
> 
> Backend validation is updated so NTP servers can be specified as *hostnames or IPs* (new `hostname_or_ipv4_or_ipv6` validator), and the NTP client now passes through hostnames instead of filtering them out.
> 
> When network settings are applied, the running time sync service is updated with the new network config and a background `Sync()` is triggered; UI localization files add new/updated strings (notably for time sync inputs) across multiple languages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 10e2220909df9f72e468842db03aeac1645e1274. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


adamshiervani (changes requested):
I tested this PR on a device and found a few issues preventing custom NTP servers from actually being used. I don't have push access to your fork, so here are the findings:

#### 1. Hostnames silently filtered out at runtime

`filterNTPServers` in `internal/timesync/ntp.go` uses `net.ParseIP()` to check each server entry. Hostnames like `pool.ntp.org` return `nil` from `ParseIP` and get skipped via `continue`. The config parser accepts them fine (`hostname_or_ipv4_or_ipv6`), but this separate runtime code path quietly drops them. The NTP library handles DNS resolution itself, so hostnames should be passed through to the filtered list instead of skipped.

#### 2. Ordering never includes `ntp_user_provided`

The backend default for `TimeSyncOrdering` is `["ntp", "http"]`, which tells the sync loop to use the built-in fallback servers. When custom mode is selected, the ordering needs to include `ntp_user_provided` (and `http_user_provided` if HTTP URLs are set) so the sync loop actually queries the user's servers. This can be done in `prepareSettings` when `time_sync_mode === "custom"`. Just put the `x_user_provided` at the start of the array and call it day.

#### 3. Stale config pointer in TimeSync

`TimeSync` is initialized at startup with a pointer to `config.NetworkConfig`. When settings are saved via `rpcSetNetworkSettings`, `config.NetworkConfig` gets replaced with a new pointer, but `TimeSync` still holds the old one. A `SetNetworkConfig()` method on `TimeSync`, called after saving, fixes this.

#### 4. Empty DNS arrays trigger a false confirmation dialog

`dirty.ipv6_static?.dns?.every(dirty => dirty)` returns `true` on empty arrays (vacuous truth in JS). When the DNS array defaults to `[]`, modifying any other field can falsely show an "IPv6 DNS" change in the confirmation dialog. Adding a `.length` check before `.every()` on both IPv4 and IPv6 DNS prevents this.

agh:
Thanks for testing on hardware and the thorough write-up — all four issues were real. Rebased onto current `dev` and pushed fixes in 9f209a1:

**1. Hostnames filtered out at runtime** — `filterNTPServers` now passes entries through when `net.ParseIP()` returns nil, treating them as hostnames. The NTP library (`beevik/ntp`) handles DNS resolution itself, so these are valid inputs. IP-literal entries still go through the IPv4/IPv6 connectivity check as before.

**2. Ordering never includes `ntp_user_provided`** — Rather than setting the ordering from the frontend in `prepareSettings`, I handled this in `getSyncMode()` on the backend: when `TimeSyncMode` is `"custom"`, the default ordering becomes `[ntp_user_provided, http_user_provided, ntp_dhcp, ntp, http]`. This way any API client gets the right behaviour without needing to know about ordering semantics. If either server list is empty, `filterNTPServers`/`queryAllHttpTime` return nil and the loop falls through to the next source. An explicit `TimeSyncOrdering` in the config still takes precedence.

**3. Stale config pointer** — Added `SetNetworkConfig()` on `TimeSync`, called from `rpcSetNetworkSettings` immediately after `config.NetworkConfig` is reassigned. Follows the same pattern as the existing `SetDhcpNtpAddresses` setter.

**4. Empty DNS arrays** — Added `.length > 0` guards before the `.every()` calls on both IPv4 and IPv6 DNS dirty arrays.

jkremlacek:
@adamshiervani can you please check if any additional changes are needed? It would be great to have this PR merged.

PR #1297, opened 2026-03-16
fix: auto-recover USB gadget when host power-cycles

### LINKED ISSUE #128: USB not detected if not present at startup

adamshiervani:
Fixes #128 

### Summary
When the JetKVM is powered via the DC power extension rather than USB, the connected computer rebooting or losing power causes the USB gadget to enter a "not attached" state that never recovers - leaving keyboard and mouse permanently disconnected until a full power cycle.

This PR adds automatic USB gadget self-healing: the existing 500ms state-polling loop now detects a detached gadget and triggers a UDC rebind, closes stale /dev/hidg* file descriptors, and retries opening the new device nodes with increasing delays (~20s window). If the lightweight rebind fails, it escalates to a full gadget reconfigure. During recovery, transient HID errors (ENXIO, ESHUTDOWN, ENOTCONN, EPIPE) are suppressed so callers don't see noise, and reports are silently dropped while USB is down. Recovery is rate-limited (5s cooldown) and respects intentional user actions - disabling USB emulation or changing USB config won't trigger spurious recovery attempts.

### Checklist
- [ ] Lints pass; CI green


internal/usbgadget/hid_keyboard.go
```diff
@@ -324,8 +386,9 @@ func (u *UsbGadget) keyboardWriteHidFile(modifier byte, keys []byte) error {
 	_, err := u.writeWithTimeout(u.keyboardHidFile, append([]byte{modifier, 0x00}, keys[:hidKeyBufferSize]...))
 	if err != nil {
 		u.logWithSuppression("keyboardWriteHidFile", 100, u.log, err, "failed to write to hidg0")
-		u.keyboardHidFile.Close()
-		u.keyboardHidFile = nil
+		u.keyboardLock.Lock()
```
IDisposable, line 389: This is a different lock from the `keyboardWriteHidFileLock` that gets acquired at line 380 (with a `defer`ed  `Unlock`), are we always acquiring them in the same order, or should we be using that lock? Locks taken in different orders is the most common source of hangs.

This seems odd because the `ResetHIDFiles` only references this new lock.

Makefile
```diff
@@ -92,12 +92,12 @@ test_e2e:
 	$(MAKE) check
 	$(MAKE) build_dev VERSION_DEV=0.0.1-test-baseline SKIP_UI_BUILD=1
 	mv bin/jetkvm_app bin/jetkvm_app_baseline
-	$(MAKE) build_dev VERSION_DEV=$(TEST_VERSION) SKIP_UI_BUILD=1
-	cd ui && npm ci && npx playwright install chromium && cd ..
```
IDisposable, line 96: Is the removal of the `npm ci` intentional? Seems odd.

usb.go
```diff
@@ -79,14 +93,109 @@ func rpcGetKeysDownState() (state usbgadget.KeysDownState) {
...
+	usbState     = usbgadget.USBStateUnknown
 	usbStateLock sync.Mutex
+
+	usbEmulationDesired = true
+	lastUSBRecoveryTry  time.Time
 )
 
+func usbReadyForHidReports() bool {
+	usbStateLock.Lock()
+	state := usbState
```
IDisposable, line 105: Would making state just be an atomic be clearer at this point?

usb.go
```diff
@@ -79,14 +93,109 @@ func rpcGetKeysDownState() (state usbgadget.KeysDownState) {
...
+
+	// Clear stale /dev/hidg* handles from the pre-rebind gadget instance.
+	// The next write/open must use the newly recreated device nodes.
+	gadget.ResetHIDFiles()
+
+	// After rebind, the kernel recreates /dev/hidg* but the character
+	// devices take several seconds to become usable (ENXIO until the
+	// function driver attaches). Retry the keyboard HID file open with
+	// increasing delays up to ~20 seconds total.
+	delays := []time.Duration{
```
IDisposable, line 159: _nit_ there's really no need for this to be external to the `tryReopenKeyboard` function, should really move it inside that (just above the `for`) and remove from the parameter lists.

techlobo:
It appears as if this may have introduced another issue: see #1540

PR #1303, opened 2026-03-16
feat(build): add podman support for dev deploy scripts

tvinhas:
## Summary
 - Add podman as an alternative container runtime to docker for dev builds
 - Auto-detect available runtime (docker preferred, podman fallback), overridable via `CONTAINER_CMD` env var
 - Fix cross-compilation on Apple Silicon with podman by patching the Dockerfile
   build context to hardcode `linux/amd64` platform (podman/buildah ignores
   `--build-arg BUILDPLATFORM` overrides unlike Docker BuildKit)
 - Add `--pull=never` for podman `run` to use the locally-built image instead
   of attempting to pull from the private registry

## Test plan
 - [x] Verified `podman build` produces a working amd64 container image
 - [x] Verified cross-compiler toolchain runs correctly inside the container
 - [x] Verified full `build_release` completes and produces correct ARM binary
 - [x] Docker path is unchanged (no behavioral difference for docker users)

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -37,6 +37,21 @@ DOCKER_GO_CACHE_DIR=${DOCKER_GO_CACHE_DIR:-$(pwd)/.cache}
 
 BUILD_IN_DOCKER=${BUILD_IN_DOCKER:-false}
 
+# Auto-detect container runtime: prefer docker, fall back to podman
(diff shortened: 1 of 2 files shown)


adamshiervani (approved):
LGTM — clean detection logic, Podman workaround makes sense.

PR #1309, opened 2026-03-17
feat: offline update upload via web UI

### LINKED ISSUE #96: [feat] system: offline update
Is it possible to add a way o update firmware in offline mode?

This would means:
- adding version tags in repo
- having releases binaries available on release tab here and/or on website

UI would looks as simple as this (with maybe a link to grab latest release):
Image

agh:
Implements the offline update feature requested in #96, following the design direction from https://github.com/jetkvm/kvm/issues/96#issuecomment-4073557850.

## Problem

Devices on air-gapped networks have no way to update firmware through the web UI. The only options are DFU mode or manual SSH — neither accessible to most users. The online update path requires both internet access (for downloading binaries) and keyserver access (for fetching the GPG signing key), making it unusable on isolated networks.

## What this does

Adds an "Offline Update" section to Settings → General that lets users upload `.tar.gz` update archives directly through the web UI. No internet connection required on the device — not for the binary, not for the signing key.

## Archive format

Each offline update archive contains four files:

```
jetkvm_app            # the update binary
jetkvm_app.sha256     # SHA256 hash (hex digest)
jetkvm_app.sig        # detached GPG signature
jetkvm_app.pub        # armored GPG public key used to produce the signature
```

System archives follow the same structure with `update_system.tar` as the binary name. The `.pub` file is what makes fully offline verification possible — the device validates it against the pinned root key fingerprint (`internal/ota/gpg.go:21`) before using it, the same trust anchor as the online path.

## Signature verification

Verification is required and binary: it passes or it fails. There is no bypass option.

1. Extract archive, validate structure (exactly 4 files, no path traversal)
2. Compute SHA256 of the binary, compare to `.sha256` — reject on mismatch
3. Parse the bundled `.pub`, validate its fingerprint against the hardcoded root fingerprint via `parseAndValidateKeyring` — reject if it does not match
4. Verify the `.sig` against the binary using the validated key — reject on failure

This is equivalent in security to the online path. The trust anchor is identical: the root key fingerprint compiled into the running firmware. A malicious archive cannot substitute a rogue key because the fingerprint check will reject it, whether the key arrives from a keyserver or from the archive.

### Key rotation model

`rootKeyFP` is currently a single fingerprint. The rotation model for future key changes would be:

| Release | Trusts | Signed with | Can install offline archives signed by |
|---------|--------|-------------|---------------------------------------|
| v1 | Key A | Key A | A |
| v2 | Key A, Key B | Key A | A or B |
| v3 | Key A, Key B | Key B | A or B |
| v4 | Key B | Key B | B only |

Rollback from v4 to v1 correctly fails — you cannot roll back past a key removal. Expanding `rootKeyFP` to `[]string` is a separate change; not needed until a rotation is imminent.

## Backend

Two HTTP endpoints behind `protectedMiddleware`:

- **`POST /ota/upload`** — accepts a multipart `.tar.gz` archive with a `component` field (`app` or `system`). Extracts to a temp directory, runs the verification pipeline above, stages the verified binary at the standard OTA path on success.

- **`POST /ota/apply`** — applies a previously verified and staged offline update. Triggers `rk_ota` for system updates or binary swap for app updates. The device reboots on completion.

`VerifySignatureFromFileWithKey` is a new method on `GPGVerifier` that accepts raw armored key bytes, validates the fingerprint via `parseAndValidateKeyring`, and calls `CheckDetachedSignature` with the resulting single-entity keyring. No keyserver interaction, no cache mutation. Depends on the `parseAndValidateKeyring` fix from #1316.

The existing `applySystemImage()` (extracted from `updateSystem()`) is reused for system updates.

## Frontend

`OfflineUpdateCard` component with two independent sections (app / system). Each has a file input, upload progress bar, and verification status indicators showing hash and signature results. Rendered on the general settings page between the auto-update toggle and the reboot button.

## Release pipeline

`offline_archive_app` Makefile target packages `jetkvm_app` + `.sha256` + `.sig` + `.pub` into `jetkvm_app_offline_update.tar.gz`. Wired into the production release flow.

The equivalent system archive target is not included — the system image is built in a separate repo that does not have GPG signing in its pipeline yet.

## Tests

31 tests in `internal/ota/offline_test.go`:

**Extraction** (10): valid app/system, missing hash/sig/pub/binary, unexpected files, path traversal, corrupt gzip, nested directories

**Verification** (8): valid signature, hash mismatch, invalid signature, wrong key (fingerprint mismatch on bundled `.pub`), empty signature, empty public key, truncated signature, corrupted binary on disk

**End-to-end** (5): full extract→verify pipeline — valid archive, tampered binary, wrong signer, wrong hash, system component

**Utilities** (3): `ComponentUpdatePath` (app/system/unknown)

**Other** (5): `hashFile`, `VerifySignatureFromFileWithKey` (covered transitively through offline tests and `TestParseAndValidateKeyring_FiltersRogueKeys` in `gpg_test.go`)

## Not included

- E2E Playwright tests (want feedback on the approach first)
- Drag-and-drop on the file input
- `offline_archive_system` Makefile target (blocked on system repo signing pipeline)
- `rootKeyFP` → `[]string` expansion (separate change)

Closes #96

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -314,17 +314,18 @@ _build_release_inner: build_native
 		-o $(BIN_DIR)/jetkvm_app cmd/main.go
 
 # Package a signed app binary into an offline update archive.
-# Expects bin/jetkvm_app, bin/jetkvm_app.sha256, and bin/jetkvm_app.sig
-# to already exist (produced by the signing step in release/test_production_release).
+# Expects bin/jetkvm_app, bin/jetkvm_app.sha256, bin/jetkvm_app.sig,
+# and bin/jetkvm_app.pub to already exist (produced by the signing step
+# in release/test_production_release).
 offline_archive_app:
 	@echo "Creating offline update archive for app..."
-	@for f in jetkvm_app jetkvm_app.sha256 jetkvm_app.sig; do \
+	@for f in jetkvm_app jetkvm_app.sha256 jetkvm_app.sig jetkvm_app.pub; do \
 		if [ ! -f "$(BIN_DIR)/$$f" ]; then \
 			echo "Error: $(BIN_DIR)/$$f not found. Run signing step first."; exit 1; \
 		fi; \
 	done
 	tar czf $(BIN_DIR)/jetkvm_app_offline_update.tar.gz \
-		-C $(BIN_DIR) jetkvm_app jetkvm_app.sha256 jetkvm_app.sig
+		-C $(BIN_DIR) jetkvm_app jetkvm_app.sha256 jetkvm_app.sig jetkvm_app.pub
 	@echo "✓ Created $(BIN_DIR)/jetkvm_app_offline_update.tar.gz"
 
 release: git_check_dev check_r2
--- a/Makefile
+++ b/Makefile
@@ -313,6 +313,20 @@ _build_release_inner: build_native
 		$(GO_RELEASE_BUILD_ARGS) \
 		-o $(BIN_DIR)/jetkvm_app cmd/main.go
 
+# Package a signed app binary into an offline update archive.
+# Expects bin/jetkvm_app, bin/jetkvm_app.sha256, and bin/jetkvm_app.sig
+# to already exist (produced by the signing step in release/test_production_release).
+offline_archive_app:
+	@echo "Creating offline update archive for app..."
+	@for f in jetkvm_app jetkvm_app.sha256 jetkvm_app.sig; do \
+		if [ ! -f "$(BIN_DIR)/$$f" ]; then \
+			echo "Error: $(BIN_DIR)/$$f not found. Run signing step first."; exit 1; \
+		fi; \
+	done
+	tar czf $(BIN_DIR)/jetkvm_app_offline_update.tar.gz \
+		-C $(BIN_DIR) jetkvm_app jetkvm_app.sha256 jetkvm_app.sig
+	@echo "✓ Created $(BIN_DIR)/jetkvm_app_offline_update.tar.gz"
+
 release: git_check_dev check_r2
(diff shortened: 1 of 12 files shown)


equinox0815:
hmmm, now that i have looked deeper into this and #1188 my [question](https://github.com/jetkvm/kvm/issues/96#issuecomment-4073646141) from earlier is answered.

Boy-oh-boy i have to say that i find it very hard to express my bewilderment about how the signature mechanism works without using foul language. Why is this so horrendously complicated? Downloading the singing key from a keyserver is just something i have never seen any software do. And sorry i don't mean that in a good way. This dependency on a keyserver being reachable in order to upgrade the device makes the whole offline upgrade feature moot. How many devices out there will be unable to download the binaries directly but will be able to reach a keyserver? 
What were the design goals in mind when the signature verification was implemented? Usually adding this much unneeded complexity is not the path to a robust and secure mechanism. I strongly encourage you to revisit the design decisions that yielded this result. All i can say is that now i am even more convinced than ever that it was a good idea to put this device in my management VLAN without any internet access. 

Of course, this is not your fault @agh. This is also likely not the place to discuss this but #1188 has already been merged so apparently the project team does not see it my way. I guess i have to live with that. The only thing i can think of to make upgrades truly offline would be to also add a keyring containing the signing key to the tarball and importing it as if it would come from a keyserver. Of course this makes this setup even more complex but at least it makes sense this way.

agh:
Okay, @adamshiervani, I think this is probably worth reviewing now. 👍🏻

Gunni:
Remove dependency on keyserver, unless you already did, it makes the whole feature moot.

My device will never have internet access.

PR #1312, opened 2026-03-19
feat(remote): add support for custom tailscale control servers

dotWee:
Closes ##1311

### Summary

- Add support for configuring a custom Tailscale control server (Headscale-compatible) by introducing a persisted `tailscale_control_url` setting.
- Expose new JSON-RPC methods to read and update the control server URL:
  - `getTailscaleControlURL`
  - `setTailscaleControlURL`
- Apply control-server changes to the local Tailscale client using CLI parity with `--login-server`, including fallback behavior:
  - first try `tailscale set --login-server=...`
  - fallback to `tailscale up --login-server=...` if needed
- Extend Tailscale status responses with `controlURL` so the UI can display the effective endpoint.
- Update the device UI Tailscale card to:
  - show active control server
  - allow editing and saving a custom control server URL
  - allow clearing the field to use the default Tailscale control plane
- Add/expand tests for:
  - URL normalization/validation
  - status parsing with effective control URL
  - command fallback logic (`set` -> `up`)
  - RPC save/apply behavior and “not installed” behavior

### Motivation

Users running self-hosted control planes (for example, Headscale) need the device-side Tailscale integration to point at non-default control servers.  

The current implementation only reports status and does not support control-plane customization, which blocks self-hosted deployments.

### Implementation Details

- Added config field:
  - `tailscale_control_url` (`omitempty`) in persisted config.
- Added effective default handling:
  - empty configured value resolves to `https://controlplane.tailscale.com`.
- Added strict control URL validation to prevent ambiguous or unsafe values:
  - requires `http://` or `https://`
  - requires host
  - rejects userinfo, query, fragment, and non-root paths.
- Added deterministic test seams for installation checks, command execution, and config save, enabling reliable unit tests independent of host environment.
- Kept status behavior non-fatal when daemon interaction fails (existing behavior), now with `controlURL` still reported.

### Backward Compatibility

- Existing users without `tailscale_control_url` continue using the default Tailscale control plane.
- New fields/methods are additive; no existing RPC methods were removed or behavior-broken.
- Clearing a custom value reverts to default behavior.

## Test Plan

- [x] Unit tests cover status parsing with custom/default control URL.
- [x] Unit tests cover control URL validation and normalization.
- [x] Unit tests cover command fallback (`set` failure -> `up` success).
- [x] Unit tests cover RPC setter behavior (save/apply and not-installed path).
- [x] UI lint checks passed for modified frontend files.

### Checklist

- [x] Lints pass; CI green


adamshiervani (changes requested):
Two issues:

1. **Config saved before apply — no rollback on failure.** `rpcSetTailscaleControlURL` persists the new URL to disk (line 218-219) before calling `applyTailscaleControlURL` (line 227). If the apply fails, the error is returned to the UI, but the config on disk already has the broken URL. On next reboot, the daemon will try the failed URL. Fix: move the save after a successful apply, or rollback the config on failure.

https://github.com/jetkvm/kvm/blob/793c373a8bb78f77763677648a1d5e615cb727ee/tailscale.go#L210-L232

2. **`tailscale up` fallback is destructive.** When `tailscale set --login-server` fails, `applyTailscaleControlURL` falls back to `tailscale up --login-server` (line 133). `tailscale up` re-establishes the full connection — it logs the device out of the current control plane and starts a new auth flow. If the device is currently connected via Tailscale and the new URL is wrong, the operator loses remote access. `tailscale set` is the correct non-destructive command (available since Tailscale 1.34+). Drop the `tailscale up` fallback, or at minimum skip it when `BackendState != "Stopped"`.

https://github.com/jetkvm/kvm/blob/793c373a8bb78f77763677648a1d5e615cb727ee/tailscale.go#L123-L144

dotWee:
Thanks for the review @adamshiervani! Legitimate issues, I covered point 1 in https://github.com/jetkvm/kvm/pull/1312/commits/bf43db76379c68b6605a0207cbf00ee7bb0a5878 and point 2 in https://github.com/jetkvm/kvm/pull/1312/commits/13ca066a7033ab32ec24f790a99651ce2b8088c9.

Haven't addressed the main documentation in https://github.com/jetkvm/docs yet, I'd followup a PR once this feature is approved here and ready to merge.

DEVELOPMENT.md
```diff
@@ -250,6 +251,32 @@ curl -X POST http://<IP>/auth/password-local \
   -d '{"password": "test123"}'
 ```
 
+### Tailscale control server testing
```
adamshiervani, line 254: This is not true. We don't have an endpoint for rpc calls

ui/src/components/TailscaleCard.tsx
```diff
@@ -55,6 +81,30 @@ export default function TailscaleCard() {
...
+                size="XS"
+                theme="light"
+                type="button"
+                text={isSavingControlURL ? "Saving..." : "Save"}
+                disabled={isSavingControlURL}
+                onClick={saveControlURL}
+              />
+            </div>
+          </div>
+
```
adamshiervani: Could you make this a dropdown instead with 2 options. Default and Custom. If Custom is selected the input field show up and you can enter the value. The idea is progressive disclosure on features most people wont use.

adamshiervani (changes requested):

DEVELOPMENT.md
dotWee: Thanks for the review! Addressed this in commit https://github.com/jetkvm/kvm/pull/1312/commits/d5dc46a000335920c27e0993a4dea9fe26e8efed.

ui/src/components/TailscaleCard.tsx
dotWee: Addressed this in commit https://github.com/jetkvm/kvm/pull/1312/commits/d5dc46a000335920c27e0993a4dea9fe26e8efed as well.

adamshiervani (approved):

PR #1316, opened 2026-03-22
fix(ota): return only the matched entity from parseAndValidateKeyring

agh:
`parseAndValidateKeyring` validates that at least one entity in a fetched keyring matches the pinned root key fingerprint, but on match it returns the entire keyring — including any additional entities the keyserver included in its response.

`openpgp.CheckDetachedSignature` accepts a signature from any key in the provided keyring. A compromised or malicious keyserver could return a response containing the legitimate JetKVM release key (satisfying the fingerprint check) alongside an attacker-controlled key. A binary signed with the attacker key would then pass signature verification in both `VerifySignature` and `VerifySignatureFromFile`.

### What changed

Single-line fix in `internal/ota/gpg.go`: `parseAndValidateKeyring` now returns `openpgp.EntityList{entity}` (only the matched entity) instead of the full keyring.

New test `TestParseAndValidateKeyring_FiltersRogueKeys` in `internal/ota/gpg_test.go` constructs a two-entity armored keyring (trusted + rogue), asserts only the trusted key survives filtering, and confirms `CheckDetachedSignature` rejects a signature produced by the rogue key.

### What this does not change

The keyserver fetch logic, caching behaviour, and fingerprint matching are all untouched. The fix is narrowly scoped to the return value.

Credit to @equinox0815 for identifying this in https://github.com/jetkvm/kvm/issues/96#issuecomment-4103220615.

### DIFF
--- a/internal/ota/gpg.go
+++ b/internal/ota/gpg.go
@@ -195,7 +195,7 @@ func (g *GPGVerifier) parseAndValidateKeyring(key []byte) (openpgp.EntityList, e
 
 		fp := normalizeFingerprint(hex.EncodeToString(entity.PrimaryKey.Fingerprint[:]))
 		if fp == expected {
-			return keyring, nil
+			return openpgp.EntityList{entity}, nil
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


adamshiervani (approved):
Thanks for taking care of this!

PR #1452, opened 2026-05-08
feat(video): Add 120hz Support

zalo:
# feat(video): Add 120hz Support

https://github.com/user-attachments/assets/29e4429f-377a-4720-bd88-c28059b06875

## Summary

Two changes that together let the JetKVM stream **1280×720 @ 120 fps** end-to-end on existing v1 hardware, without a separate "low latency" toggle or a separate EDID entry in the dropdown:

1. **Add 1280x720@120 to the JetKVM default EDID** as base-block DTD1 (replacing the previous 720p60 DTD; 720p60 is still advertised via the Standard Timings block at offset 40, so all existing modes stay reachable). Source picks the rate via OS display settings (`xrandr --rate 120`, Windows Display Settings, etc.) — no EDID swap required.
2. **Plumb the source vrefresh into the MPP encoder's rate-control config**. The encoder previously defaulted to `fps fix [30/1]` because the firmware never set `Src/DstFrameRateNum/Den`, so the H.264 bitrate budget was sized for 30 fps regardless of source rate. Empirically the encoder still forwards 120 input frames at the wire rate either way — but with rate control sized for 30, frame *quality* visibly degrades at 120 fps (perceptible juddering / smearing). Setting the rate-control fields to the actual source rate eliminates the choppiness. GOP also scales to `fps/2`, keeping the IDR cadence at ~0.5 s for any refresh.

Glass-to-glass latency on the source-capture leg drops from ~16.7 ms (60 fps) to ~8.3 ms (120 fps) when the source PC is running at 120 Hz, with `inbound-rtp.framesPerSecond ≈ 120` sustained on a clean LAN.

## Why DTD1 in the base block, not the CTA extension

NVIDIA's display driver enumerates base-block DTDs reliably but ignores DTDs in the CTA-861 extension that don't carry a CTA VIC. 1280x720@120 isn't a CTA VIC, so an earlier iteration that put 720p120 in the CTA extension's first DTD slot silently dropped the 120 Hz mode on GeForce hosts — `xrandr` only listed 1080p60 / 720p60 / 640p variants, with no 119.91 Hz row.

Solution: put 720p120 in the base block (DTD1) where every driver respects it. The displaced 720p60 mode survives via the existing Standard Timings entry `0x81C0` at offset 40 of the base block, which is read by every modern driver.

## Why 120 Hz is the ceiling on JetKVM v1

The capture chip is the **Toshiba TC358743XBG** (vendor-modified `tc35874x` driver in the BSP). Its datasheet specifies "Video Formats Support (Up to 1080P @60fps)" and characterizes the chip only at 60 Hz. The kernel `dv_timings_cap` is permissive (0–310 MHz, CTA-861/DMT/CVT/GTF/CUSTOM), but the chip's silicon — specifically the blocks above the TMDS PHY — doesn't reliably lock above ~120 Hz vrefresh.

Bench results on real hardware:

| Mode | Pixel clock | TC358743 lock | WebRTC stats (15 s) |
|---|---|---|---|
| **1280×720 @ 120 Hz** | 131.75 MHz | ✓ | 119.9 fps, 0 dropped |
| 1280×720 @ 130/140/144/150 Hz | 132–160 MHz | ✗ | — |
| 848×480 @ 125–240 Hz | 75–130 MHz | ✗ | — |

All failed modes are well under the chip's 165 MHz TMDS ceiling, so this is not a PHY budget limit — it's the chip's downstream logic refusing to lock above 120 Hz. Exceeding 120 Hz on JetKVM hardware would require an HDMI 2.0-class capture chip (Lontium LT6911UXC, ITE IT6802, RK628F).

## Files changed

| File | Change |
|---|---|
| `internal/native/video.go` | `DefaultEDID` updated. Base block: DTD0 = 1080p60 (preferred, unchanged), **DTD1 = 1280x720@120** (was 720p60), DTD2 = range descriptor, DTD3 = name "JetKVM v1". Standard Timings byte at offset 40 keeps 1280x720@60 advertised. Audio data block, YCbCr 4:2:2, vendor-specific data block in CTA extension all preserved. |
| `internal/native/cgo/video.c` | New `detected_fps` global tracked from the `VIDIOC_QUERY_DV_TIMINGS` event handler. `populate_venc_attr` and `venc_start` now take a `fps` parameter and write it into both `u32SrcFrameRateNum/Den` and `fr32DstFrameRateNum/Den` of the H.264/H.265 VBR rate control struct. GOP scales to `fps/2`. The format-detection thread also restarts the streaming pipeline when the rounded fps changes by more than ±1 fps (so a same-resolution rate change like 720p60 → 720p120 actually reconfigures the encoder). |
| `config.go` | `LoadConfig` migration extended with two new triggers: the previous JetKVM v1 EDID (1080p60 / 720p60 only) and the intermediate JetKVM v1 EDID with 720p120 in the CTA extension only. Both auto-rewrite to `native.DefaultEDID` so existing devices pick up the new layout on next boot. Comparisons use `strings.EqualFold` for case-insensitive match. |
| `ui/src/routes/devices.$id.settings.video.tsx` | Default EDID hex updated to match the new backend constant. |

EDID validates clean (0 errors / 0 warnings) at edidcraft.com.

## Test plan

- [ ] Apply the new firmware. Devices with `EdidString` matching either of the two prior JetKVM defaults get auto-migrated to the new combined EDID; devices with a custom user EDID are left alone.
- [ ] On the source PC, force the GPU to re-read the new EDID. NVIDIA caches aggressively: `xrandr --output HDMI-0 --off && xrandr --output HDMI-0 --auto`. Confirm `xrandr | grep -A 6 HDMI-0` lists `1280x720 119.91`.
- [ ] `xrandr --output HDMI-0 --mode 1280x720 --rate 120` (or Windows Display Settings → 120 Hz). Confirm `v4l2-ctl -d /dev/v4l-subdev2 --query-dv-timings` reports `1280×720, 120.00 frames per second`.
- [ ] Confirm device-side `dmesg | grep "fps fix"` shows `fps fix [120/1] -> fix [120/1] gop i [60]` — the encoder reconfigured to the 120 fps source.
- [ ] Browser WebRTC `inbound-rtp` stats: `framesPerSecond: 120` sustained, zero `nackCount` / `pliCount`. On a 120 Hz monitor, motion should appear smooth (no judder).
- [ ] Switch source PC back to 60 Hz: chip relocks at 60, encoder reconfigures to `fps fix [60/1]`, stream stays continuous.
- [ ] Pick a custom EDID in the UI dropdown, then switch back to "JetKVM Default" — verify the round-trip writes the new combined EDID hex.

## Source-side note

Picking the JetKVM Default EDID does **not** switch the source PC's display mode. The EDID just tells the source which modes the JetKVM is willing to accept. The user must manually pick `1280x720@120` (or `@60`, or stay at `1920x1080@60`) in their OS display settings:

- **Windows**: Display Settings → Advanced display settings → Display adapter properties → List All Modes
- **Linux**: `xrandr --output <port> --mode 1280x720 --rate 120`
- **macOS**: System Settings → Displays (note: macOS often refuses non-native modes and may need a third-party tool)

After source-side mode change, NVIDIA's proprietary driver caches EDIDs aggressively. On the source PC, the `xrandr --output HDMI-0 --off && xrandr --output HDMI-0 --auto` cycle forces a fresh DDC re-read if the new modes don't appear immediately.

## Future work (out of scope)

- **End-to-end glass-to-glass latency measurement**: this PR demonstrates source-side latency reduction (display→encode), but the full chain includes WebRTC jitter buffer behavior worth profiling separately.
- **JetKVM v2 / hardware-revision path**: replacing the TC358743 with an HDMI 2.0 receiver would unlock 1080p120 / 720p240 / 480p240+; this PR doesn't preclude that.
- **Audio re-test at 120 Hz**: the EDID still advertises audio support via the existing CTA-861 audio data block; should be verified that 720p120 + audio composition still negotiates correctly on common sources.

---

> [!NOTE]
> **Medium Risk**
> Touches the native video capture/encode pipeline and EDID migration logic; incorrect fps/EDID handling could cause streaming restarts, degraded quality, or source display-mode regressions.
> 
> **Overview**
> Enables **1280x720@120** as part of the default JetKVM EDID (base-block DTD) and updates the UI’s default EDID to match, improving high-refresh compatibility (notably on NVIDIA hosts).
> 
> Plumbs detected source refresh rate into the Rockchip VENC rate-control configuration (including GOP scaling) and restarts streaming when fps changes, so bitrate/quality and IDR cadence track 60 Hz vs 120 Hz sources.
> 
> Extends config migration to auto-rewrite older built-in EDIDs (including prior JetKVM defaults and the Toshiba chip default) to the new `native.DefaultEDID` using case-insensitive matching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53662f3d04aa671c37b7fa2d9b033ca71cca980f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/src/routes/devices.$id.settings.video.tsx
```diff
@@ -174,37 +169,9 @@ export default function SettingsVideoRoute() {
...
-        } else {
-          setEdid("custom");
-          setCustomEdidValue(receivedEdid);
-        }
-      });
-    });
-  };
-
-  const handleEDIDChange = (newEdid: string) => {
+  // applyEDID does the JSONRPC call + state update without showing a
```
IDisposable: This comment is pretty much noise and looks like Claude explaining a change detritus. Make sure you tell Claude to ONLY insert comments when the "why" isn't obvious (and delete this comment)

ui/src/routes/devices.$id.settings.video.tsx
```diff
@@ -214,17 +181,32 @@ export default function SettingsVideoRoute() {
...
-          edid: edids.find(x => x.value === newEdid)?.label ?? "the custom EDID",
+          edid: matched?.label ?? "the custom EDID",
         }),
       );
-      // Update the EDID value in the UI
-      setEdid(newEdid);
     });
   };
 
+  const handleLowLatencyChange = (enabled: boolean) => {
```
IDisposable: Honestly, I don't see the need for the low-latency flag *at all*. If we want to augment the EDID choices with more commentary, that's fine... but it seems to me that just picking a high frame rate/lower resolution should be an obvious thing here :)

IDisposable (changes requested):
Cool idea. I don't think we need to have the optional flag and all the behind-the-scenes flipping about. Just add the new resolutions+frame rate to the monitor EDID chooser :)

IDisposable (changes requested):
Much better!

The EDID should really be all-caps.

They don't validate, though. Use https://edidcraft.com/?tab=validator-tab

For example 720p120:

Image

Image

ui/src/routes/devices.$id.settings.video.tsx
zalo: I was hedging too much; thank you! 😄

zalo:
~~Hmm... testing it more thoroughly locally with https://testufo.com/frameskipping~~, ~~I'm still dropping half of the frames for an effective framerate of ~60hz~~ (despite the connection stats saying there are 120 frames per second coming in, 0 lost packets, and only a 19ms jitter buffer).~~

~~Hoping this isn't some unbreakable issue... still troubleshooting...~~

Ahah!  It was just because I had a 60hz secondary display plugged in in mirroring mode; Linux doesn't like that (even when the slow display is not the primary monitor), so it was only spitting out 60 unique frames per second.

The new 120hz EDID works end to end! It eventually settles on a 16ms jitter buffer!

zalo:
I added a video to the OP to demonstrate that it does indeed seem to work (though, my Wi-Fi is acting up I think, so it's dropping frames more frequently than I'd like...)

adamshiervani (approved):

PR #1454, opened 2026-05-08
feat(mdns): advertise _jetkvm._tcp via Bonjour/DNS-SD

mcuelenaere:
### Summary

JetKVM devices now advertise themselves on the local network as the DNS-SD service type `_jetkvm._tcp`, so Bonjour browsers can discover every device on a LAN. Today the responder only answers A/AAAA for `jetkvm.local`, so PTR queries for the service type return nothing.

### What's advertised

- **Instance**: the device hostname (e.g. `jetkvm-abc123`)
- **Host**: `jetkvm-abc123.local.` — existing local-name resolution preserved
- **Port**: 80, or 443 when TLS is enabled
- **TXT**:
  - `version=` — `builtAppVersion`
  - `id=` — `GetDeviceID()`
  - `setup=true|false` — flips after the user finishes `POST /device/setup`

### Implementation

Uses `pion/mdns/v2`, which is already in the tree transitively via `pion/webrtc`. pion takes the IPv4 and IPv6 multicast packet conns as separate args, so the existing `MDNSMode=ipv4_only`/`ipv6_only` config keeps working — we just don't bind the disabled family. No second mDNS library, no custom responder. The responder is switched from the legacy `Server()` (A/AAAA only) to `NewServer()` so PTR/SRV/TXT are answered when a service is configured.

The diff against `internal/mdns` is purely additive: the existing A/AAAA path, packet-conn setup, and locking are untouched; the change adds an `MDNSService` option that maps to pion's `WithService`, with TXT entries built from the typed `NewTXTString` constructors (re-exported through `internal/mdns` so the `kvm` package stays pion-agnostic).

Publishing TXT records needs the exported `TXTEntry` API from [pion/mdns#277](https://github.com/pion/mdns/pull/277), so this bumps `pion/mdns/v2` to **v2.2.0**, the first release carrying it. (`golang.org/x/sys` moves with it, pulled by `pion/transport/v4 v4.1.0`.)

### Lifecycle

- Starts on the first `networkStateChanged` event (existing behavior).
- Refreshes after `handleSetup` so the `setup=` TXT flips promptly.
- Refreshes after `setTLSState` so the advertised port follows TLS.
- `SIGINT`/`SIGTERM` calls `mDNS.Stop()`, which closes the multicast sockets. Browsers age the entry out via record TTL (pion does not emit DNS-SD goodbye packets).

### Checklist

- [x] Lints pass (`go vet ./...` green via the buildkit container)

### Test plan (pending hardware)

- [ ] `dns-sd -B _jetkvm._tcp` (macOS) shows the device with the configured hostname, port, and TXT.
- [ ] `avahi-browse -r _jetkvm._tcp` (Linux) shows the same.
- [ ] Two devices on the same LAN appear as separate entries.
- [ ] `MDNSMode=ipv4_only` / `ipv6_only` advertises on that family only; `disabled` advertises nothing.
- [ ] `POST /device/setup` flips TXT `setup` from `false` to `true` on the next browse.
- [ ] `dig @224.0.0.251 -p 5353 jetkvm.local` still resolves.


mcuelenaere:
I'm moving this PR back to draft: an easier way forward should be to wait for https://github.com/pion/mdns/pull/277 to get merged and then just build on top of that

PR #1462, opened 2026-05-14
feat: binary HID-RPC wheel dispatch

mcuelenaere:
### Summary

Wires the `TypeWheelReport` (0x04) binary HID-RPC opcode into firmware dispatch and updates the browser frontend to use it when the HID-RPC channel is ready. The opcode and queue routing already existed in `internal/hidrpc/hidrpc.go`, but `handleHidRPCMessage` had no `case` for it, so binary wheel frames hit the default arm and were silently dropped. Wheel input had to take the JSON-RPC slow path.

Now wheel rides `hidrpc-unreliable-ordered` like mouse motion: per-event JSON encode/parse is gone, latency drops on the ARM, and the binary path matches the existing mouse/keyboard pattern. Wire format is `[0x04][deltaY: int8][deltaX: int8]` — both axes carried.

Two commits:
- `feat(hidrpc): dispatch binary TypeWheelReport (0x04) frames` — length-strict `WheelReport` decoder, dispatch case in `handleHidRPCMessage`, 10 table-driven test cases.
- `feat(ui): send mouse wheel via binary HID-RPC when available` — new `WheelReportMessage` class, `reportWheelEvent` hook, `getMouseWheelHandler` switches to `rpcHidReady ? binary : send(\"wheelReport\", ...)` matching the rel/abs mouse pattern.

The JSON-RPC `wheelReport` method is unchanged; old clients and the transient WebRTC/handshake window keep working.

### Checklist

- [ ] Ran \`make test_e2e\` locally and passed
- [x] Lints pass; CI green (locally: \`go test ./internal/hidrpc/...\` + \`tsc --noEmit\` + \`oxlint\`)

---

> [!NOTE]
> **Medium Risk**
> Changes the end-to-end mouse wheel input path (new binary opcode handling in Go + new frontend message), which could affect scroll behavior/latency if the wire format or channel selection is wrong; fallback to existing JSON-RPC reduces blast radius.
> 
> **Overview**
> Enables mouse wheel events to flow over the binary HID-RPC channel by adding `TypeWheelReport` handling in the Go dispatcher and a strict `WheelReport` decoder (with new unit tests) in `internal/hidrpc/message.go`.
> 
> Updates the UI to emit wheel input as a new `WheelReportMessage` over the unreliable-ordered HID-RPC data channel when `rpcHidReady`, while keeping the existing JSON-RPC `wheelReport` path as a backward-compatible fallback. Also tightens a macro marshal error message and logs RTC channel errors using `e.type`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5f6d70ae4fb3eb43176b12650843af5859292d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/src/hooks/useMouse.ts
+++ b/ui/src/hooks/useMouse.ts
@@ -142,12 +142,7 @@ export default function useMouse() {
 
       if (wheelY === 0 && wheelX === 0) return;
 
-      if (rpcHidReady) {
-        reportWheelEvent(wheelY, wheelX);
-      } else {
-        // kept for backward compatibility
-        send("wheelReport", { wheelY, wheelX });
-      }
+      reportWheelEvent(wheelY, wheelX);
 
       // Apply blocking delay based of throttling settings
       if (scrollThrottling && !blockWheelEvent) {
@@ -155,7 +150,7 @@ export default function useMouse() {
         setTimeout(() => setBlockWheelEvent(false), scrollThrottling);
       }
     },
-    [send, reportWheelEvent, rpcHidReady, blockWheelEvent, scrollThrottling, invertScroll],
+    [reportWheelEvent, blockWheelEvent, scrollThrottling, invertScroll],
   );
 
   const resetMousePosition = useCallback(() => {
--- a/ui/src/hooks/useMouse.ts
+++ b/ui/src/hooks/useMouse.ts
@@ -25,7 +25,7 @@ export default function useMouse() {
 
   // RPC hooks
   const { send } = useJsonRpc();
-  const { reportAbsMouseEvent, reportRelMouseEvent, rpcHidReady } = useHidRpc();
+  const { reportAbsMouseEvent, reportRelMouseEvent, reportWheelEvent, rpcHidReady } = useHidRpc();
   // Mouse-related
 
   const sendRelMouseMovement = useCallback(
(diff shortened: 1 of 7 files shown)


ui/src/hooks/useMouse.ts
```diff
@@ -138,15 +138,20 @@ export default function useMouse() {
 
       if (wheelY === 0 && wheelX === 0) return;
 
-      send("wheelReport", { wheelY, wheelX });
+      if (rpcHidReady) {
+        reportWheelEvent(wheelY, wheelX);
+      } else {
+        // kept for backward compatibility
```
adamshiervani: No need to be backwards compatible nowadays. The Cloud SPA is versioned, and the LAN access always serves a matching Go API and JS SPA. 

Please remove the `wheelReport` code path.

adamshiervani (changes requested):

ui/src/hooks/useMouse.ts
mcuelenaere: Addressed in a5f197b — dropped the JSON-RPC fallback in `getMouseWheelHandler` and removed the `wheelReport` entry from the JSON-RPC handler map. Binary 0x04 is now the only path; `rpcWheelReport` stays but is reached only via `handleHidRPCMessage`.

PR #1479, opened 2026-05-26
Add USB audio gadget permission setting

adamshiervani:
Adds USB Audio to the USB Devices setting as the explicit permission gate while keeping the live audio gadget enabled only when both USB Audio and Audio are enabled.

When USB Audio is disabled, Audio settings now disable the toggle and point users back to Hardware with a concise, inline hint.

---

> [!NOTE]
> **Medium Risk**
> Changes USB composite gadget composition and WebRTC audio attachment, which can trigger host re-enumeration; logic is covered by unit tests and coordinated RPC/config paths.
> 
> **Overview**
> This PR adds **USB Audio** as an explicit permission in **USB Devices**, while the host only sees the audio gadget when both that toggle and **Audio** streaming are enabled.
> 
> **Backend:** `usbgadget.Devices` gains an `audio` flag; gadget config and JSON-RPC (`setUsbDevices`, `setUsbDeviceState`, `setAudioConfig`) keep `AudioEnabled`, USB audio permission, and the live gadget in sync—disabling USB audio clears streaming config, stops audio, and reapplies the composite device via `effectiveUsbDevices()` / `effectiveAudioEnabled()`. Config load reads the file as raw JSON so older configs without `usb_devices.audio` pick up the new default and force `audio_enabled` off when USB audio is off. WebRTC only attaches an audio track when both gates pass.
> 
> **UI:** Hardware settings expose the USB Audio checkbox and updated presets; the Audio page disables its toggle with an inline hint linking to Hardware when USB Audio is off. Locales and a small e2e keyboard readiness check are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 966d43f09d91179c1dec9863c703c1133de48320. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mcuelenaere:
FYI: this just bit me while I was rebasing my PRs: my CDC-NCM usb gadget was silently (partially) broken, because of too few available USB IN endpoints. The RX side was working fine, but the TX side didn't do anything (and I spent some time chasing phantom bugs).

I'll add a UI warning to https://github.com/jetkvm/kvm/pull/1470 about exceeding the IN-endpoint budget to prevent users from making the same mistake in the future

PR #1519, opened 2026-07-06
Fix absolute mouse position on multi-monitor targets + touch input

### LINKED ISSUE #1446: Absolute Mouse Position in Multi-Screen Setup
Hi!

I'd like to elaborate on two scenarios. My main use case for JetKVM is remote control of a lighting console (from the entertainment industry). I'm using an MA3 lighting console which has two integrated touchscreens and a DisplayPort connection for an external touchscreen. Using appropriate adapters, I've connected JetKVM to this port for the external touchscreen and USB for the mouse and keyboard. The lighting console runs on a custom Linux system, without any further configuration options.

JetKVM is set to "absolute mouse position," but unfortunately, it behaves less like a touchscreen (mouse position 1:1 on the HDMI image) and more like a mouse for the entire multi-screen setup. Therefore, the mouse mapping is only applied to the upper right quarter of the screen.

I've also tried "relative mouse position," but that doesn't work either (though I think "absolute mouse position" should be the correct setting).

Here is a screen-cast showing the isse: https://github.com/user-attachments/assets/cca9eb42-8825-4bbc-9965-15b2af2fad38

I've also already spoken with the lighting console manufacturer. He doesn't see any fault on his end. Touchscreens are fully supported and work correctly with touch mapping. The question is whether I've done something wrong, or overlooked a setting that should solve my problem?

I'm aware that JetKVM registers as a "mouse" rather than a touchscreen, but is there any solution I could consider? Does this even fall under the category of "feature request"?

My second test setup involves a Windows laptop. Here, I connected JetKVM to the laptop's HDMI port and extended the screen instead of mirroring it. I didn't record a video in this setup. However, the mouse behaves as if the screen were mirrored, meaning the mouse cursor is only on the laptop's internal screen, not on the extended "display" (JetKVM via HDMI). According to Windows, the supposedly external screen is to the right of the internal screen, yet I can't get the mouse cursor to move to the external screen at all.

The latter setup is not as important to me as the aforementioned lighting control setup, because with Windows I could always use other remote solutions like Anydesk or similar, which is not possible with lighting control.

Any help is highly appreciated! Thank you,

Martin

imcynic:
## Summary

Two related absolute-mouse positioning fixes:

**1. Multi-monitor coordinate mapping (fixes #1446).** The USB HID absolute pointer (0..32767) is interpreted by the target OS across the bounding box of the *entire* multi-monitor desktop, while JetKVM captures a single screen — so on multi-monitor targets, clicks land far away from where you click on the video (measured 1700–3000 px off on a 3-monitor Linux target). This adds an optional device-side remap of absolute coordinates onto the captured screen's rectangle within the full desktop:

- `AbsMouseMappingConfig` in the device config (disabled by default, nil-safe, validated: positivity, rect-fits-in-desktop with overflow-safe arithmetic, sane upper bound)
- Applied at the single `rpcAbsMouseReport` choke point, covering both the HID-RPC datachannel path and the legacy JSON-RPC path; invalid/absent config falls back to identity (stock behavior)
- `getAbsMouseMapping` / `setAbsMouseMapping` JSON-RPC methods + authenticated `GET/POST /mouse-mapping` HTTP endpoints (the latter lets host-side automation keep the mapping in sync when the target's display layout changes, without a WebRTC session)
- New "Multi-Monitor Mapping" section under Settings → Mouse
- Set failures roll back in-memory state; updates are serialized; persistence errors return 500, validation errors 400

**2. Touch-input handling in absolute mode (iPad/tablets).** Relative mode needs pointer lock, which doesn't exist on iPadOS — absolute is the only usable mode there, and it had a position offset plus no drag support:

- Use `clientX/clientY` against `getBoundingClientRect()` instead of `e.offsetX/offsetY` — offsetX/Y is unreliable for touch-synthesized events on iPadOS/WebKit and caused a large offset for touch clients
- Absolute mode now listens to `pointermove` (relative mode keeps `mousemove` for pointer lock), and the video element gets `touch-action: none`, so touch drags move the remote mouse instead of scrolling the page
- Non-primary touches are ignored; `pointercancel` releases buttons; `setPointerCapture` on pointerdown so a release outside the video element isn't lost (touch already captures implicitly; this covers mouse/pen)
- Fixed a `fullscreenchange` listener that was added without cleanup

## Test plan

- [x] Unit tests (`mouse_mapping_test.go`) for validation, the remap math, and clamping — table-driven, incl. identity/disabled/invalid fallback, boundary (0 / 32767), screen==desktop, overflow and bound cases; run on a physical JetKVM (linux/arm)
- [x] Closed-loop hardware verification on a 3-monitor Linux (KDE/Wayland) target: browser input → JetKVM USB HID → measured host cursor position. 9-point grid across the video: **0 px max error** with mapping enabled (was 1700–3000 px off before); mapping disabled reproduces existing behavior exactly
- [x] Touch verification with trusted touch events: tap coordinates byte-exact at the target's evdev layer; drags deliver continuous moves with the button held, clean press/release
- [x] `ui` lint + `tsc` clean; frontend + release builds green
- [ ] `make test_e2e` not run (requires a resettable test device)

Notes: new UI strings are currently in `en.json` only — happy to run the machine-translate pass or adjust wording per your conventions. Also happy to drop the HTTP endpoints if you'd prefer the RPC surface only.


imcynic:
> [CLA assistant check](https://cla-assistant.io/jetkvm/kvm?pullRequest=1519) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/jetkvm/kvm?pullRequest=1519) before we can accept your contribution.
> 
> **Bryan** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/jetkvm/kvm?pullRequest=1519) it.

Not signing a CLA, you're welcome to reference and rewrite

Issue #520, opened 2025-05-24
JetKVM Account Is Restricted To Google Mail Address


Xen0byte:
Are there any plans to allow people to sign up to the cloud dashboard with their non-Google email addresses? To be perfectly honest, if this limitation had been clear from the start, I would have waited until the product gained more maturity. I would be very surprised if I'm the only one who finds this to be a very poor choice, which essentially forces me to have to set up my own VPN just to use the product that I supported in good faith.

Rahzadan:
I think you have to have your non-google e-mail address tied to your google account.

My personal google account doesn't even have a GMAIL address.

Xen0byte:
Thanks @Rahzadan, while this is good to know, the fact that a Google account is required is an absolute deal-breaker. I wish this information was available from the start, but I'll host my own instance for now. It's just really disappointing to see the direction that a Berlin-based company advocating for openness has taken with this.

hhauschild:
I would love to see other social-login options, like "login with github". Google is a bit of a pain sometimes

enykeev:
Google is a massive pain when you want to have a separate account for KVM to share with another person and you already verified two accounts with your phone this year.

adamshiervani:
Thanks for raising this — and to everyone who's added to the thread. You're right that the cloud dashboard currently only supports Google login, and I understand this is frustrating, especially for those who don't want to tie a Google account to their KVM setup or need to share access with others.

Supporting additional login providers (e.g. GitHub, email/password, or other OIDC options) is on our roadmap for the cloud platform. I can't give a firm timeline right now, but the demand is clear and it's not a trivial-but-forgotten request.

enykeev:
It's definitely not trivial. I went through the code base trying to figure out what it would take and was unpleasantly surprised how deeply the dependency on Google OAuth runs. It's not impossible to untangle, but I don't see it happening until someone on the team makes it a priority.

Myself, I just switched to NanoKVM Pro which managed to avoid the issue entirely. GL.iNet Comet Pro seems to also be a decent option.

Issue #99, opened 2025-01-27
[Feature Request] Microsoft login and multi user access


TJP507:
My company uses Microsoft and DUO to authenticate everything. We would like to see the ability to integrate with Microsoft SSO.

With this, we would like to have an "admin" account in the cloud who can share/control access with other users. We would like to have multiple users able to access hundreds of JetKVM devices from their respective Microsoft SSO accounts.

nhalstead:
Even adding a local authentication for OIDC would be beneficial.
Since OIDC has all of the state, the JetKVM doesn't need to store a user database, just store it in a session.

This specific ask, OIDC for Microsoft SSO in an organization structure would be appropriate to open in the [cloud-api](https://github.com/jetkvm/cloud-api) repo.

sunjam:
Found this while looking for overall custom oidc support.  No doubt lots of people have their different SSO services, be them Github or Nextcloud, etc.

Nevexo:
This will happen eventually, just takes a fair amount of work as OIDC is in the cloud API and KVM itself.

ODickins:
If there were a way to vote for features, please add mine to this. The ability to share access over JetKVM Cloud - perhaps even for a limited time, to a third party - would be marvelous.

PR #1533, opened 2026-07-18
Support provider-agnostic cloud OIDC tokens

eekdood:
This updates the device-side cloud registration/session logic to accept provider-agnostic OIDC fields from the cloud API while preserving compatibility with the existing Google-specific fields.

This pairs with the cloud-api change that makes the OIDC provider configurable.  https://github.com/jetkvm/cloud-api/pull/64

## Changes

- Accept provider-agnostic cloud registration fields:
  - `oidcToken`
  - `oidcClientId`
  - `oidcIssuer`
- Preserve legacy fallback fields:
  - `oidcGoogle`
  - `clientId`
- Store provider-aware OIDC identity data:
  - `oidc_issuer`
  - `oidc_identity`
- Keep populating `google_identity` for backward compatibility.
- Verify cloud session OIDC tokens against the stored issuer.
- Fall back to Google issuer/identity behavior for legacy configurations.
- Update WebRTC session auth to accept `OidcToken` while preserving `OidcGoogle`.
- Update the adoption UI to consume and forward the new provider-agnostic fields.

## Compatibility

Existing Google-based cloud registrations continue to work through the legacy fields and fallback issuer. This PR intentionally avoids renaming existing Google-specific config fields beyond adding the new provider-agnostic fields, so the behavioral change stays focused and low-risk.

## Testing

- `gofmt`
- `git diff --check`
- `npm run build:device`
- Fork GitHub Actions passed:
  - lint
  - build/test
- Added tests for:
  - provider-agnostic field precedence
  - legacy Google field fallback
  - issuer/client/subject identity format
  - missing OIDC identity claims
  - fake cloud adoption using a local OIDC issuer/JWKS and signed ID token

### DIFF
--- a/cloud.go
+++ b/cloud.go
@@ -23,10 +23,34 @@ import (
 )
 
 type CloudRegisterRequest struct {
-	Token      string `json:"token"`
-	CloudAPI   string `json:"cloudApi"`
-	OidcGoogle string `json:"oidcGoogle"`
-	ClientId   string `json:"clientId"`
+	Token        string `json:"token"`
+	CloudAPI     string `json:"cloudApi"`
+	OIDCToken    string `json:"oidcToken"`
+	OIDCClientID string `json:"oidcClientId"`
+	OIDCIssuer   string `json:"oidcIssuer"`
+	OidcGoogle   string `json:"oidcGoogle"`
+	ClientId     string `json:"clientId"`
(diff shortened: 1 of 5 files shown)


eekdood:
This should address some of the concerns raised in #99 and #520

PR #1560, opened 2026-08-29
fix: preserve virtual media during USB reconfiguration

### LINKED ISSUE #1559: Mounted virtual media is cleared during USB gadget reconfiguration
### Disclaimer

### Application version

0.5.8

### System version

0.2.8

### Device model

JetKVM

### Remote device Hardware

x86_64 Linux PC

### Remote device OS

NixOS, Linux 7.1.4

### Bug description

A mounted virtual-media image is silently detached whenever the USB gadget configuration is rebuilt. The UI continues reporting the image as mounted, but the kernel mass-storage LUN has no backing file and the host sees an empty placeholder device.

#### Steps to reproduce

1. Upload an ISO to JetKVM storage.
2. Mount it as CD/DVD from the Virtual Media dialog.
3. Confirm the host sees the mounted image.
4. Open **Settings → Hardware**.
5. Change **USB Device → Identifiers** from `JetKVM Default` to `Logitech Universal Adapter`.

Changing the identifier triggers a full USB gadget reconfiguration.

#### Before reconfiguration

JetKVM:

```text
/sys/kernel/config/usb_gadget/jetkvm/functions/mass_storage.usb0/lun.0/file
= /userdata/jetkvm/images/metal-amd64-secureboot.iso
```

Linux host:

```text
NAME  SIZE    TYPE  FSTYPE   LABEL
sr0   297.3M  rom   iso9660  TALOS_V1_10_6
```

#### After reconfiguration

JetKVM:

```text
/sys/kernel/config/usb_gadget/jetkvm/functions/mass_storage.usb0/lun.0/file
= <empty>
```

Linux host:

```text
NAME  SIZE  TYPE  FSTYPE  LABEL
sr0     1G  rom
```

The Virtual Media dialog still reports `metal-amd64-secureboot.iso` as mounted.

#### Expected behavior

The mounted image should remain attached across gadget reconfiguration. If JetKVM must rebuild and re-enumerate the USB gadget, the existing mass-storage backing file should be restored.

#### Root cause

The virtual-media mount path writes the image path directly to the live configfs `lun.0/file` attribute. It does not update the corresponding in-memory gadget configuration.

A later `UpdateGadgetConfig()` rebuild therefore uses the stale empty `file` value and clears the LUN.

This affects deliberate configuration changes and recovery paths that escalate to a full gadget reconfiguration.

#### Verification of proposed fix

A patched build synchronizing both the live configfs attribute and the in-memory gadget configuration was tested on the same JetKVM and Linux host.

The test performed two complete gadget rebuilds:

1. JetKVM identifiers → Logitech identifiers
2. Logitech identifiers → JetKVM identifiers

After both rebuilds:

- the LUN still referenced the mounted ISO;
- the Linux host rediscovered the same 297.3 MiB ISO9660 medium;
- the UI and kernel state remained consistent.

Unit and ARM integration tests also reproduce the configuration transaction and verify that the backing file survives `UpdateGadgetConfig()`.

#### Related issues

- #1512 and PR #1516 exercise full gadget reconfiguration during HID recovery, but do not preserve mounted media.
- #1535 also reports an empty `lun.0/file`, but its image fails during the initial mount rather than being lost during reconfiguration.
- #1540 concerns repeated no-host recovery and a kernel dentry leak; it is a separate problem.

kahlstrm:
Fixes #1559

### Summary

Keep mass-storage backing-file changes synchronized between configfs and the in-memory gadget configuration so rebuilding the USB gadget does not silently detach mounted media.

Route virtual-media updates through the synchronized helper and add unit and ARM device regression tests covering the complete reconfiguration transaction.

### Testing

- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/usbgadget`
- `TestSetMassStorageBackingFileIsPreservedInReconfigurationPlan` verifies the live configfs write and subsequent reconfiguration plan.
- `TestMountedMassStorageImageSurvivesGadgetReconfiguration` on a real JetKVM:
  - unpatched build lost the backing file after `UpdateGadgetConfig()`;
  - patched build preserved it.
- Linux host regression test:
  - stable 0.5.8 lost the mounted ISO after changing USB identifiers;
  - patched build preserved the ISO through JetKVM → Logitech → JetKVM identifier changes.

### Checklist

- [x] Lints pass; CI green

### DIFF
--- a/internal/usbgadget/mass_storage.go
+++ b/internal/usbgadget/mass_storage.go
@@ -1,5 +1,11 @@
 package usbgadget
 
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
 var massStorageBaseConfig = gadgetConfigItem{
 	order:      3000,
 	device:     "mass_storage.usb0",
@@ -23,4 +29,27 @@ var massStorageLun0Config = gadgetConfigItem{
 		// Vendor (8 chars), product (16 chars)
 		"inquiry_string": "JetKVM  Virtual Media",
 	},
+}
+
+func (u *UsbGadget) SetMassStorageBackingFile(imagePath string) error {
+	u.configLock.Lock()
+	defer u.configLock.Unlock()
+
+	item, ok := u.configMap["mass_storage_lun0"]
+	if !ok {
+		return fmt.Errorf("mass storage LUN config not found")
+	}
+	if _, ok := item.attrs["file"]; !ok {
+		return fmt.Errorf("mass storage backing file config not found")
+	}
+
+	backingFilePath := filepath.Join(u.kvmGadgetPath, filepath.Join(item.path...), "file")
+	if err := os.WriteFile(backingFilePath, []byte(imagePath), 0644); err != nil {
+		return fmt.Errorf("failed to set mass storage backing file: %w", err)
+	}
+
+	item.attrs["file"] = imagePath
+	u.configMap["mass_storage_lun0"] = item
+
+	return nil
 }
--- a/internal/usbgadget/mass_storage_test.go
+++ b/internal/usbgadget/mass_storage_test.go
@@ -1,0 +1,67 @@
+package usbgadget
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/rs/zerolog"
+)
+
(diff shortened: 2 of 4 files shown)


kahlstrm:
Superseded by #1563, which implements non-destructive USB recovery and includes integration coverage. Closing this earlier approach.

PR #1562, opened 2026-08-31
fix: only recover USB gadget when a data host is present

### LINKED ISSUE #1534: JetKVM unreachable after ~1-2 weeks uptime, started after enabling MQTT/HA discovery
### Disclaimer

### Application version

0.5.8

### System version

0.2.8

### Device model

JetKVM

### Extension model

ATX Power Control

### Remote device Hardware

Generic PC

### Remote device OS

Proxmox

### Bug description

I'm running into an issue where my JetKVM becomes unreachable on the network after roughly 1-2 weeks of uptime. A power cycle (unplug/replug) resolves it every time, but obviously that's not a real fix.

**When it started**:
This began after I configured MQTT and Home Assistant discovery. Before that, I hadn't noticed this issue, though I can't rule out that it was already present and just went unnoticed.

**Details**:
Firmware: App 0.5.8 | Sys 0.2.8
I don't know whether the device is still otherwise functioning normally (e.g. still serving KVM video/input locally) when it drops off the network, since I can only access it remotely — I only know it's unreachable over the network at that point. However on an ethernet link level is does still seem to be connected according to my network switch.

**Questions**:
Could this be related to the MQTT/HA discovery feature specifically (e.g. a memory leak, thread hang, or reconnect loop over time)?
Alternatively, could this be temperature related? If so, what are the acceptable temperature thresholds for the device, and is there a way to check historical temperature via MQTT or logs after the fact (since I can't view the display in person)?
Any recommended way to get logs or diagnostics after a hang like this?

Happy to enable Developer Mode / dig into the shell if that helps troubleshoot, just let me know what to run next time it happens.

Also, I enabled the memory sensor (which is disabled by default) and started monitoring it through home assistant. I rebooted the JetKVM at 03:00 last night and I can confirm a slight increase in memory usage over time as can be seen in this graph. I'll continue to monitor if this trend continues even further.

Image

### LINKED ISSUE #1540: Dentry leak from USB gadget rebind loop (no host attached)
## Environment

- App 0.5.8 (official binary, sha256 `ae9616fa9cf5877e2b5ca9d78b628308c990d7952354b72af284e9d74a39e1d3`), System 0.2.8
- microSD-slot hardware variant (RV1106)
- USB-C data port NOT connected to any host (device used headless/out-of-band; HDMI state irrelevant - reproduced with and without HDMI attached)
- All four USB classes enabled (keyboard, absolute/relative mouse, mass storage) - the default

## Summary

With USB emulation enabled but no USB host attached, the app loops forever on:

```
WRN jetkvm usb USB gadget is detached while USB emulation should be enabled; rebinding USB gadget
WRN jetkvm usb keyboard HID file not ready after UDC rebind; attempting full USB gadget reconfigure
```

at roughly 2.8 rebinds/minute. Each rebind cycle leaks ~34 kernel dentry objects that remain
**live-referenced** (pinned) forever. On a 204MB-RAM device this accumulates ~136K dentries/day of
slab until the kernel thrashes and the device appears dead (web UI unreachable at TCP level, local
screen black) while SSH stays up. Only a reboot recovers it. Recurrence is every ~2.5 days at full
rate (historically 4-5 days, because the leak throttles itself as the box slows).

## Measurements

Steady state, host detached (sampled every 600s over 19.7h on one boot):

```
3,290 rebind log lines; in-use dentries 5,967 -> 117,638  =  33.9 dentries per rebind
rate constant at ~136,000/day across the whole window (linear, not bursty)
earlier 4.0h window, same boot: 698 rebinds, +23,526 in-use = 33.7/rebind (same cost)
```

Live phase-lock (10s polling of `/proc/sys/fs/dentry-state` vs the rebind counter):

```
14:30:24  inuse=118018  rebinds=3301
14:30:34  inuse=118018  rebinds=3301   <- flat between events
14:30:44  inuse=118052  rebinds=3302   <- +34 exactly at the rebind
14:30:55  inuse=118052  rebinds=3302
14:31:05  inuse=118085  rebinds=3303   <- +33
```

Intervention test: attaching a USB host (Raspberry Pi) stops the rebind loop and the dentry growth
completely, with every other subsystem (timesync, mDNS, etc.) unchanged - isolating the gadget
rebind path as the source. Host-attached control run, same sampling: **26.2 hours, 158 samples,
0 rebind log lines, in-use dentries 5,773 -> 5,782 (net +9, within noise)**. The identical window
with no host attached accumulates ~148,000.

## End state when it wedges (full forensics available)

At in-use nr_dentry 342,876 (204MB device):

- `dentry-state: 342884 8 45 0 2 0` - only **8 unused**; `echo 3 > drop_caches` frees nothing
  because the objects are live-referenced; kswapd cannot reclaim them either
- meminfo: `Slab 175,524K / SReclaimable 170,768K / SUnreclaim 4,756K / MemFree 2,360K / Cached 6,708K`
- CPU 0% idle / 76% iowait; load pinned ~10.8 (app threads stuck in page-fault I/O; executable pages
  thrash from storage because page cache is squeezed to ~6MB)
- app still "up" but glacial: `[GIN] 200 | 1m45s | GET "/"` - web UI times out at TCP accept for
  real clients, local display blank, SSH usable
- NOT userspace-visible resource leaks: app fds 61 total, system file-nr 118, 11 mounts, 113 /dev
  nodes, 26 configfs nodes (single clean gadget tree) - the references are kernel-internal
- `MemAvailable` stays high (~157MB) throughout and is misleading - the "reclaimable" slab is not
  reclaimable in practice
- no /proc/slabinfo on this kernel, so the leaking cache cannot be named from userspace
- dmesg clean throughout; no OOM (nothing anonymous to kill - the growth is all slab)

## Expected

An idle KVM with nothing plugged into its USB port should be a supported steady state. Either the
rebind loop should back off / not cycle the gadget when no host is present, or the cycle should not
leak dentry references (suspect: something in the configfs/functionfs teardown path holds a dget()
per cycle).

## Related issues (checked before filing - this is not a duplicate)

- #1483 reports load pinned at 9.00 with the same nine media kthreads (`venc`, `vpss`, `vrga_*`,
  etc.) in D state. That signature is **separate and apparently benign**: we observe the identical
  load-9 / D-state pattern on a fully healthy device (memory fine, dentries flat, everything
  working). It should not be conflated with this issue - the leak here is identified by
  `/proc/sys/fs/dentry-state` and slab growth, not by load average.
- #885 (comment): a user describes the device "becoming sluggish / unresponsive" every couple of
  weeks, resolved only by reboot/power-cycle - consistent with this leak at a slower rebind cadence,
  undiagnosed.

## Workarounds tried

- Disabling all USB classes would stop the loop but crashes the app at startup (filed separately)
- `drop_caches`: ineffective (references are live)
- Attaching any USB host: works (loop stops) - current mitigation
- Reboot: full reset, fuse re-arms

Happy to provide the full CSV time series, wedge-state captures, or run patched builds/diagnostics.

lhw:
Fixes #1540 
Fixes #1534 

### Summary

The USB watchdog recovers a gadget that reads 'not attached' by rebinding the UDC and, on a failed HID chardev reopen, reconfiguring the configfs gadget. On the Rockchip RV1106 vendor kernel (5.10.160) this loop leaks live-refcounted dentries even before the configfs reconfigure is reached, about 85 dentries/min with no host attached, until the 204MB device thrashes and appears dead (#1540). This is the upstream 'tree-in-dcache' controlled leak that Al Viro's DCACHE_PERSISTENT series addresses.

---

This was for now tested with my local JetKVM hooked up to a desktop pc that spends a lot of the day off, hence I was running into the issue as described. A better fix would be a kernel that has the issue mentioned above fixed but for now this seems fairly solid without having any downsides as far as I can see. The commits that added the code (#128) were taken into account and should still be correct.

Please note that the code was implemented with the help of an LLM, which also helped in the debugging session as I was off by quite a bit.

### Checklist
- [ ] Lints pass; CI green

### DIFF
--- a/internal/usbgadget/host_present_test.go
+++ b/internal/usbgadget/host_present_test.go
@@ -1,0 +1,85 @@
+package usbgadget
+
+import "testing"
+
+func TestExtconHostState(t *testing.T) {
+	tests := []struct {
+		name        string
+		payload     string
+		wantPresent bool
+		wantKnown   bool
+	}{
+		{
+			name:        "no host (all zero)",
+			payload:     "USB=0\nUSB-HOST=0\nUSB_VBUS_EN=0\nSDP=0\nCDP=0\nDCP=0\nSLOW-CHARGER=0\n",
+			wantPresent: false,
+			wantKnown:   true,
+		},
+		{
+			name:        "real host (SDP set)",
+			payload:     "USB=1\nUSB-HOST=1\nUSB_VBUS_EN=1\nSDP=1\nCDP=0\nDCP=0\nSLOW-CHARGER=0\n",
+			wantPresent: true,
+			wantKnown:   true,
+		},
+		{
+			name:        "USB-HOST set but no SDP",
+			payload:     "USB=1\nUSB-HOST=1\nUSB_VBUS_EN=1\nSDP=0\nCDP=0\nDCP=0\nSLOW-CHARGER=0\n",
+			wantPresent: true,
+			wantKnown:   true,
+		},
+		{
+			name:        "charger only (DCP, VBUS on) -> not a host",
+			payload:     "USB=1\nUSB-HOST=0\nUSB_VBUS_EN=1\nSDP=0\nCDP=0\nDCP=1\nSLOW-CHARGER=0\n",
+			wantPresent: false,
+			wantKnown:   true,
+		},
+		{
+			name:        "charger only (SLOW-CHARGER) -> not a host",
+			payload:     "USB=1\nUSB-HOST=0\nUSB_VBUS_EN=1\nSDP=0\nCDP=0\nDCP=0\nSLOW-CHARGER=1\n",
+			wantPresent: false,
+			wantKnown:   true,
+		},
+		{
+			name:        "VBUS only, no host-capable/charger cable -> known no host",
+			payload:     "USB=1\nUSB_VBUS_EN=1\n",
+			wantPresent: false,
+			wantKnown:   true,
+		},
+		{
+			name:        "empty payload -> unknown (fall back)",
+			payload:     "",
+			wantPresent: false,
+			wantKnown:   false,
+		},
+		{
+			name:        "host-capable lines all zero -> known no host",
+			payload:     "USB=0\nUSB-HOST=0\nSDP=0\nCDP=0\n",
+			wantPresent: false,
+			wantKnown:   true,
+		},
+		{
+			name:        "malformed value ignored, no known cable -> unknown",
+			payload:     "USB_VBUS_EN=notanumber\n",
+			wantPresent: false,
+			wantKnown:   false,
+		},
+		{
+			name:        "recognizable host cable + malformed charger -> known host",
+			payload:     "SDP=1\nDCP=notanumber\n",
+			wantPresent: true,
+			wantKnown:   true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			present, known := extconHostState(tt.payload)
+			if present != tt.wantPresent || known != tt.wantKnown {
+				t.Fatalf(
+					"extconHostState(%q) = (%v, %v), want (%v, %v)",
+					tt.payload, present, known, tt.wantPresent, tt.wantKnown,
+				)
+			}
+		})
+	}
+}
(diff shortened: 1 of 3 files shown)


lhw:
Superseded by #1563

PR #1591, opened 2026-09-07
Fix reliable mouse button state across reconnects

adamshiervani:
### Summary

A mouse report suppressed before the HID handshake can still update the cached button state, and that cache survives channel replacement. Subsequent reports can therefore omit the reliable button transition. Reset the cache for each channel/handshake and update it only after sending a report.

### Validation

The browser regression covers held-button reconnects, reports before the handshake, and reliable release delivery: `CHROME_BIN=/usr/bin/google-chrome node tests/hid-button-generation.mjs` passes. The test also supports Playwright's default browser when `CHROME_BIN` is unset.

UI lint passes with existing warnings; `npx vite build --mode=device` passes. `npx tsc -b` reports the same diagnostics as unchanged `dev` at d58e6e44 (compared after normalizing line numbers). Full device E2E has not been run for this draft.

---

> [!NOTE]
> **Medium Risk**
> Touches live HID pointer reporting and channel selection; a regression could leave mouse buttons stuck or missing on the host after reconnect.
> 
> **Overview**
> Fixes a bug where **absolute mouse button state** could be wrong after WebRTC/HID channel replacement or when reports were dropped before the HID handshake completed.
> 
> In `useHidRpc`, **`sendMessage` now returns a boolean** so callers know whether a report was actually sent. **`lastAbsButtons` resets** whenever `rpcHidChannel` or handshake readiness changes (not a single long-lived ref), and the cached button mask is **updated only after a successful send**. That forces the next report after reconnect or a failed pre-handshake send to treat a button change as new and route it on the **reliable** channel instead of silently skipping the transition.
> 
> Adds a **Playwright remote-agent E2E** (`ra-hid-reconnect.spec.ts`) for session takeover: press, second tab release, “Use Here”, then press/release on the host. Adds a **bundled browser regression** (`hid-button-generation.mjs`) for held buttons across channel swap, pre-handshake suppression, and release delivery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 153d4168f61b528880f7a63bc19b6887486ef258. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/e2e/remote-agent/ra-hid-reconnect.spec.ts
+++ b/ui/e2e/remote-agent/ra-hid-reconnect.spec.ts
@@ -1,0 +1,92 @@
+import { test, expect } from "@playwright/test";
+import { ensureNoPasswordViaAPI, ensureRpcReady, sendAbsMouseMove } from "../helpers";
+import { createRemoteAgent } from "./remote-agent";
+
+const agent = createRemoteAgent();
+
+test("mouse buttons reach the host after a session takeover and reconnect", async ({
+  page,
+  browser,
+}) => {
+  test.setTimeout(90_000);
+  test.skip(!agent, "JETKVM_REMOTE_HOST not set");
+  await agent!.ensureDeployed();
+  await ensureNoPasswordViaAPI();
+  await page.addInitScript(() => {
+    const original = RTCDataChannel.prototype.send;
+    (window as unknown as { __pointerChannels: string[] }).__pointerChannels = [];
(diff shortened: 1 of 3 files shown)


adamshiervani:
Added a hardware E2E regression for session takeover and reconnect. On RV1106 with a Debian Wayland host, the test observes mouse-button events on the host and confirms that the first press and release after reconnect both reach the device. Passed (1 test); the controlled browser regression also passes.

PR #1592, opened 2026-09-07
Stop leaking the DHCP lease label update timer

adamshiervani:
### Summary

The lease lifetime label starts an untracked zero-delay interval for its initial update. That timer keeps running after the label changes or unmounts. Update immediately once, then retain only the existing 30-second interval with cleanup.

### Validation

UI lint passes with existing warnings; `npx vite build --mode=device` passes. `npx tsc -b` reports the same diagnostics as unchanged `dev` at d58e6e44 (compared after normalizing line numbers). Full device E2E has not been run for this draft.

---

> [!NOTE]
> **Low Risk**
> Small presentation-layer timer lifecycle change with a regression test; no auth, data, or API contract changes.
> 
> **Overview**
> Fixes a **timer leak** in `LifeTimeLabel` on the network settings DHCP lease view: the initial “time remaining” refresh no longer uses `setInterval(..., 0)`, which was untracked and kept firing after unmount or when the lease changed.
> 
> The label now **updates once immediately** on mount/effect run, then only the existing **30-second** interval runs, still cleared on cleanup. Shared `LifeTimeLabel` usage (e.g. IPv6 lifetimes) gets the same behavior.
> 
> Adds a **Playwright E2E** (`network-lease-timer.spec.ts`) that loads network settings, drives lease expiry via a test `networkState` RPC message, and asserts a 30s interval is registered with **no zero-delay intervals** after a short wait.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c27d27733baa832bce8aee53581c3ad42c03657b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/e2e/network-lease-timer.spec.ts
+++ b/ui/e2e/network-lease-timer.spec.ts
@@ -1,0 +1,56 @@
+import { test, expect } from "@playwright/test";
+import { callJsonRpc, ensureNoPasswordViaAPI, ensureRpcReady, waitForWebRTCReady } from "./helpers";
+
+test("the DHCP lease view does not install zero-delay intervals", async ({ page }) => {
+  await ensureNoPasswordViaAPI();
+  await page.addInitScript(() => {
+    const original = window.setInterval;
+    (window as unknown as { __intervalDelays: number[] }).__intervalDelays = [];
+    (window as unknown as { __fastIntervals: number }).__fastIntervals = 0;
+    window.setInterval = ((callback: TimerHandler, delay?: number, ...args: unknown[]) => {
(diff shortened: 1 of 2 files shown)


adamshiervani:
Added and passed a browser E2E test against the RV1106 device UI. It verifies that the lease countdown creates a 30-second interval and never creates a zero-delay interval. Since the connected DHCP client omits expiry, the test uses the real lease with an explicit future-expiry notification to exercise that UI path. One test passed.

PR #1628, opened 2026-09-15
fix(jiggler): jiggle using absolute mouse again

### LINKED ISSUE #1604: Mouse jiggler broken on 0.5.9-dev202609071814
### Disclaimer

### Application version

0.5.9-dev202609071814

### System version

0.2.8

### Device model

JetKVM

### Extension model

None

### Remote device Hardware

_No response_

### Remote device OS

_No response_

### Bug description

Jiggler appears not to be working after upgrading from latest stable (0.5.8) to [0.5.9-dev202609071814](https://github.com/jetkvm/kvm/releases/tag/release%2F0.5.9-dev202609071814)

I've tried to disable and enable it, as well as tried some of the default profiles, the result is the same: The computer locks.

Unsure if this can in any way be related to HDMI sleep, I've disabled it in attempts to fix the locking issue.

Jonher937:
PR #1424 switched the jiggler from absolute mouse writes to relative mouse writes. The bug it was fixing (#290) wasn't that jiggling interferes with normal use in general: as long as the mouse keeps moving, every move report resets the inactivity timer, so the jiggler never fires. The actual case was holding a mouse button down without moving, e.g. pausing mid-drag. No movement means no new report, so the timer goes stale even though the button is still held, and the jiggler could fire mid-drag and jump the cursor (and whatever was being dragged) to a fixed corner.

But the chosen fix broke the jiggler for most people: absolute mode is the default, and the set of users actively switching to relative mode is probably few, so the relative mouse HID interface rarely sees any real traffic outside of the jiggler itself.

This puts the jiggle back on the absolute mouse interface, which is the one actually driving the visible cursor, while keeping the idea of #1424: instead of jumping to a fixed spot, it nudges a small random amount from wherever the cursor currently is, tracked via a new lastAbsX/lastAbsY on the gadget. Since the offset is tiny, moving a window mid-drag would not be an issue the way it was with the old fixed jump to (1,1). Falls back to the old relative-mouse jitter if absolute mouse has been explicitly disabled in USB device settings.

Reading the last position and writing the nudge happen atomically under one lock now, so a real mouse report landing in between can't get overwritten. The jiggle also reuses the currently-held button mask instead of clearing it, so it can't release a button mid-drag - closing the exact gap #1424 originally introduced. Falls back to relative jitter if absolute mouse is disabled, or if it hasn't seen a real absolute position yet and relative mode is available; if neither is available it seeds from the center of the coordinate space instead.

Fixes #1604

### Summary
- Jiggle via the mouse mode that is used. 
- Adopt the same behaviour as #1424 implemented across both mouse modes.

### Checklist
- [x] Lints pass; CI green


Jonher937:
It looks like the output of the test job is inconclusive and might need a retry. Does not seem like I can rerun the action workflow.

@adamshiervani can you give the action a retry unless you see something off?
go tests pass locally

Jonher937:
Found the issue with go test, I didn't add the go build tag. Should be ready now :rocket:

ddska:
Thanks for tracking this down - and sorry, the regression is mine (#1424). Your diagnosis matches what I found.

One blocking issue before this lands. The interface choice is gated on `known`:

```
if x, y, known := gadget.GetAbsMousePosition(); gadget.HasAbsoluteMouse() && known {
```

`lastAbsKnown` is only ever set by `AbsMouseReport`, and its only non-jiggler callers are a live session's pointer events (`hidrpc.go`) and `adoptLiveGadget` under `h.AbsPressed`. So after a cold boot with nobody connected, `known == false` and the jiggle falls back to `/dev/hidg2` - the interface this PR is moving away from. The jiggler can't bootstrap the flag itself, because it can't reach the branch that sets it.

That means it works if you connect and move the mouse first, and silently doesn't for an unattended device - which is the case the jiggler exists for. I confirmed both halves on a device: with a session, `hidg1` gets opened; after a cold boot with no session, nothing opens `hidg1` at all.

Suggestion: when the position is unknown, seed from screen center (`absMouseMaxCoord/2`) and still use the absolute interface, rather than falling back to relative.

Two smaller notes, both optional and happy to send them as a follow-up PR rather than grow this one:

- A clamped offset can return the *same* coordinate once the cursor is against a screen edge, and an absolute report repeating the previous position moves nothing. Worth flipping the offset in that case - the random walk does reach the edges over a long idle period.
- The report is still fully swallowed: `writeWithTimeout` reports an expired write deadline as success (deliberately, from #1598) and `rpcHidReport` swallows transient errors, so a jiggler failing every tick logs nothing. Fine for interactive bursts, not for one report per minute.

Jonher937:
@ddska good input, I've pushed a fix that refactors this a bit. Have another look.
I'll try and run it tomorrow

ddska:
@Jonher937  Nice iteration - the atomic read-offset-write under `absMouseLock` is the right structure, and clearing `lastAbsButtons` in `resetHidHandover` is a catch I'd missed.

Two things I think still need resolving.

1. The `known` gate still sends an unattended cold boot to the relative mouse.

```go
if gadget.HasAbsoluteMouse() && (absPositionKnown || !gadget.HasRelativeMouse()) {
```

On the default config (`absolute_mouse: true`, `relative_mouse: true`) with no session since boot, `absPositionKnown` is false and `HasRelativeMouse()` is true, so this evaluates to `(false || !true)` → false and the relative branch runs. That makes the centre-seed path in `JiggleAbsMouse` unreachable on a default device, and the case from #1604 - a device nobody has connected to since it powered on - still jiggles on `/dev/hidg2`. I confirmed on hardware that `hidg1` never gets opened in that scenario.

2. The lock doesn't address the staleness.

The comment says the single lock keeps the cursor from "jumping from a stale position", but it only serialises our own reports. `lastAbsX/lastAbsY` is JetKVM's view of the cursor and the host never reports its own back, so the moment the cursor moves by something we don't see - someone at the target's own mouse, another KVM, the OS - the anchor is stale and the next jiggle is a teleport rather than a ~10 px nudge. Note when the jiggler fires: after 30–60 s with no JetKVM input, which is exactly when someone is working locally on the target.

I ran an equivalent absolute-anchored jiggle on a device and the feedback was immediate - the cursor moves too far and gets in the way during normal work. Locking can't fix this, because the anchor is fundamentally ours rather than the host's.

Worth considering instead: keep #1424's property of never setting a position, only a delta - relative, out-and-back (`+N,+N` then `-N,-N`) with N around 10–20 px instead of 1–3. Net displacement zero, nothing to go stale. Raw writes to `/dev/hidg2` are accepted instantly on the device I've tested, so the endpoint is alive, and the 0.5.9 failure may well have been magnitude (+/- 1–3 px swallowed by pointer acceleration or an idle-detection threshold) rather than the interface being ignored.

Also, when testing, we need to check and absolute and relative mouse.

Issue #1528, opened 2026-07-14
Windows 11 boot delayed by 120 seconds when JetKVM Mass Storage is enabled


esver:
### Disclaimer

- [x] I have read and understood the disclaimer.

### Application version

0.5.9-dev202606301105

### System version

0.2.8

### Device model

JetKVM

### Remote device Hardware

PC Intel Core i7 12700KF, 64Go Ram, RTX 3090

### Remote device OS

Windows 11 25H2

## Summary
Windows take a long time to boot.
I've use WPR to check the boot and found that the failing device is : 
```
FriendlyName   : USB Mass Storage Device
InstanceId     : USB\VID_1D6B&PID_0104&MI_06\7&4D2978B&0&0006
Service        : USBSTOR
Problem        : CM_PROB_FAILED_START
ProblemCode    : 10
ProblemStatus  : 3221225473 / 0xC0000001
HardwareID     : USB\VID_1D6B&PID_0104&MI_06
Bus description: Mass Storage
```

The other JetKVM composite interfaces enumerate successfully:

* `MI_00` to `MI_03`: HID input devices
* `MI_04`: Media Source/Sink
* `MI_06`: USB Mass Storage — fails to start

When I've disabled the USB Mass Storage in JetKVM parameters, Wndows boot quickly.

## Steps to reproduce

1. Enable Mass Storage in the JetKVM settings.
2. Keep the JetKVM USB connection attached to the Windows 11 computer.
3. Shut down and cold-boot the Windows computer.
4. Observe that the Windows take too long to start.
5. Disable Mass Storage in the JetKVM settings.
6. Cold-boot the Windows computer again.
7. Observe that Windows now starts normally without the long delay.

For the moment I don't use the USB Mass Storage so it's OK to start without it.

ImVantexHD:
Does this delay occur on Windows 10 as well?

esver:
I haven't tested it on another PC, I'll try to do some tests on other PC.

For more information, the motherboard is a MAG-B660-TOMAHAWK-WIFI-DDR4 with Intel B660 Chipset.

I've configured the USB to be always on : if my PC is off, JetKvm have USB power from the motherboard, I can access to the JetKvm and start the PC with the Wake-On-Lan command.

I've enabled Audio.

anttitennivaara:
[...]

Issue #412, opened 2025-05-04
All keyboard input not working (paste, virtual keyboard, physical keyboard) after reboot


mgleason2002:
App: 0.3.9
System: 0.2.3

Using an Minisforum MS-01 connected with USB C to A cable that came in the box. OS is Proxmox 8.3.5.

After rebooting the machine, JetKVM does not register any inputs from the Paste text box, virtual keyboard or physical keyboard.

Semi-reliable workaround has been to go to Settings > Hardware > Identifiers and select a different keyboard emulation. Sometimes it requires two or three changes to any random identifier, then all keyboard input, paste, etc will start working normally. Works while the system is up and running, has to be redone after reboot every time.

harrywustl:
Same issue as well, with HP Z240.

molotch:
Same problem, also running proxmox on a ASUS PN51 nuc. Had problems getting USB to be recognized in the first place, but now it says connected but no keystrokes are getting through to the OS.

geozeke:
Had the same issue after the `0.4.1` update. Just happen to check `Settings > Keyboard`, and noticed that my keyboard layout switched from my preferred, to `Belgisch Nederlands` for some reason. Switched it back to my preferred layout, and all is well.

molotch:
Seems to be some rolling updates. I'm still on App: 0.3.9, System: 0.2.4 and update says I'm running the latest version.

I don't even have a Settings > Keyboard setting, only mouse.

Still keyboard presses aren't going through.

molotch:
Mine started working once I updated to 0.4.1 (you have to enable Dev Channel Updates under Advanced in settings). I use the settings Hardware -> Usb Device -> Classes: Keyboard Only and Identifier JetKVM Default.

pessimal:
Everything was working fine yesterday, today no keyboard inputs of any kind. None of the workarounds posted here have worked so far.
App: 0.4.4
System 0.2.4

Edit: In the end I had to physically change the usb port for things to start working again, not even unbinding and rebinding the old one had any effect

Crash1602:
Unfortunately, I’m experiencing the same issue.

App: 0.4.6
System: 0.2.5

[...]

Issue #133, opened 2025-02-02
USB not attached when using ATX board


Pedro-Barreiro:
I connected the 5v power supply to the the ATX board and the RJ-11 to the JetKVM and a separate usb cable to the the jetkvm and my server running proxmox and it gives the error seen in the screenshot.
Tried restarting with it already plugged, different cables and nothing was working.
What ended up working was connecting the power supply and my server to the usb splitter. It seems that even when the RJ-11 is connected with power supply connected to the atx board, the usb-c port on the jetkvm will still try to get the power from the computer usb port, thus giving the error.

rp-:
Yeah, facing the same usb disconnected issue, if jetkvm was connected in the "wrong" order.

But just to be sure it is meant, the the jetkvm is directly connected to the host USB (for keyboard&mouse) and the ATX board USB-C to the power supply?

s1dh:
> Yeah, facing the same usb disconnected issue, if jetkvm was connected in the "wrong" order.
> 
> But just to be sure it is meant, the the jetkvm is directly connected to the host USB (for keyboard&mouse) and the ATX board USB-C to the power supply?

With the same set, I am facing the same issue , at the moment , to be able to use keyboard , we need to unplug the ATX board and power directly the jetkvm with usb-c (and using splitter) , am I right ?

agarity:
I see same problem with the relay board as well.. it appears when power is provided to the jetkvm outside of usb port, the gadget doesnt get connected.  I did seem to find a potential workaround by unbinding and rebinding the dwc3 driver:
  

```
echo ffb00000.usb > /sys/bus/platform/drivers/dwc3/unbind  
echo ffb00000.usb > /sys/bus/platform/drivers/dwc3/bind  

```

After running this, usb reports connected in UI and seems to work.

exodious:
Nothing new to contribute but noting that I have the same issue.

jocado:
Same issue here with the DC Power Control extension in use, after power the device no from cold start of everything connected.

[This workaround](https://github.com/jetkvm/kvm/issues/133#issuecomment-3123064915) works for me too.

jocado:
If anyone know a way to hook into the extensions being triggered, or monitor the status of the extra modules [ DC or ATX power modules status/voltage ] perhaps we could cobble some automation together.

tjrexer:
I just added a script to init.d that runs the workaround after everything else. Restarted the device & USB came right back up. We'll see how it works out in the long term.

```bash
[...]

PR #1641, opened 2026-09-19
fix(usb): leave a bound gadget alone while the host is booting

### LINKED ISSUE #684: System doesn't boot with JetKVM connected
I’m using a JetKVM to control a headless laptop (screen broken) that serves as a Proxmox server. Here’s the issue:

    Works after boot: If I connect the JetKVM after the laptop is already powered on, it functions perfectly.

    Fails to boot: When the JetKVM is connected during startup, the laptop won’t boot (no POST, no BIOS screen).

    BIOS settings tried:

        Toggled USB legacy support (enabled/disabled)

        Verified boot order (no conflicts)

        Tested other relevant BIOS options

    External display works: With a physical monitor attached, the laptop boots normally, and I can access the BIOS.

    Connection attempts:

        USB 2.0/3.0 ports

        USB-C via Y-cable

        USB through a PD hub
        → Same result: No boot if JetKVM is connected at startup.

Question: Any suggestions to fix this? Is this a BIOS limitation, KVM quirk, or something else?

Leafgard:
Fixes #684

### Summary
- While the host reboots, the recovery loop soft-reconnects the gadget every 5 s and rebinds it at the sixth cycle or when VBUS returns. A reconnect with no host fails inside dwc3 (`failed to enable ep0out`) and a rebind that lands as the host controller comes up leaves the mass storage function unable to complete a bulk transfer, so the host firmware stalls on its `INQUIRY` for minutes or never boots. #1528 is the same failure seen from Windows.
- When extcon knows the host state, leave a bound gadget alone: do nothing while the host is absent, wait while it is present but not enumerating (the kernel enumerates the gadget by itself once the controller is back), with one corrective rebind per episode after two minutes for a truly wedged controller. Devices without extcon keep the previous behaviour. Decision logic in `HostEnumerationWait` (`internal/usbgadget/recovery.go`) with unit tests; the wiring in `usb.go` is not unit-tested, like the rest of that file.
- Tested on a Radxa Orion O6 (EDK2 firmware, serial console with the firmware's boot-stage timestamps), JetKVM 0.5.9, virtual media mounted, warm resets and mains cold boots in a loop. Before: 10 of 12 boots stalled on the dead mass storage function. After: 0 of 12 (6 warm resets, 6 cold boots), every boot at the normal 4 s, no `failed to enable ep0out` in `dmesg`, no rebind in the app log, keyboard and virtual media enumerated by the host. Full trace of the failure in #684.

### Checklist
- [ ] Ran `make test_e2e` locally and passed (no e2e setup here; `go test ./internal/usbgadget/` and `go vet` pass, and the change was run on device as a dev build)
- [x] Lints pass; CI green


Leafgard:
Since this PR removes the recovery loop added in #1297, I tested the case that loop was written for (#133, #412): the JetKVM starting while the host is off, and the host rebooting with the KVM attached.

Bench: Radxa Orion O6 (Arm, EDK2 firmware) with a serial console on the firmware, mains and ATX under software control, JetKVM on 0.5.9 versus a dev build of this branch. Virtual media unmounted throughout. 20 cycles, 5 per build per scenario. The firmware prints timestamped post codes, so `E1FF` to `E550` is the time the boot manager spends before the console stage; about 4 s is normal.

### Scenario A, the app starts with the host off, then the host powers on

| App | Gadget attached in time | Time from power button to `configured` | Firmware gap |
| --- | --- | --- | --- |
| 0.5.9 | 3 of 5 | 1.7 to 3.0 s, or 232 s and 253 s | 3.8 s, or 43.8 s |
| this branch | 5 of 5 | 2.0 to 2.9 s | 3.8 s |

In the two failing stock cycles the firmware never saw the gadget. The rebind (`escalating to UDC rebind`, then `UDC had already stopped` in dmesg) landed as the host powered on, UEFI paid a 40 s USB timeout, and the gadget only enumerated four minutes later when the Linux kernel reset the xHCI. The three stock cycles that did attach still logged `failed to enable ep0out` from the soft reconnects.

### Scenario B, warm reset with the KVM attached

| App | Time to `configured` | Firmware gap |
| --- | --- | --- |
| 0.5.9 | 1.6 to 2.1 s, 5 of 5 | 1.5 s in 2 of 5, 15.6 s in 3 of 5 |
| this branch | 2.0 to 2.7 s, 7 of 7 | 1.5 s in 7 of 7 |

### One more thing the logs show

On 0.5.9 the gadget is reset and reconfigured every 21 to 23 s for as long as the host is up, on an idle bench with nothing mounted. On this branch that stops: 4 resets per boot, which are the two host side pairs (power on, then the xHCI handoff), and nothing after. The periodic churn recreates the `/dev/input/by-id/...-event-kbd` node under whatever is reading it, which is a plausible mechanism for the "keyboard stops working after a while" reports (#412). One key read in the stock run failed that way, with the device still present.

Checks in every cycle: `lsusb -d 1d6b:0104` on the host, the five input nodes, and a key sent through the gadget and read from the host. Device present 22 of 22, key read 20 of 22 (one miss per build, both with the device present, my read window was too short).

One thing I cannot explain: during one cycle on this branch the JetKVM rebooted 97 s after the host reset, with the gadget already attached and the host still booting. The dev build's log was in tmpfs and did not survive, and there was no pstore record. It did not happen in the other seven cycles on this branch, and the build has run since without incident. I will keep the log on `/userdata` next time and report if it recurs.

Happy to share the per cycle records and dmesg snapshots if useful.

Leafgard:
Two more reports in #1528 match the scenario this PR targets. @urbenlegend (Arch Linux) and @esver (Windows 11, who opened the issue) both keep the JetKVM on standby USB power that stays on while the PC is off, @esver so they can wake the PC through the JetKVM's Wake-on-LAN. Both see the boot held up by the mass storage function. On Linux it shows as INQUIRY timeouts and resets 22 s apart, and on Windows as the `MI_06` interface failing with code 10. Disabling mass storage makes both boots normal again.

That makes three hosts with the same trigger: an Arm board on EDK2, a Linux PC and a Windows PC. Happy to adjust the approach if you'd rather handle the recovery loop differently.

